PR #1, opened 2019-07-24
Fix for Team Up Tyranitar

lmar50:


axpendix:
By the way, @lmar50 you didn't have to close your PR when there's additional things requested, you should've just committed to the same branch for the fix, then this PR would've get automatically updated.

PR #17, opened 2019-09-04
Uxie Patch

TTTrainer:
This is a concept I was hoping to work around. I am not sure if it works but in theory, it should.
https://github.com/TTTrainer/tcgone-engine-contrib/commits?author=TTTrainer

### DIFF
--- a/src/tcgwars/logic/impl/gen7/UnifiedMinds.groovy
+++ b/src/tcgwars/logic/impl/gen7/UnifiedMinds.groovy
@@ -2187,7 +2187,11 @@ public enum UnifiedMinds implements CardInfo {
 				bwAbility "Secret Territory", {
 					text "If you have Mesprit and Azelf in play, apply Weakness for each Pokémon (both yours and your opponent's) as ×4 instead."
 					actionA {
-            // TODO:
+            getterA (GET_WEAKNESSES) {h->
+              if(h.effect.target.owner == self.owner.opposite && h.effect.target.basic && h.effect.target.active && my.all.find({it.name == "Azelf"}) && ({it.name == "Mespirit") {
+                h.object == x4
+              }
+            }
 					}
 				}
 				move "Psyshot", {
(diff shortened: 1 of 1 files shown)


src/tcgwars/logic/impl/gen7/UnifiedMinds.groovy
```diff
@@ -2187,7 +2187,11 @@ public enum UnifiedMinds implements CardInfo {
 				bwAbility "Secret Territory", {
 					text "If you have Mesprit and Azelf in play, apply Weakness for each Pokémon (both yours and your opponent's) as ×4 instead."
 					actionA {
-            // TODO:
+            getterA (GET_WEAKNESSES) {h->
+              if(h.effect.target.owner == self.owner.opposite && h.effect.target.basic && h.effect.target.active && my.all.find({it.name == "Azelf"}) && ({it.name == "Mespirit") {
```
axpendix:
```suggestion
              if(self.owner.pbg.all.find{it.name == "Azelf"} && self.owner.pbg.all.find{it.name == "Mespirit"}) {
```

src/tcgwars/logic/impl/gen7/UnifiedMinds.groovy
```diff
@@ -2187,7 +2187,11 @@ public enum UnifiedMinds implements CardInfo {
 				bwAbility "Secret Territory", {
 					text "If you have Mesprit and Azelf in play, apply Weakness for each Pokémon (both yours and your opponent's) as ×4 instead."
 					actionA {
-            // TODO:
+            getterA (GET_WEAKNESSES) {h->
```
axpendix: Move this block outside `actionA` and remove `actionA`.

src/tcgwars/logic/impl/gen7/UnifiedMinds.groovy
```diff
@@ -2187,7 +2187,11 @@ public enum UnifiedMinds implements CardInfo {
...
+              if(h.effect.target.owner == self.owner.opposite && h.effect.target.basic && h.effect.target.active && my.all.find({it.name == "Azelf"}) && ({it.name == "Mespirit") {
+                h.object == x4
```
axpendix: this won't work like that because h.object is `List<Weakness>` and "X4" is the feature of Weakness. 
it should work like below:

```suggestion
                h.object = h.object?.collect{
                  def weakness = it.copy()
                  weakness.feature = "X4"
                  weakness
                }
```

PR #36, opened 2019-09-10
Blaine's Quiz Show

TTTrainer:
I know it's not a priority but I'm not that great at advanced stuff so I thought I'd try this. https://github.com/axpendix/tcgone-engine-contrib/commits?author=TTTrainer


axpendix:
That is just a copy paste from the other card (:

axpendix:
It won't work as this requires a new functionality to request a free form message from client in the game engine.

PR #82, opened 2020-03-03
Fix ZACIAN_V_138

Najzeko:
fixed metalEnergies being defined as water energy, now correctly metal energy


luongthomas:
Whoops I didn't see this pull request, I've already fixed it in this commit

https://github.com/axpendix/tcgone-engine-contrib/commit/1958b7322a94036d8ad1f9c42245df3143fd1e56

PR #85, opened 2020-03-17
Add Initial Groovy template of Ruby Sapphire

luongthomas:
This PR adds the groovy implementation file that was generated from Carddb-tools.

I'm not sure the ramifications of adding this to the codebase since there is already an existing implementation in java (from my understanding) that is not exposed in this repo.

I've also changed the set code and series code to `ex` and `EX` respectively.
`RUBY_SAPPHIRE(211, "Ruby & Sapphire", "EX", "ex1"),`

This is consistent to what I've seen in this file:
https://github.com/PokemonTCG/pokemon-tcg-data/blob/master/json/cards/Ruby%20%26%20Sapphire.json

I didn't see this set in KirbyUK's repo so I went with the other json source.

Let me know if this is safe to merge.

### DIFF
--- a/src/tcgwars/logic/impl/gen3/RubySapphire.groovy
+++ b/src/tcgwars/logic/impl/gen3/RubySapphire.groovy
@@ -1,0 +1,2118 @@
+package tcgwars.logic.impl.gen3;
+
+import static tcgwars.logic.card.HP.*;
+import static tcgwars.logic.card.Type.*;
+import static tcgwars.logic.card.CardType.*;
+import static tcgwars.logic.groovy.TcgBuilders.*;
+import static tcgwars.logic.groovy.TcgStatics.*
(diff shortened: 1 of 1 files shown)


axpendix:
Hi, unfortunately this is not safe to merge. Class `RubySapphire` already exists in legacy form in the main repo so we cannot add the same class there.

axpendix:
Though, we need to reimplement them using Groovy. In order to do that, we can to rename this class to `RubySapphireNG` change the definition in `Collection` to match that. 

Note that `Collection` already has `RUBY_SAPPHIRE` entry and you should not modify it right now. Please add a new entry as `RUBY_SAPPHIRE_NG (...,)`.

luongthomas:
Going to close this one, I'll make a new PR for this feature when it's a higher priority

PR #95, opened 2020-04-19
Implemented 40 and 43

ufodynasty:


src/tcgwars/logic/impl/gen8/RebelClash.groovy
```diff
@@ -936,9 +936,14 @@ public enum RebelClash implements LogicCardInfo {
         weakness M
         bwAbility "Playful", {
           text "Once during your turn, when you play this card from your hand to evolve a Pokemon, you may choose an Energy attached to your  opponent’s Active Pokemon and return it to the top of your opponent’s deck."
-          actionA {
-            // TODO
-          }
+          onActivate {r->
+              if(r==PLAY_FROM_HAND && opp.active.cards.filterByType(ENERGY)) {
```
luongthomas, line 940: This if block should be shifted back by a few spaces, if you haven't already you should set your ide to use 2 space indenting

src/tcgwars/logic/impl/gen8/RebelClash.groovy
```diff
@@ -995,7 +1000,9 @@ public enum RebelClash implements LogicCardInfo {
           energyCost C, C
           attackRequirement {}
```
luongthomas, line 1001: Could you add a requirement to check if the opponent's deck is not empty?

src/tcgwars/logic/impl/gen8/RebelClash.groovy
```diff
@@ -995,7 +1000,9 @@ public enum RebelClash implements LogicCardInfo {
...
           onAttack {
-            // TODO
+            opp.deck.subList(0,1).discard()
+             if (self.cards.findAll { it.name=="Cursed Shovel" }) opp.deck.subList(0,1).discard()
```
luongthomas, line 1004: This should discard additional 2 cards off the top off the opponent's deck

PR #115, opened 2020-04-30
Added a confirm to flapple_22 ability

ufodynasty:
I misclicked it while testing and though something as impactful as shuffling a pokemon into its owners deck would benefit from a confirmation


src/tcgwars/logic/impl/gen8/RebelClash.groovy
```diff
@@ -746,12 +746,14 @@ public enum RebelClash implements LogicCardInfo {
...
-            powerUsed()
-            directDamage(20, opp.all.select())
-            self.cards.moveTo(my.deck)
-            removePCS(self)
-            shuffleDeck()
+            if(confirm("Use Apple Drop?"){
+              checkLastTurn()
+              powerUsed()
+              directDamage(20, opp.all.select())
+              qself.cards.moveTo(my.deck)
```
luongthomas, line 753: qself should be self

PR #116, opened 2020-04-30
Fixed Hatterene confirmation

ufodynasty:


src/tcgwars/logic/impl/gen8/RebelClash.groovy
```diff
@@ -2107,21 +2107,23 @@ public enum RebelClash implements LogicCardInfo {
         bwAbility "Mind Hat", {
           text "Once during your turn, you may have each player discard 1 card from their hand. (Your opponent discards first. If either player has no cards in their hand, that player does not discard.)"
           actionA {
-            checkLastTurn()
```
luongthomas, line 2110: I believe we still need checkLastTurn()

src/tcgwars/logic/impl/gen8/RebelClash.groovy
```diff
@@ -2107,21 +2107,23 @@ public enum RebelClash implements LogicCardInfo {
...
+            assert (my.hand || opp.hand): "Neither player have any cards in their hand."
+            if(confirm("Use Mind Hat"){
             powerUsed()
```
luongthomas, line 2112: This should be indented once over to the right

ufodynasty:
Closing and reopening because ipad is being weird

PR #124, opened 2020-05-02
Changed Bronzong_130s Gyro ball

ufodynasty:
Your opponent only switches their active pokemon with one of their benched pokemon if you do first.


luongthomas:
[BRONZONG_130](https://trello.com/c/yhaPc2jl/351-bronzong130)

PR #128, opened 2020-05-02
fixed electivire indent and zamzenta guard press

ufodynasty:
now it prevents damage after w/r and from any of the opponent’s defending pokemon


ufodynasty:
[ZAMAZENTA_140](https://trello.com/c/Qpk32406/361-zamazenta140)

PR #130, opened 2020-05-02
Bandaid sirfetch’d

ufodynasty:
Not sure if this will work, its worth a shot tho

### DIFF
--- a/src/tcgwars/logic/impl/gen8/RebelClash.groovy
+++ b/src/tcgwars/logic/impl/gen8/RebelClash.groovy
@@ -2411,21 +2411,9 @@ public enum RebelClash implements LogicCardInfo {
           onAttack {
             damage 180
             cantUseAttack(thisMove, self, 10000)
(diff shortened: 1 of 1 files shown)


src/tcgwars/logic/impl/gen8/RebelClash.groovy
```diff
@@ -2410,21 +2410,22 @@ public enum RebelClash implements LogicCardInfo {
           attackRequirement {}
           onAttack {
             damage 180
+            cantUseAttack(thisMove, self, 10000)
```
luongthomas, line 2413: We can remove the commented out text here, it'll still be in git history if we need it

luongthomas: Maybe add a TODO tag instead

luongthomas:
[GALARIAN_SIRFETCH_D_95](https://trello.com/c/xJpxYagu/316-galariansirfetchd95)

PR #195, opened 2020-05-07
Ninetales V

ufodynasty:
Trying new ways of reseting the energy cost


luongthomas:
This one looks promising

PR #211, opened 2020-05-11
Grimsley up to 3 counters

Najzeko:
Restructured Grimsley to allow for moving of up to 3 damage counters

### DIFF
--- a/src/tcgwars/logic/impl/gen7/UnifiedMinds.groovy
+++ b/src/tcgwars/logic/impl/gen7/UnifiedMinds.groovy
@@ -4419,12 +4419,21 @@ public enum UnifiedMinds implements LogicCardInfo {
         return supporter (this) {
           text "Move up to 3 damage counters from 1 of your opponent’s Pokémon to another of their Pokémon."
           onPlay {
-            def pcs = opp.all.findAll {it.numberOfDamageCounters}.select("Move damage counters from")
-            def tar = opp.all.findAll {it != pcs}.select("To?")
-            def num = Math.min(3, pcs.numberOfDamageCounters)
-            pcs.damage -= hp(10*num)
-            tar.damage += hp(10*num)
-            bc "Moved $num damage counters from $pcs to $tar"
+            def src = opp.all.findAll {it.numberOfDamageCounters}.select("Move damage counters from?")
+            def dest = opp.all.findAll {it != src}.select("To?")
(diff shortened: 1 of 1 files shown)


ufodynasty:
As long as you are changing this, could you use directDamage instead of changing the hp value so  abilities that block damage counters from being placed would properly prevent this?

Najzeko:
Done

src/tcgwars/logic/impl/gen7/UnifiedMinds.groovy
```diff
@@ -4419,12 +4419,22 @@ public enum UnifiedMinds implements LogicCardInfo {
...
-            def num = Math.min(3, pcs.numberOfDamageCounters)
-            pcs.damage -= hp(10*num)
-            tar.damage += hp(10*num)
-            bc "Moved $num damage counters from $pcs to $tar"
+            def src = opp.all.findAll {it.numberOfDamageCounters}.select("Move damage counters from?")
+            def dest = opp.all.findAll {it != src}.select("To?")
+            src.damage -= hp(10)
+            dest.damage += hp(10)
+            bc "moved one damage counter"
+            2.times {
```
luongthomas, line 4427: The syntax for looping a fixed number of times is:

```
(1..x).each {
   bc "Iteration number $it"
   doThing()
}
```

Where `x` is the number of times to loop

src/tcgwars/logic/impl/gen7/UnifiedMinds.groovy
```diff
@@ -4419,12 +4419,22 @@ public enum UnifiedMinds implements LogicCardInfo {
...
-            def pcs = opp.all.findAll {it.numberOfDamageCounters}.select("Move damage counters from")
-            def tar = opp.all.findAll {it != pcs}.select("To?")
-            def num = Math.min(3, pcs.numberOfDamageCounters)
-            pcs.damage -= hp(10*num)
-            tar.damage += hp(10*num)
-            bc "Moved $num damage counters from $pcs to $tar"
+            def src = opp.all.findAll {it.numberOfDamageCounters}.select("Move damage counters from?")
+            def dest = opp.all.findAll {it != src}.select("To?")
+            src.damage -= hp(10)
+            dest.damage += hp(10)
```
ufodynasty, line 4425: This is still changing the hp directly rather than using directDamage. You got it correct inside the loop.

PR #221, opened 2020-05-12
Sunflora test

ufodynasty:
If this works next commit will fix ttgx bug and unregister the effect


luongthomas:
I wonder if ef.move needs to be casted as Move.  probably not though

PR #223, opened 2020-05-12
Final Sunflora

ufodynasty:
The effect works as intended including ttgx costs. It can even be copied by copying attacks properly.


ufodynasty:
Also huge thanks to lithogen for finding ef.move. Wouldn’t be possible without that.

PR #231, opened 2020-05-12
Implement Aggron ability

ufodynasty:


src/tcgwars/logic/impl/gen3/PowerKeepers.groovy
```diff
@@ -213,8 +213,10 @@ public enum PowerKeepers implements LogicCardInfo {
         pokePower "Terraforming", {
           text "Once during your turn (before your attack), you may look at the top 5 cards from your deck and put them back on top of your deck in any order. This power can't be used if Aggron is affected by a Special Condition."
           actionA {
-            // TODO
             checkNoSPC()
```
luongthomas, line 216: We should add a checkLastTurn() here too

luongthomas: It would also be good to have an assert my.deck : "Deck is empty"

src/tcgwars/logic/impl/gen3/PowerKeepers.groovy
```diff
@@ -213,8 +213,12 @@ public enum PowerKeepers implements LogicCardInfo {
         pokePower "Terraforming", {
           text "Once during your turn (before your attack), you may look at the top 5 cards from your deck and put them back on top of your deck in any order. This power can't be used if Aggron is affected by a Special Condition."
           actionA {
-            // TODO
+            assert my.deck : “Your deck is empty.”
```
luongthomas, line 216: I wonder why the `Your` and `.` is red.  Is it because of the type of quotes you're using?

PR #232, opened 2020-05-12
Fix build

ufodynasty:
Since when could you not include punctuation in strings?


src/tcgwars/logic/impl/gen3/PowerKeepers.groovy
```diff
@@ -213,7 +213,7 @@ public enum PowerKeepers implements LogicCardInfo {
         pokePower "Terraforming", {
           text "Once during your turn (before your attack), you may look at the top 5 cards from your deck and put them back on top of your deck in any order. This power can't be used if Aggron is affected by a Special Condition."
           actionA {
-            assert my.deck : “Your deck is empty.”
+            assert my.deck : “Your deck is empty”
```
luongthomas, line 216: You can use punctuation, but I think your quotes are the wrong ones

luongthomas: “ vs "

PR #253, opened 2020-05-13
Altaria EX check fix

ufodynasty:
it.EX is for pokemon ex it.POKEMON_EX is for pokemon EX. Dont you love the naming conventions?


luongthomas:
yeah this tripped me up when I started working on ex sets.  I forgot to go back and change the initial sets I worked on

PR #256, opened 2020-05-13
Implement anorith_29 attack

ufodynasty:


src/tcgwars/logic/impl/gen3/LegendMaker.groovy
```diff
@@ -1119,9 +1119,14 @@ public enum LegendMaker implements LogicCardInfo {
         move "Influence", {
           text "Search your deck for Omanyte, Kabuto, Aerodactyl, Lileep, or Anorith and put up to 2 of them onto your Bench. Shuffle your deck afterward. Treat the new Benched Pokémon as Basic Pokémon."
           energyCost C
-          attackRequirement {}
+          attackRequirement {assert my.deck : "You have no cards in your deck"}
           onAttack {
-            // TODO
+            def maxSpace = Math.min(my.bench.freeBenchCount, 2)
+            my.deck.search(max:maxSpace, "Search your deck for up to 2 cards named Omanyte, Kabuto, Aerodactyl, Lileep, or Anorith", {it.name == "Omanyte" || it.name == "Kabuto" || it.name == "Aerodactyl" || it.name == "Lileep" || it.name == "Anorith"}).each {
```
luongthomas, line 1128: I wonder if we have to add min:0?

luongthomas: to prevent softlock if there are none in deck

PR #265, opened 2020-05-14
Fix Electrode EX FRLG 107

Najzeko:
Fix so it can no longer attach energies to EX pokemon

### DIFF
--- a/src/tcgwars/logic/impl/gen3/FireRedLeafGreen.groovy
+++ b/src/tcgwars/logic/impl/gen3/FireRedLeafGreen.groovy
@@ -2523,11 +2523,11 @@ public enum FireRedLeafGreen implements LogicCardInfo {
               checkLastTurn()
               powerUsed()
               new Knockout(self).run(bg)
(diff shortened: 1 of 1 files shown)


src/tcgwars/logic/impl/gen3/FireRedLeafGreen.groovy
```diff
@@ -2523,11 +2523,11 @@ public enum FireRedLeafGreen implements LogicCardInfo {
               checkLastTurn()
               powerUsed()
               new Knockout(self).run(bg)
```
luongthomas, line 2525: You could go with this approach with the name but you'd have to modify it to say `!it.name.contains(" ex ")` 

Another approach is just to change it to check for the Pokemon SubType by using `!it.topPokemonCard.cardTypes.is(EX)`

Najzeko:
Done

ufodynasty:
Just for future reference, the code in its original state could have worked if it.pokemonEX was replaced by it.EX. I only learned this recently but it.EX refers to pokemon ex and it.pokemonEX refers to pokemon EX. Hope this helps. :D

PR #276, opened 2020-05-15
Charizard delta 4

ufodynasty:
Allows multiple activations of different peal of thunders


ufodynasty:
I was able to successfully evolve from charmeleon delta.

PR #288, opened 2020-05-16
Slacking ability block fix

ufodynasty:


src/tcgwars/logic/impl/gen3/PowerKeepers.groovy
```diff
@@ -534,12 +534,24 @@ public enum PowerKeepers implements LogicCardInfo {
         weakness F
         pokeBody "Lazy", {
           text "As long as Slaking is your Active Pokémon, your opponent's Pokémon can't use any Poké-Powers."
-          getterA (IS_ABILITY_BLOCKED) { Holder h ->
```
luongthomas, line 537: is eff3 used anywhere?

ufodynasty: no it is not

PR #337, opened 2020-05-20
Holon’s castform

ufodynasty:
added energy check and nothing else at all ;)


src/tcgwars/logic/impl/gen3/HolonPhantoms.groovy
```diff
@@ -250,7 +250,7 @@ public enum HolonPhantoms implements LogicCardInfo {
               before BETWEEN_TURNS, {
                 if (turnCount + 1 <= bg.turnCount) {
                   if (all.contains(pcs)) {
-                    bc "Harsh Fluid activates"
```
luongthomas, line 253: Hmmm

PR #339, opened 2020-05-20
Tool Scrapper qol fix

ufodynasty:
Now has you choose the controler of the pokemon you want to remove a tool from.


luongthomas:
def tar

PR #341, opened 2020-05-20
Tool scrapper qol extended

ufodynasty:
Allows the player to backtrack out of their decisions and stop more easily


luongthomas:
GLHF

PR #361, opened 2020-05-21
Fix Buizel UL-45 Muddy Water empty opponent bench

flagrama:


src/tcgwars/logic/impl/gen4/Unleashed.groovy
```diff
@@ -1143,7 +1143,9 @@ public enum Unleashed implements LogicCardInfo {
             energyCost W
             onAttack {
               damage 10
-              damage 10, opp.bench.select()
+              if opp.bench {
```
luongthomas, line 1146: Could you wrap the `if` condition in parenthesis?  I believe it won't build otherwise

src/tcgwars/logic/impl/gen4/Unleashed.groovy
```diff
@@ -1143,7 +1143,9 @@ public enum Unleashed implements LogicCardInfo {
...
+                damage 10, opp.bench.select()
```
luongthomas, line 1147: It's also helpful to add a `string` argument in `select()` to help the user what they're selecting and why

src/tcgwars/logic/impl/gen4/Unleashed.groovy
flagrama: Sorry, was having issues with git, will fix.

flagrama:
Hmm, there seems to be a lot of strings with both "Pokemon" and "Pokémon". Is one of them preferred?

luongthomas:
> Hmm, there seems to be a lot of strings with both "Pokemon" and "Pokémon". Is one of them preferred?

Pokémon is preferred but I'm okay either way

flagrama:
I'll keep that in mind for future PRs.

PR #365, opened 2020-05-21
unown 123

ufodynasty:


src/tcgwars/logic/impl/gen3/UnseenForces.groovy
```diff
@@ -3203,14 +3203,15 @@ public enum UnseenForces implements LogicCardInfo {
             assert opp.bench.notFull : "Opponent's Bench is full"
           }
           onAttack {
-            if (opp.hand.findAll { it.cardTypes.is(BASIC) }) {
+            opp.hand.showToMe("Opponent's hand.")
+            if (opp.hand.findAll { it.cardTypes.is(BASIC) } && opp.bench.notFull) {
```
luongthomas, line 3207: I think the `opp.bench.notFull` is redundant here since it is already an assert in `attackRequirement`

PR #367, opened 2020-05-21
Solrock_25 check

ufodynasty:
Shouldn’t crash the game but im working with code I dont understand


luongthomas:
> Shouldn’t crash the game but im working with code I dont understand
You and me both

PR #369, opened 2020-05-21
azurill 20

ufodynasty:
method stolen from mew *


src/tcgwars/logic/impl/gen3/DeltaSpecies.groovy
```diff
@@ -861,9 +861,9 @@ public enum DeltaSpecies implements LogicCardInfo {
             def chosenEnergy = my.hand.filterByType(BASIC_ENERGY).select(min:0, max: 1, "Select a Basic Energy card to search for a Pokemon of that Type.")
 
             if (chosenEnergy) {
-              def cardType = chosenEnergy.asEnergyCard().type
+              def cardType = chosenEnergy.basicType
               my.deck.search(count: 1, "search for a Basic Pokémon or Evolution card", {
-                it.cardTypes.is(POKEMON) && it.asPokemonCard().types.contains(chosenType)
```
luongthomas, line 866: Dang it looks like my chosenType variable wasn't even defined

luongthomas: Big oof for me

PR #377, opened 2020-05-22
Unown 127

flagrama:
Status effects are applied, not damage

Will `apply` correctly target the new defending Pokemon? Or will it try to apply to the old one? I'm assuming `sw` will have set opp().active to the new one already.

rant: Also, I know this is accurate to what is on the card, but this card text is confusing, since when is there more than one defending Pokemon in the TCG? Either there is a mechanic I was never aware of, or Nintendo was bad at writing card text.


luongthomas:
Double battles were introduced during the EX days and thus there could be multiple defending Pokemon.  For implementation’s sake, we can assume tcgone will not support double battles.

PR #384, opened 2020-05-22
Rename druddigon DRM 45

Najzeko:
Renamed from Druggidon to Druddigon


luongthomas:
I will close this PR because it is not possible to change the card ID of a card that has already been released into production as this will crash the deckbuilder on play.tcgone.net

PR #421, opened 2020-05-24
Revert isPokemonDelta

ufodynasty:


axpendix:
Why did you revert this @ufodynasty

luongthomas:
> Why did you revert this @ufodynasty

CI was having trouble building after this method was merged

axpendix:
I see, it's due to casting of `TcgStatics.bg().em().retrieveObject("Holon_Veil")`, it should be cast, and I also don't recommend using list<player_type> object. Use the PlayerType as a suffix to the key. Increment the value when the ability is activated, and decrement it when it's deactivated. Then, on the new method, check whether it is greater than zero.

axpendix:
```
onActivate{
  key = "holon_veil_"+self.owner
  bg.em().retrieveAndStore(key, {it ? it+1: 1})
}
onDeactivate{
  key = "holon_veil_"+self.owner
  bg.em().retrieveAndStore(key, {it-1})
}
isDelta() {
...
  Integer value = (Integer) Battleground.getInstance().em().retrieveObject(key);
  if(value != null && value > 0) {
    ...
  }
}
```

axpendix:
@ufodynasty

PR #422, opened 2020-05-24
Hardcode Holon_Veil into is

ufodynasty:
This feels like way to deep of a fix for a single card. Still hoping for a review from @axpendix . 
The goal of this change is that the delta check will return true even if the card isn’t a delta card as long as ampharos delta’s ability is active. I store the list of playerTypes for whom Holon Veil is active in the stored object Holon_Veil.
Marking as a draft so it doesn’t get merged by accident again.

### DIFF
--- a/src/tcgwars/logic/util/CardTypeSet.java
+++ b/src/tcgwars/logic/util/CardTypeSet.java
@@ -34,7 +34,7 @@ public class CardTypeSet extends TreeSet<CardType>{
   }
   public boolean is(CardType o){
     List<PlayerType> Holon_Veil = TcgStatics.bg().em().retrieveObject("Holon_Veil");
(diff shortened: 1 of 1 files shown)


src/tcgwars/logic/util/CardTypeSet.java
```diff
@@ -31,7 +33,8 @@ public CardTypeSet(CardType... cardTypes) {
     }
   }
   public boolean is(CardType o){
-    return contains(o);
+    List<PlayerType> Holon_Veil = TcgStatics.bg().em().retrieveObject("Holon_Veil");
+    return (contains(o) || (o == CardType.DELTA && contains(CardType.POKEMON) && Holon_Veil.contains(player)));
```
ufodynasty, line 37: I know that player doesnt exist within cardtype but am unsure how else to access the owner of a card that is not in a pcs

PR #424, opened 2020-05-24
Mark as basic

flagrama:
This reverts the changes I've made so far. According to @axpendix this is not the correct way to go about having the Pokemon cards put into play by these effects be considered a Basic Pokemon. Apparently, a Pokemon card in play without another card under it should already be considered basic. However, the way other cards check if a Pokemon card is basic doesn't seem to line up with this. I'm going to need to figure out what the best way to handle this actually is.

### DIFF
--- a/src/tcgwars/logic/impl/gen3/PowerKeepers.groovy
+++ b/src/tcgwars/logic/impl/gen3/PowerKeepers.groovy
@@ -1608,15 +1608,7 @@ public enum PowerKeepers implements LogicCardInfo {
           energyCost W
           attackRequirement {}
           onAttack {
-            if(my.deck){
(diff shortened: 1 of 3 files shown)


src/tcgwars/logic/impl/gen3/PowerKeepers.groovy
```diff
@@ -1608,11 +1608,9 @@ public enum PowerKeepers implements LogicCardInfo {
           energyCost W
           attackRequirement {}
```
axpendix, line 1609: This attack should have 2 requirements: bench not full and deck not empty, it's better to put these in the attackRequirement clause like this: 

```suggestion
          attackRequirement {
              assert my.bench.notFull : "Bench is full"
              assert my.deck : "Deck is empty"
          }
```

and remove `if(my.deck)` inside `onAttack`.

PR #426, opened 2020-05-25
Holon’s Castform

ufodynasty:
modified charjabug implementation from unb. Can’t test this yet but I think it will work.


flagrama:
Shouldn't this use `new CustomCardInfo(HOLON_S_CASTFORM_44)` instead?

And is `if(choose([1,2], ["Pokémon", "Energy"], "Play this card as a Pokémon or as an energy?") == 2 [...]` going to ask you if you want to play as a Pokemon or Energy every time? If so, doesn't `[...] && bg.em().retrieveObject("Castform_Energy") != bg.turnCount)` mean that you can end up choosing Energy, but it will play down as a Pokemon instead? I'd think the second check would be better as an assert assuming they break out of an effect entirely. If not that, at least reversed so that the "already used" check could shortcircuit the logic and play as a Pokemon without asking for the choice.

PR #432, opened 2020-05-26
Revert "Optimized Perfection"

ufodynasty:
Didn’t work. I don’t know any other way to compare moves within lists that would work as moves with the same name can have different effects.


flagrama:
Is the main issue duplicate Pokemon cards duplicating the same move in the list? If so, maybe you could use a second list with the card names that you've already checked, either using the `toString()` method, the `getEnumName()` method, or some other way to check uniqueness of the Pokemon card with it's name and set number before adding a move to the move list?

Or maybe use a Map/Dictionary in a similar manner with the Key being the Pokemon's name using one of the previous methods, and the value being the list of moves they have. Might complicated the logic because then you have to iterate over each value in the list for each key/value pair.

PR #437, opened 2020-05-28
Moves list reset timing

ufodynasty:
If this works I won’t understand why but at this point im reaching my limit.


ufodynasty:
I got it working and didn’t even realize

PR #446, opened 2020-05-29
Clefable 74 RCL - Prankish not only once per turn

flagrama:
Prankish should not be limited to only once per turn. It should trigger
every time the Clefable is played from the hand to evolve a Pokemon.

This should fix
https://forum.tcgone.net/t/br-clefable-rcl-75-you-should-be-able-to-use-more-than-one-p/9930

Note: The text attribute for the ability states "Once during your turn," but [the actual card does not](https://tcgone.net/cards/clefable-rcl-75). Should this be corrected as well?


flagrama:
Retreat cost is also wrong. Is 1, should be 2. I'll correct that too I suppose?

PR #456, opened 2020-05-30
DP: Fixed Dialga (1), Implemented 2-50

starg09:
Electivire (DP 3) is partially implemented, missing a way to check for Elekid as a pre-evolution. This will be needed for multiple cards in this set and future ones. **(Edit: Testing a potential solution on Roserade, just added to the PR)**

Dialga (DP 1) should now return all energy properly. And Magnezone (DP 8) should be able to evolve (had Magneton*e* as its pre-evolution)

Also Fixed Huntail (GE 42) and Milotic (CIN 27).

### DIFF
--- a/src/tcgwars/logic/impl/gen4/DiamondPearl.groovy
+++ b/src/tcgwars/logic/impl/gen4/DiamondPearl.groovy
@@ -700,7 +700,8 @@ public enum DiamondPearl implements LogicCardInfo {
             energyCost G, G, C, C
             attackRequirement {}
             onAttack {
-              damage 0
+              damage 50
+              removeDamageCounterEqualToDamageDone()
             }
           }
 
@@ -713,7 +714,9 @@ public enum DiamondPearl implements LogicCardInfo {
             energyCost C
             attackRequirement {}
             onAttack {
-              damage 0
+              heal 40, self
+              clearSpecialCondition(self)
+              apply ASLEEP, self
             }
           }
           move "Take Down", {
@@ -721,7 +724,8 @@ public enum DiamondPearl implements LogicCardInfo {
             energyCost C, C, C
             attackRequirement {}
             onAttack {
-              damage 0
+              damage 60
+              damage 10, self
             }
           }
 
@@ -735,7 +739,9 @@ public enum DiamondPearl implements LogicCardInfo {
             energyCost G, C
             attackRequirement {}
             onAttack {
-              damage 0
+              damage 30
+              if(defending.getRemainingHP() < self.getRemainingHP())
+                damage 30
             }
           }
           move "Wring Out", {
@@ -743,7 +749,11 @@ public enum DiamondPearl implements LogicCardInfo {
             energyCost G, G
             attackRequirement {}
             onAttack {
-              damage 0
+              damage 20
+              flip {
+                apply PARALYZED
+                discardDefendingEnergy()
+              }
             }
           }
 
@@ -756,7 +766,7 @@ public enum DiamondPearl implements LogicCardInfo {
             energyCost C, C
             attackRequirement {}
             onAttack {
-              damage 0
+              flip 2, { damage 30 }
             }
           }
           move "Metronome", {
@@ -764,7 +774,17 @@ public enum DiamondPearl implements LogicCardInfo {
             energyCost C, C, C
             attackRequirement {}
             onAttack {
-              damage 0
+              //TODO: Maybe generalize into a metronome() or exMetronome() method? Taken from FRLG CLEFABLE_EX_106
+              def moveList = []
+              def labelList = []
+
+              moveList.addAll(defending.topPokemonCard.moves);
+              labelList.addAll(defending.topPokemonCard.moves.collect{it.name})
+
+              def move=choose(moveList, labelList)
+              def bef=blockingEffect(ENERGY_COST_CALCULATOR, BETWEEN_TURNS)
+              attack (move as Move)
+              bef.unregisterItself(bg().em())
             }
           }
 
@@ -786,7 +806,8 @@ public enum DiamondPearl implements LogicCardInfo {
             energyCost P, P, C
             attackRequirement {}
             onAttack {
-              damage 0
+              damage 50
+              apply POISONED
             }
           }
 
@@ -800,15 +821,22 @@ public enum DiamondPearl implements LogicCardInfo {
             energyCost C
             attackRequirement {}
             onAttack {
-              damage 0
+              def tar = my.discard.findAll{it.cardTypes.is(SUPPORTER) || it.cardTypes.is(POKEMON)}
+              tar.select(min: 0, max: 5,"Choose up to 5 in any combination of Pokémon and Supporter cards to put back into your deck").moveTo(my.deck)
+              shuffleDeck()
             }
           }
           move "Explosive Smoke", {
-            text "60 damage. ."
+            text "60 damage. Does 10 damage to each Benched Pokémon (both yours and your opponent’s). (Don’t apply Weakness and Resistance for Benched Pokémon.)."
             energyCost P, P, P
             attackRequirement {}
             onAttack {
-              damage 0
+              damage 60
+              all.each {
+                if (it.benched) {
+                  damage 10, it
+                }
+              }
             }
           }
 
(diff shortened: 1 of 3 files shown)


src/tcgwars/logic/impl/gen4/DiamondPearl.groovy
```diff
@@ -521,19 +616,23 @@ public enum DiamondPearl implements LogicCardInfo {
...
           move "Accelerative Dive", {
-            text "30 damage. until the end of your next turn."
+            text "30 damage. Flip a coin. If tails, this attack does nothing. If heads, prevent all damage done to Staraptor by attacks (both yours and your opponent’s) until the end of your next turn."
             energyCost C
             attackRequirement {}
             onAttack {
-              damage 0
+              damage 30
+              flip {
+                //TODO: preventAllDamageNextTurn() won't cut it here sadly.
```
flagrama, line 625: What if you reapply it on the `BEGIN_TURN` effect for your opponent as a one off delayed effect?

src/tcgwars/logic/impl/gen4/DiamondPearl.groovy
```diff
@@ -567,15 +668,19 @@ public enum DiamondPearl implements LogicCardInfo {
...
             text "40+ damage. If Azumarill has 3 or more Energy attached to it, this attack does 40 damage plus 20 more damage. If Azurill is anywhere under Azumarill, flip a coin. If heads, the Defending Pokémon is now Paralyzed."
             energyCost W, W
             attackRequirement {}
             onAttack {
-              damage 0
+              damage 40
+              if(self.cards.energyCount(W) >= 3){
+                damage 20
+              }
+              //TODO: Find out how to check for Azurill specifically (See Electivire/Roserade).
```
flagrama, line 683: > (See Electivire/Roserade)

Same :)

src/tcgwars/logic/impl/gen4/DiamondPearl.groovy
```diff
@@ -825,17 +988,30 @@ public enum DiamondPearl implements LogicCardInfo {
...
             }
           }
           move "Flop", {
             text "30 damage. You may switch Lopunny with 1 of your Benched Pokémon."
             energyCost C, C, C
             attackRequirement {}
             onAttack {
-              damage 0
+              damage 30
+              //TODO: This could definitely be generalized into a method, there's a couple dozens of cards with this optional effect across the DPPt era.
```
flagrama, line 1009: `switchYourActive()` like Archeops in UnifiedMinds but with a `confirm`? You can probably add something slightly altered to `TcgStatics.groovy`

src/tcgwars/logic/impl/gen4/DiamondPearl.groovy
```diff
@@ -910,14 +1101,16 @@ public enum DiamondPearl implements LogicCardInfo {
             attackRequirement {}
             onAttack {
               damage 0
+              //TODO
```
flagrama, line 1104: DeltaSpecies Slowking's ability looks to work the same way.

src/tcgwars/logic/impl/gen4/DiamondPearl.groovy
```diff
@@ -929,17 +1122,22 @@ public enum DiamondPearl implements LogicCardInfo {
           move "Minor Errand-Running", {
             text "Search your deck for a basic Energy card, show it to your opponent, and put it into your hand. Shuffle your deck afterward."
             energyCost C
-            attackRequirement {}
+            attackRequirement {
+              assert my.deck
+            }
             onAttack {
-              damage 0
+              //TODO: Convert this attack effect into a method, used in multiple gen 3/4 cards
```
flagrama, line 1129: `attachEnergyFrom(basic: true, my.deck, self)`
defined in `TcgStatics.groovy`

src/tcgwars/logic/impl/gen4/DiamondPearl.groovy
```diff
@@ -973,15 +1172,18 @@ public enum DiamondPearl implements LogicCardInfo {
...
             }
           }
           move "Ease Up", {
             text "40+ damage. Snorlax is now Asleep. If Snorlax is evolved from Munchlax, this attack does 40 damage plus 30 more damage."
             energyCost C, C, C, C
             attackRequirement {}
             onAttack {
-              damage 0
+              damage 40
+              //TODO: Baby Evolution check (see Electivire/Roserade)
```
flagrama, line 1184: Leaving a comment here this doesn't get lost from the previous related ones as well.

src/tcgwars/logic/impl/gen4/DiamondPearl.groovy
```diff
@@ -462,15 +531,20 @@ public enum DiamondPearl implements LogicCardInfo {
...
             }
           }
           move "Spike Whip", {
             text "50 damage. If Budew is anywhere under Roserade, choose 1 of your opponent’s Benched Pokémon. This attack does 30 damage to that Pokémon."
             energyCost G, G, C
             attackRequirement {}
             onAttack {
-              damage 0
+              damage 50
+              //TODO: Find out how to check for Budew specifically.
```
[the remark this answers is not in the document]
flagrama: Ah, I see, this was a later change.

starg09:
> I'mma just look through this for fun. Nothing I mention here is required afaik, and possibly could be bad form so should probably confirm with one of the devs before changing it.

Sounds good, I'm online though and would want to fix a couple of them so please give me a couple minutes before merging/closing! 😄

flagrama:
Don't worry, I don't have that power :laughing:

src/tcgwars/logic/impl/gen4/DiamondPearl.groovy
```diff
@@ -480,16 +554,31 @@ public enum DiamondPearl implements LogicCardInfo {
           weakness F, PLUS30
           resistance P, MINUS20
           pokePower "Darkness Fan", {
-            text "Once during your turn , if Shiftry is your Active Pokémon, you may flip a coin. If heads, choose 1 Evolved Pokémon on your opponent’s Bench, remove the highest Stage Evolution card from that Pokémon, and put it back into his or her hand. This power can’t be used if Shiftry is affected by a Special Condition."
+            text "Once during your turn (before your attack), if Shiftry is your Active Pokémon, you may flip a coin. If heads, choose 1 Evolved Pokémon on your opponent’s Bench, remove the highest Stage Evolution card from that Pokémon, and put it back into his or her hand. This power can’t be used if Shiftry is affected by a Special Condition."
             actionA {
+              checkLastTurn()
+              assert self.active : "$self is not your active Pokémon."
+              assert !(self.specialConditions) : "$self is affected by a Special Condition"
```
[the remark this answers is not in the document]
starg09: Ahh, not sure why I put that assert there :P
It was a bit late at night and I was still getting used to the different methods available I guess, haha (Seems like it was used on a couple Deoxys, Emerald and TRR cards; will commit a fix for those later!)

src/tcgwars/logic/impl/gen4/DiamondPearl.groovy
```diff
@@ -480,16 +554,31 @@ public enum DiamondPearl implements LogicCardInfo {
...
+              def list = opp.bench.findAll { it.evolution }
+              assert list : "Your opponent has no evolved Pokémon."
+              powerUsed()
+              flip {
+                def pcs = list.select("Devolve one of your opponent's evolved Benched Pokémon.")
+                assert pcs
```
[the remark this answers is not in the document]
starg09: Yup, agree looking at it in more detail. Will remove it from UNM's CELEBI_4 later as well...

One thing I want to check about this is how it'd interact with non-evolved evolutions. I imagine it distinguishes but it may be needed to add a separate `it.evolved` attribute for it to work properly.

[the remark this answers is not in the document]
flagrama: That's something I'm not totally sure of after discussing the fossil Pokemon as basics due to some of the gen 3 cards with axpendix. I think `.evolution` checks if the Pokemon is "evolved" which means there is another PCS underneath it. You may be able to directly call the correct evolution call in `PokemonCardSet.java` on a PCS directly.

src/tcgwars/logic/impl/gen4/DiamondPearl.groovy
starg09: Aha! It seems to be implemented already in that command itself, thanks a lot <3

starg09:
Alright, I'm gonna stop here for now, wanna create a new local branch for future DP-related commits (so I can submit potential fixes while working on that). So I won't add any more implementations on this PR. If it could be checked out so I can test these (specially Dialga!) on dev I'd appreciate it 👍

PR #458, opened 2020-05-30
Clefable RCL 75 - Correct ability text in log

flagrama:
I over-removed when fixing the ability. Without powerUsed the "used
ability" text doesn't seem to appear in the game log, which is certainly
problematic.

Whoops.


flagrama:
Also, I suppose that any ability that doesn't display text when it is used is technically a bug?

PR #467, opened 2020-05-30
Fix build maybe

ufodynasty:
The metronome code looked good so im guessing there was a missing import. It wasn’t obvious to me which import was missing so I included everything that was in pokemod that wasn’t in this as the same code works properly there.


starg09:
Ok that works. Was about to PR myself, but with only DET (which copies attacks in a similar fashion to metronome, and only 4 imports)

PR #472, opened 2020-05-31
More DP Fixes and Baby Check Implementation on 3, 18 & 37

starg09:
Electivire specifically will need testing, see if the Poké-Power works consistently.

Fixes:

* Missing `def` on Lopunny (10) 
* Lopunny (10) now does bench damage on Flop.
* Medicham (32) now puts damage counters on bench instead of plain damage.
* Munchlax (33) gives a proper warning, and actually discards 2 cards before attacking.
* Noctowl (34) should now place the new prize facing up.
* Wobbuffet (35)... Hopefully this works, as of rn it ignored cards like Metal Goggles (TEU)
* Cascoon (44) throws an error when trying to give through [DEV], hoping this fixes it.
* Missing text on Grotle (49)
* Haunter (50) now does the "Defending Pokémon is Asleep" check on `attackRequirement`. Most attacks validate that way (and impede attacking for nothing), so switched it up.

### DIFF
--- a/src/tcgwars/logic/impl/gen4/DiamondPearl.groovy
+++ b/src/tcgwars/logic/impl/gen4/DiamondPearl.groovy
@@ -291,7 +291,13 @@ public enum DiamondPearl implements LogicCardInfo {
             text "As often as you like during your turn (before your attack), if Elekid is anywhere under Electivire, you may move a [L] Energy attached to your Pokémon to Electivire. This power can’t be used if Electivire is affected by a Special Condition."
             actionA {
               checkNoSPC()
-              //TODO: How to check for Elekid specifically? Can't reuse gen 3's "Stages of Evolution" implementation, since that's used on Electabuzz as a basic.
+              assert self.cards.findAll {it.name.contains("Elekid")} : "Elekid is not found under $self, you can't use this Poké-Power"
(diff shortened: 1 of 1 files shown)


src/tcgwars/logic/impl/gen4/DiamondPearl.groovy
```diff
@@ -1300,7 +1300,10 @@ public enum DiamondPearl implements LogicCardInfo {
             onAttack {
               flip {
                 targeted (defending) {
-                  defending.damage += self.damage
+                  //defending.damage += self.damage
+                  directDamage self.damage, defending
+                  def dc = self.damage / 10
+                  bc "Moved $dc damage counters from $self to $defending."
```
flagrama, line 1310: I think `directDamage` should already state how much damage it did to the defending Pokemon. Probably better repurposed to mention that it heals Wobbuffet

starg09: Yeah, it seems to do. I'm rewriting this entirely though, since wobbuffett actually moves the counters (so it doesn't count as healing, checking at some rulings).

Need to check on metal goggles, since it doesn't seem to work with the moving counters way however... (checked on Xerneas GX)

PR #537, opened 2020-06-03
Implemented Dialga 4's Adamant Orb held item

Grayahm:
Credit to starg09 :)


flagrama:
I notice that it states explicitly attacks from Dialga, but it doesn't look like the if statement checking the conditions actually checks that `it.from == self` and perhaps `self.active` to ensure that. As it is, I believe this would apply to all Pokémon, not just Dialga.

Grayahm:
Fixed

PR #551, opened 2020-06-03
MT: Implemented 68-97, fixes on 111

starg09:
TODO:
* Write the check for Vigoroth (68) to increase the damage in "Wake-Up Punch".
Fixed:
* Searches should now work correctly on Fossil Excavator (111).

### DIFF
--- a/src/tcgwars/logic/impl/gen4/MysteriousTreasures.groovy
+++ b/src/tcgwars/logic/impl/gen4/MysteriousTreasures.groovy
@@ -1707,12 +1707,15 @@ public enum MysteriousTreasures implements LogicCardInfo {
       case BUIZEL_75:
         return basic (this, hp:HP060, type:WATER, retreatCost:1) {
           weakness L, PLUS10
-          move "", {
-            text "If damage. "
-            energyCost ()
-            attackRequirement {}
-            onAttack {
-              damage 0
+          //Chesto Berry: If Buizel is Asleep, remove the Special Condition Asleep from Buizel at the end of each player’s turn.
+          customAbility {
+            delayedA {
+              before BEGIN_TURN,{
+                if(self.isSPC(ASLEEP)) {
+                  bc "Chesto Berry activates"
+                  clearSpecialCondition(self, ATTACK, [ASLEEP])
+                }
+              }
             }
           }
           move "Quick Attack", {
@@ -1734,7 +1737,7 @@ public enum MysteriousTreasures implements LogicCardInfo {
             energyCost C, C
             attackRequirement {}
             onAttack {
-              damage 0
+            flip { preventAllEffectsNextTurn() }
             }
           }
           move "Double-edge", {
@@ -1757,7 +1760,7 @@ public enum MysteriousTreasures implements LogicCardInfo {
             energyCost ()
             attackRequirement {}
             onAttack {
-              damage 0
+              flip { apply ASLEEP }
             }
           }
           move "Jump On", {
@@ -1779,7 +1782,7 @@ public enum MysteriousTreasures implements LogicCardInfo {
             energyCost C
             attackRequirement {}
             onAttack {
-              damage 0
+              flip { opponentCantPlaySupporterNextTurn(delegate) }
             }
           }
           move "Finger Poké", {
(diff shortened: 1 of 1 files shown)


src/tcgwars/logic/impl/gen4/MysteriousTreasures.groovy
```diff
@@ -1607,15 +1611,20 @@ public enum MysteriousTreasures implements LogicCardInfo {
             energyCost P
             attackRequirement {}
             onAttack {
-              damage 0
+              my.deck.subList(0,5).select(min:0,"Choose a card to put in your hand").moveTo(hidden: true, my.hand)
```
flagrama, line 1614: Are you supposed to be able to choose 0? The text implies to me that you _must_ select 1.

src/tcgwars/logic/impl/gen4/MysteriousTreasures.groovy
```diff
@@ -1607,15 +1611,20 @@ public enum MysteriousTreasures implements LogicCardInfo {
             energyCost P
             attackRequirement {}
```
flagrama, line 1612: Probably needs a deck assertion too.

src/tcgwars/logic/impl/gen4/MysteriousTreasures.groovy
starg09: Yup, should be 1 indeed. Will correct it with the build fix 👍

src/tcgwars/logic/impl/gen4/MysteriousTreasures.groovy
```diff
@@ -1660,32 +1672,39 @@ public enum MysteriousTreasures implements LogicCardInfo {
...
-            text "Any damage. "
-            energyCost ()
-            attackRequirement {}
-            onAttack {
-              damage 0
+          //Wacan Berry: Any damage done to Bidoof by attacks from [L] Pokémon is reduced by 20 (after applying Weakness and Resistance).
+          customAbility {
+            delayedA {
+              before APPLY_ATTACK_DAMAGES, {
+                if(ef.attacker.owner != self.owner && (ef.attacker.types.contains(R) || ef.attacker.types.contains(W))) {
```
flagrama, line 1691: Shouldn't this just be checking for `L` type?

src/tcgwars/logic/impl/gen4/MysteriousTreasures.groovy
```diff
@@ -1660,32 +1672,39 @@ public enum MysteriousTreasures implements LogicCardInfo {
...
+                  bg.dm().each{
+                    if(it.to == self && it.notNoEffect && it.dmg.value) {
+                      bc "Chesto Berry -20"
```
flagrama, line 1694: I'm guessing this whole thing was copy and pasted from another card and you forgot to change it :laughing:

src/tcgwars/logic/impl/gen4/MysteriousTreasures.groovy
starg09: Ah, forgot to edit the check whoops. Thanks!

src/tcgwars/logic/impl/gen4/MysteriousTreasures.groovy
starg09: Replied above 😝

src/tcgwars/logic/impl/gen4/MysteriousTreasures.groovy
```diff
@@ -1993,7 +2044,8 @@ public enum MysteriousTreasures implements LogicCardInfo {
             energyCost W, W
             attackRequirement {}
             onAttack {
-              damage 0
+              //TODO: Modularize
```
flagrama, line 2047: Something like

```
def doDamage = true
flip 2, {}, {doDamage = false}
if(doDamage) damage 120
```
should work as well, for when you modularize this, right?

src/tcgwars/logic/impl/gen4/MysteriousTreasures.groovy
```diff
@@ -2024,11 +2074,11 @@ public enum MysteriousTreasures implements LogicCardInfo {
         return basic (this, hp:HP050, type:PSYCHIC, retreatCost:1) {
           weakness P, PLUS10
           move "Call for Family", {
-            text "Search your deck for Nidoran Male or Nidoran Female and put it onto your Bench. Shuffle your deck afterward."
+            text "Search your deck for Nidoran ♂ or Nidoran ♀ and put it onto your Bench. Shuffle your deck afterward."
             energyCost C
 
-            // TODO
-            callForFamily(basic:true, 1, delegate)
+            //TODO: Check this finding all Nidorans, male and female.
```
flagrama, line 2080: This should work fine based of how `callForFamily` works

PR #553, opened 2020-06-03
Update GreatEncounters.groovy

Grayahm:
Implemented:
- Moonlight Stadium
- Felicity's Drawing
- Rare Candy (copied from EX Sandstorm)
- Baltoy
- Claydol

### DIFF
--- a/src/tcgwars/logic/impl/gen4/GreatEncounters.groovy
+++ b/src/tcgwars/logic/impl/gen4/GreatEncounters.groovy
@@ -115,7 +115,7 @@ public enum GreatEncounters implements LogicCardInfo {
   AMULET_COIN_97 ("Amulet Coin", 97, Rarity.UNCOMMON, [TRAINER]),
   FELICITY_S_DRAWING_98 ("Felicity's Drawing", 98, Rarity.UNCOMMON, [TRAINER]),
   LEFTOVERS_99 ("Leftovers", 99, Rarity.UNCOMMON, [TRAINER]),
-  MOONLIGHT_STADIUM_100 ("Moonlight Stadium", 100, Rarity.UNCOMMON, [STADIUM]),
+  MOONLIGHT_STADIUM_100 ("Moonlight Stadium", 100, Rarity.UNCOMMON, [TRAINER, STADIUM]),
   PREMIER_BALL_101 ("Premier Ball", 101, Rarity.UNCOMMON, [TRAINER]),
   RARE_CANDY_102 ("Rare Candy", 102, Rarity.UNCOMMON, [TRAINER]),
   CRESSELIA_LV_X_103 ("Cresselia LV.X", 103, Rarity.ULTRARARE, [LEVEL_UP, EVOLUTION, POKEMON, _PSYCHIC_]),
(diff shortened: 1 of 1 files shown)


src/tcgwars/logic/impl/gen4/GreatEncounters.groovy
```diff
@@ -1393,17 +1400,16 @@ public enum GreatEncounters implements LogicCardInfo {
           move "Psychic Balance", {
             text "If you have less cards in your hand than your opponent, draw cards until you have the same number of cards as your opponent. (If you have more or the same number of cards in your hand as your opponent, this attack does nothing.)"
             energyCost C
-            attackRequirement {}
             onAttack {
-              damage 0
```
luongthomas, line 1398: I haven't tested it yet but I wonder what the engine does when you do `draw -1`

luongthomas:
While you're in the file, could you also update line 118 for MOONLIGHT_STADIUM_100 to add the `STADIUM` subType?

src/tcgwars/logic/impl/gen4/GreatEncounters.groovy
Grayahm: How so?

luongthomas: The case when your hand size is larger than your opponent's hand size

src/tcgwars/logic/impl/gen4/GreatEncounters.groovy
```diff
@@ -469,8 +469,15 @@ public enum GreatEncounters implements LogicCardInfo {
           pokePower "Cosmic Power", {
             text "Once during your turn , you may choose up to 2 cards from your hand and put them on the bottom of your deck in any order. If you do, draw cards until you have 6 cards in you hand. This power can’t be used if Claydol is affected by a Special Condition."
             actionA {
-            }
-          }
+              checkNoSPC()
+              assert my.hand.notEmpty : "You have no cards in your hand."
```
luongthomas, line 473: `checkLastTurn()` could also be added here too

Grayahm: Fixed

src/tcgwars/logic/impl/gen4/GreatEncounters.groovy
```diff
@@ -2166,8 +2173,15 @@ public enum GreatEncounters implements LogicCardInfo {
         return basicTrainer (this) {
           text "You can play only one Supporter card each turn. When you play this card, put it next to your Active Pokémon. When your turn ends, discard this card.\nDiscard up to 2 cards from your hand. If you discard 1 card, draw 3 cards. If you discard 2 cards, draw 4 cards."
           onPlay {
-          }
-          playRequirement{
+            assert my.deck : "Your deck is empty"
+            assert my.hand.size() > 1 : "You can't play this card"
```
flagrama, line 2177: Should these assertions be playRequirements instead?

Grayahm: Yes!

src/tcgwars/logic/impl/gen4/GreatEncounters.groovy
Grayahm: It's being used like this elsewhere on Holon Scientist and Lillie so I think it's okay

src/tcgwars/logic/impl/gen4/GreatEncounters.groovy
Grayahm: Fixed

PR #554, opened 2020-06-04
RPS Implementations, minor fixes and Octolock rewrite

starg09:


src/tcgwars/logic/impl/gen8/SwordShield.groovy
```diff
@@ -2389,42 +2389,36 @@ public enum SwordShield implements LogicCardInfo {
...
+                    def list=[]
+                    for (move in h.object) {
+                      def copy=move.shallowCopy()
+                      copy.energyCost.addAll([C, C] as List<Type>)
+                      list.add(copy)
                     }
-                    bc "Attacks of $pcs will cost $energies more during next turn."
-                  bg.em().storeObject("Octolock", True)
+                    h.object=list
+                    bc "Attacks of $pcs will cost [C][C] while Octolock is active."
```
luongthomas, line 2409: Will cost C C more

starg09: Fixed 👍

starg09:
Added a fix for Build on GE.

PR #559, opened 2020-06-04
Opp's Active Energy Discard OoO fix, ADP extra prize order fix

starg09:
### ADP Extra Prize
If it works, need to apply to other cards like Guzzlord (CEC).
* Now should give the extra prize before the opponent chooses a new active.
* Removed an unnecessary afterDamage (there's no damage)
* Updated the bc to mention the extra damage enabled, and the eventual secondary effect.

### `discardDefendingSpecialEnergy()` (and a couple similar effects) Order of Operation overhaul
This shouldn't be done after the damage, otherwise cards like Horror Psychic Energy (RCL 172) will trigger when they shouldn't. 

> Q. If I use an attack that says to "discard an Energy attached to your opponent's Active Pokemon" and I choose to discard their "Dangerous Energy" card, would I still receive the 2 damage counters from "Dangerous Energy" or not?
>
> A. You would not receive the 2 damage counters. (Ancient Origins FAQ; Aug 13, 2015 TPCi Rules Team)

Affects multiple cards:
* Poliwrath (BS) [Both regular and Pokémod versions]
* Dragonair Delta (DS 42) [Also reworked it's twister flip]
* Regice-ex (EM 98)
* Typhlosion-ex (UF 110) 
* Gyarados (HS 4)
* Lycanroc-GX (BUS 136)
* Pelipper (CES 112) [Also made it be targeted so it can be blocked]
* Hydreigon (CIN 62)
* Dragalge (FLI 53) [Reworked Twister]
* Croagunk (FLI 63)
* Tyrunt (FLI 68)
* Weepinbell (GRI 2)
* Glalie (GRI 32)
* Garbodor (GRI 51)
* Drampa GX (GRI 115)
* Feraligatr (SLG 20)
* Masquerain (SUM 8) [Moves energy, instead of discarding]
* Alolan Muk (SUM 58)
* Sandile (SUM 83)
* Zoroark (SMP SM89)
* Tornadus-GX (SMP SM134) [Just made the all-Energy-discard be a targeted effect]
* Trevenant & Dusknoir-GX (SMP SM217) [Same as Tornadus-GX]
* Gyarados (TEU 30)
* Pancham (TEU 81)
* Dragonair (TEU 118) [Reworked Twister]
* Croagunk (UPR 56)
* Salandit (UNB 30)
* Lucario & Melmetal-GX (UNB 120) [Same as Tornadus-GX]
* Fearow (UNB 146) [Also simplified the repetition of actions]
* Archeops (UNM 121)
* Dragonair (UNM 149) [Reworked Twister]
* Flapple (RCL 22)
* Inteleon VMAX (RCL 50) [Like Tornadus-GX, but with energý to hand; also updated text]
* Inteleon (SSH 59) [Like Tornadus-GX]
* Mawile (SSH 129)

### DIFF
--- a/src/tcgwars/logic/groovy/TcgStatics.groovy
+++ b/src/tcgwars/logic/groovy/TcgStatics.groovy
@@ -172,11 +172,11 @@ class TcgStatics {
     new DiscardOpponentEnergyCard(Target.OPP_ACTIVE).run(bg())
   }
   static discardDefendingSpecialEnergy(Object delegate){
-    afterDamage { targeted (delegate.defending) {
+    targeted (delegate.defending) {
       if(delegate.defending.cards.filterByType(CardType.SPECIAL_ENERGY)){
         delegate.defending.cards.filterByType(CardType.SPECIAL_ENERGY).select("Discard").discard()
       }
-    } }
+    }
   }
   static discardAllSelfEnergy(Type type=null){
     def ef=new DiscardAllSelfEnergy(type)
--- a/src/tcgwars/logic/impl/gen1/BaseSetNG.groovy
+++ b/src/tcgwars/logic/impl/gen1/BaseSetNG.groovy
@@ -618,7 +618,7 @@ public enum BaseSetNG implements LogicCardInfo {
             energyCost W, W, C, C
             onAttack {
               damage 40
-              afterDamage{discardDefendingEnergy()}
+              discardDefendingEnergy()
             }
           }
 
--- a/src/tcgwars/logic/impl/gen3/DeltaSpecies.groovy
+++ b/src/tcgwars/logic/impl/gen3/DeltaSpecies.groovy
@@ -1571,23 +1571,12 @@ public enum DeltaSpecies implements LogicCardInfo {
           energyCost L, C
           attackRequirement {}
           onAttack {
-            flip 2, {}, {}, [
-              2:{ damage 20;
-                afterDamage{
-                  discardDefendingEnergy()
-                  discardDefendingEnergy()
-                }
-              },
-              1:{
-                damage 20
-                afterDamage{
-                  discardDefendingEnergy()
-                }
-              },
-              0:{
-                bc "$thisMove failed due to 2 TAILS."
-              }
-            ]
+            def noEff = true
+            flip 2, {
+              noEff = false
+              discardDefendingEnergy()
+            }
+            if (noEff) bc "$thisMove failed due to 2 TAILS." else damage 20
           }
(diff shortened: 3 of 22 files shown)


axpendix:
I believe this needs to be reverted ASAP. Energy cards must always be discarded AFTER any damage occurs, not before.

axpendix:
@starg09 @luongthomas

axpendix:
> 
> Q. If I use an attack that says to "discard an Energy attached to your opponent's Active Pokemon" and I choose to discard their "Dangerous Energy" card, would I still receive the 2 damage counters from "Dangerous Energy" or not?
> 
> A. You would not receive the 2 damage counters. (Ancient Origins FAQ; Aug 13, 2015 TPCi Rules Team)

Read the question properly, it has NO MENTION of a damage. The question only asks whether discard effect is considered as damage.

flagrama:
I would think the damage being done would be implied in the question, as the Dangerous Energy card does state it only puts the counters on if damage occurs. Claydol and Registeel are two Pokémon cards in that set that did damage and removed an energy card even.

But it definitely needs clarification before such a change is definitively made.

starg09:
> Read the question properly, it has NO MENTION of a damage. The question only asks whether discard effect is considered as damage. If the attack in question ALSO had a damage, Dangerous Energy WOULD put 2 damage counters.

From Dangerous Energy:

> Whenever the [D] Pokémon this card is attached to is your Active Pokémon and is damaged by an attack from your opponent’s Pokémon-EX (even if that Pokémon is Knocked Out), put 2 damage counters on the Attacking Pokémon-EX.

So, in the event of an attack that does damage and discards the energy (since otherwise the energy's effect wouldn't trigger:

1. The attack is declared from a Pokémon-EX to a Pokémon with a single Dangerous (or Horror [P]) Energy attached.
2. An energy (as I proposed in this PR) is removed _before_ the damage is applied. The one selected is Dangerous Energy.
3. Damage is calculated and post-damage effects are checked. Since the Dangerous Energy is no longer attached, the damage counters aren't placed. If another energy were removed, the counters would be put like normal.

On the old logic (checked by testing), Horror [P] and dangerous placed the counters regardless of it being discarded or not.

axpendix:
> On the old logic (checked by testing), Horror [P] and dangerous placed the counters regardless of it being discarded or not.

That is the correct behavior indeed

flagrama:
Further discussion happened on the Discord in the #dev channel: https://discordapp.com/channels/342318082288517120/404027209733505025/718212294294896731

PR #563, opened 2020-06-04
Optimized rockPaperScissors

ufodynasty:
Its not like it needs it but it was bugging me the first time


flagrama:
I don't think modulus will work for this. I tested it here, and it really only works with positive numbers so not all paths seems to work as intended. http://tpcg.io/rXb0sBVL

ufodynasty:
:(

flagrama:
Nice recovery :+1:

ufodynasty:
This one actually works in code instead of just in math :(

starg09:
Ohh, this looks way nicer! <3

PR #565, opened 2020-06-04
Implementations

Grayahm:


src/tcgwars/logic/impl/gen4/LegendsAwakened.groovy
```diff
@@ -332,25 +332,29 @@ public enum LegendsAwakened implements LogicCardInfo {
...
             }
           }
           move "Pester", {
             text "40+ damage. If the Defending Pokémon is affected by a Special Condition, this attack does 40 damage plus 40 more damage."
             energyCost F, C
-            attackRequirement {}
             onAttack {
-              damage 0
+              damage 40
+              if (!defending.specialConditions) {
```
flagrama, line 342: This is checking "if the defending Pokémon has no Special Conditions affecting it," it should probably be just `if (defending.specialConditions)` to check if it is affected by any conditions.

Grayahm: What a mistake lmao

Grayahm: Fixed

PR #566, opened 2020-06-05
Buzzap implementation

ufodynasty:
I really hate these energy implementations because I don't quite get everything that is involved but I think this will work none the less.


luongthomas:
👀

PR #568, opened 2020-06-05
Some DP/MT fixes, Implemented MT 1-67

starg09:
* Make Leech Seed on Cherubi (DP 75) last priority.
* Added a temp debug bc for Fossil Excavator (MT 111), to find why it crashes when selecting a stage 2 card on search.
* Fixed Retreat Cost for Pikachu (MT 94).
* Made Chesto Berry prevent the next sleep coinflip on Buizel (MT 75). Right now it checks even though the condition is already gone.
* Partially implemented Vigoroth (MT 68). It should also check for Slakoth having been asleep before evolving that turn.
* Magmortar LV.X (MT 123) should override the burn damage value to 30, not plainly increase it. Other burn cards should also do the added check for the value not being already increased as much or even higher than needed.
* Have Budew (DP 43) assert having deck before using Errand-Running
* Fixed Nidoran names in all DP-era sets, since they were written as "Nidoran Male/Female" instead of using the ♀/♂ symbols like every other set.

### DIFF
--- a/src/tcgwars/logic/groovy/TcgStatics.groovy
+++ b/src/tcgwars/logic/groovy/TcgStatics.groovy
@@ -1185,10 +1185,6 @@ class TcgStatics {
   }
 
   static boolean isValidFossilCard(Card potentialFossil){
-    if (potentialFossil.cardTypes.is(STAGE2)){
-      def debug_test = bg.gm().getBasicsFromStage2(potentialFossil.name)
-      bc "${debug_test}"
-    }
     if(
       (potentialFossil.cardTypes.is(ITEM) && potentialFossil.name.contains("Fossil")) ||
       (potentialFossil.cardTypes.is(STAGE1) && potentialFossil.predecessor.contains("Fossil")) ||
--- a/src/tcgwars/logic/groovy/TcgStatics.groovy
+++ b/src/tcgwars/logic/groovy/TcgStatics.groovy
@@ -1188,7 +1188,7 @@ class TcgStatics {
     if(
       (potentialFossil.cardTypes.is(ITEM) && potentialFossil.name.contains("Fossil")) ||
       (potentialFossil.cardTypes.is(STAGE1) && potentialFossil.predecessor.contains("Fossil")) ||
-      (potentialFossil.cardTypes.is(STAGE2) && bg.gm().getBasicsFromStage2(potentialFossil.name).any{it.contains("Fossil")})
+      (potentialFossil.cardTypes.is(STAGE2) && bg().gm().getBasicsFromStage2(potentialFossil.name).any{it.contains("Fossil")})
     )
       return true
     else
--- a/src/tcgwars/logic/groovy/TcgStatics.groovy
+++ b/src/tcgwars/logic/groovy/TcgStatics.groovy
@@ -1185,10 +1185,14 @@ class TcgStatics {
   }
 
   static boolean isValidFossilCard(Card potentialFossil){
+    if (potentialFossil.cardTypes.is(STAGE2)){
+      def debug_test = bg.gm().getBasicsFromStage2(potentialFossil.name)
+      bc "${debug_test}"
+    }
     if(
       (potentialFossil.cardTypes.is(ITEM) && potentialFossil.name.contains("Fossil")) ||
       (potentialFossil.cardTypes.is(STAGE1) && potentialFossil.predecessor.contains("Fossil")) ||
-      (potentialFossil.cardTypes.is(STAGE2) && bg.gm().getBasicsFromStage2(potentialFossil.name).findAll{it.contains("Fossil")})
+      (potentialFossil.cardTypes.is(STAGE2) && bg.gm().getBasicsFromStage2(potentialFossil.name).any{it.contains("Fossil")})
     )
       return true
     else
--- a/src/tcgwars/logic/impl/gen4/MysteriousTreasures.groovy
+++ b/src/tcgwars/logic/impl/gen4/MysteriousTreasures.groovy
@@ -268,7 +268,7 @@ public enum MysteriousTreasures implements LogicCardInfo {
                   before APPLY_ATTACK_DAMAGES, {
                     bg.dm().each {
                       if (it.to==self && it.dmg.value && it.notNoEffect && it.from.topPokemonCard.cardTypes.is(STAGE2)) {
-                        bc "-30 to Alakazam (Psychic Guard)"
+                        bc "-30 to $self ($thisMove)"
                         it.dmg-=hp(30)
                       }
                     }
--- a/src/tcgwars/logic/impl/gen4/MysteriousTreasures.groovy
+++ b/src/tcgwars/logic/impl/gen4/MysteriousTreasures.groovy
@@ -909,6 +909,17 @@ public enum MysteriousTreasures implements LogicCardInfo {
           pokeBody "Poison Sacs", {
             text "Your opponent can’t remove the Special Condition Poisoned by evolving or devolving his or her Poisoned Pokémon. (This also includes putting a Pokémon Level-Up card onto the Poisoned Pokémon.)"
             delayedA {
+              before POISONED_SPC, null, null, EVOLVE, {
+                if(ef.target == self.owner.opposite){
+                  prevent()
+                }
+              }
+              // Potential addition once level-up is implemented
+              // before POISONED_SPC, null, null, LEVEL_UP, {
+              //   if(ef.target == self.owner.opposite){
+              //     prevent()
+              //   }
+              // }
             }
           }
           move "Knuckle Claws", {
@@ -916,7 +927,8 @@ public enum MysteriousTreasures implements LogicCardInfo {
             energyCost P, C
             attackRequirement {}
             onAttack {
-              damage 0
+              damage 30
+              applyAfterDamage POISONED
             }
           }
 
--- a/src/tcgwars/logic/impl/gen4/MysteriousTreasures.groovy
+++ b/src/tcgwars/logic/impl/gen4/MysteriousTreasures.groovy
@@ -1581,14 +1581,15 @@ public enum MysteriousTreasures implements LogicCardInfo {
       case VIGOROTH_68:
         return evolution (this, from:"Slakoth", hp:HP080, type:COLORLESS, retreatCost:1) {
           weakness F, PLUS20
+          /*def asleepBeforeEvolve = false*/
           move "Wake-up Punch", {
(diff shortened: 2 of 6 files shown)


src/tcgwars/logic/impl/gen4/MysteriousTreasures.groovy
```diff
@@ -240,14 +253,31 @@ public enum MysteriousTreasures implements LogicCardInfo {
           pokePower "Power Cancel", {
             text "Once during your opponent’s turn, when your opponent’s Pokémon uses any Poké-Power, you may discard 2 cards from your hand and prevent all effects of that Poké-Power. (This counts as that Pokémon using its Poké-Power.) This power can’t be used if Alakazam is affected by a Special Condition."
             actionA {
+              //TODO: Yeah this is gonna be a fun one to do.
+              //Requires a way to capture "powerUsed", stop the Poké-Power at that point but make it count as used.
```
flagrama, line 257: Based off the usual "Is ability blocked" code here:
```
eff = getter IS_ABILITY_BLOCKED, { Holder h->
            if (self.active && (h.effect.ability instanceof PokeBody || h.effect.ability instanceof PokePower)) {
              h.object=true
            }
          }
```
I wonder if you can just set `h.effect.ability.used = true` where you set the block to true as well.

src/tcgwars/logic/impl/gen4/MysteriousTreasures.groovy
```diff
@@ -240,14 +253,31 @@ public enum MysteriousTreasures implements LogicCardInfo {
...
             attackRequirement {}
             onAttack {
-              damage 0
+              damage 50
+              afterDamage {
+                delayed {
+                  before APPLY_ATTACK_DAMAGES, {
+                    bg.dm().each {
+                      if (it.to==self && it.dmg.value && it.notNoEffect && it.from.topPokemonCard.cardTypes.is(STAGE2)) {
+                        bc "-30 to Alakazam (Psychic Guard)"
```
flagrama, line 271: Should probably use `$self` here, or not mention the Pokémon at all? I assume if the attack is copied then it would display this exact text still.

src/tcgwars/logic/impl/gen4/MysteriousTreasures.groovy
```diff
@@ -258,17 +288,29 @@ public enum MysteriousTreasures implements LogicCardInfo {
           move "Tail Influence", {
             text "30 damage. Your opponent flips a coin until he or she gets heads. For each tails, remove an Energy card attached to the Defending Pokémon and put it on the bottom of your opponent’s deck."
             energyCost C, C
-            attackRequirement {}
+            attackRequirement {
+              assert defending.cards.any{it.cards.filterByType(ENERGY)} : "The defending Pokémon has no Energy cards attached to them"
```
flagrama, line 292: This attack does damage already, is it correct to disallow it just because the opponent has not energy attached?

src/tcgwars/logic/impl/gen4/MysteriousTreasures.groovy
```diff
@@ -433,15 +588,26 @@ public enum MysteriousTreasures implements LogicCardInfo {
...
             }
           }
           move "Flame Drum", {
             text "80 damage. If Magby isn’t anywhere under Magmortar, discard 2 Energy cards from your hand. (If you can’t discard 2 Energy cards from your hand, this attack does nothing.)"
             energyCost R, R, C
-            attackRequirement {}
+            attackRequirement {
+              def condition1 = (self.getPokemonCards().any{it.name.contains("Magby")})
+              def condition2 = (my.hand.filterByType(ENERGY).size() >= 2)
+              assert (condition1 || condition2) : "Magby isn't anywhere under Magmortar, and you can't discard 2 Energy cards from your hand"
```
flagrama, line 601: The text doesn't quite match the check which may cause confusion.

src/tcgwars/logic/impl/gen4/MysteriousTreasures.groovy
```diff
@@ -433,15 +588,26 @@ public enum MysteriousTreasures implements LogicCardInfo {
...
+            }
             onAttack {
-              damage 0
+              if (!self.getPokemonCards().any{it.name.contains("Magby")}) {
```
flagrama, line 604: If it is already an attackRequirement, I don't think there is any reason to duplicate the condition in an if statement.

src/tcgwars/logic/impl/gen4/MysteriousTreasures.groovy
```diff
@@ -214,22 +214,35 @@ public enum MysteriousTreasures implements LogicCardInfo {
...
           move "Heap Up", {
-            text "40+ damage. Special Energy cards there this attack does 40 damage plus 30 more damage. Put all of those Energy cards on top of your deck. Shuffle your deck afterward."
-            energyCost C, C, M
+            text "40+ damage. Search your discard pile for all Energy cards and show them to your opponent. If you find any [M] Special Energy cards there this attack does 40 damage plus 30 more damage. Put all of those Energy cards on top of your deck. Shuffle your deck afterward."
+            energyCost C, C
             attackRequirement {}
             onAttack {
-              damage 0
+              damage 40
+              def tar = my.discard.filterByType(POKEMON_TOOL, SPECIAL_ENERGY)
```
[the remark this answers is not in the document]
starg09: Whoops! Left behind from the original card I checked, thanks! :P

src/tcgwars/logic/impl/gen4/MysteriousTreasures.groovy
starg09: I think it might require it's own handler (wanna talk about it with axpendix later if possible. Main issue is making sure the opponent _can_ use the power, and then as soon as it's marked as powerUsed() trigger the popup for activating this blocker PokéPower. 

It *could* just pop up once the power button is clicked, like I understand that one works. But I'd like 100% confirmation before proceeding with it. If it blocks as soon as the button is pressed, the opponent could just call a Power they know they can't use, get that one blocked and use the rest normally (since it's a single block per ability/card).

src/tcgwars/logic/impl/gen4/MysteriousTreasures.groovy
starg09: Yeah, I recall adding the damage at last since I missed it. Coded it thinking it only discarded thanks!

src/tcgwars/logic/impl/gen4/MysteriousTreasures.groovy
starg09: It does? Seems good to me 🤔 
* Card Text: _"**If Magby isn’t anywhere under Magmortar**, discard 2 Energy cards from your hand. (**If you can’t discard 2 Energy cards from your hand, this attack does nothing.**)"_

As long as any of those two conditions are met, it should be able to attack. I did alter the attack though, since it wasn't doing damage when Magby was under Magmortar. But don't think that's what you meant, could you let me know where in the assert do you think it's not quite matching?

src/tcgwars/logic/impl/gen4/MysteriousTreasures.groovy
starg09: Ah, ok hadn't seen this comment. Yeah, the check is fine since there are two ways this can be solved. With the change of where damage is declared it should be working properly now, and the magby check here is for doing the discard or not :P

src/tcgwars/logic/impl/gen4/MysteriousTreasures.groovy
flagrama: I mean the assertion text. Code uses an OR, text is an AND. Along with that "Magby isn't anywhere under Magmortar" is the opposite of what condition 1 is actually checking. If somebody evolved from Magby and tried using it, the text would tell them "Magby isn't anywhere under Magmortar" even though it is. It would instead be blaming them having not enough energy to discard even though they most likely do.

starg09: I don't see the code doing that. The two conditions are for different scenarios, and second doesn't need to be checked at all if Magmortar has a Magby below (which is why I used an OR for the assert):
* Magby is below Magmortar, assert is OK.
* Magby is not below Magmortar, but there are energies, assert is OK
* Both things don't happen, throw the assert fail.

flagrama: I'll just blame it being 3AM at the time, let's be honest. I see what's going on now.

src/tcgwars/logic/groovy/TcgStatics.groovy
```diff
@@ -1204,7 +1204,7 @@ class TcgStatics {
     if(
       (potentialFossil.cardTypes.is(ITEM) && potentialFossil.name.contains("Fossil")) ||
       (potentialFossil.cardTypes.is(STAGE1) && potentialFossil.predecessor.contains("Fossil")) ||
-      (potentialFossil.cardTypes.is(STAGE2) && bg.gm().getBasicsFromStage2(potentialFossil.name).any{it.contains("Fossil")})
+      (potentialFossil.cardTypes.is(STAGE2) && bg().gm().getBasicsFromStage2(potentialFossil.name).any{it.contains("Fossil")})
```
flagrama, line 1207: My understanding at this point is that `bg` doesn't require parens because it is a top-level Groovy method in the `TcgStatics` however `gm()` (and `em()`) does as it is a method on a Java `Battleground` class that `bg` returns. So this change shouldn't be necessary, but doesn't hurt either.

starg09: Yeah, doing it just to try (and to avoid doing another commit/PR in case it _was_ needed, heh). As you said, doesn't hurt :P

One "clue" I found about this is that the filter works fine on discard (where I use filterAll). The crash happens exclusively on the search menu, and only on STAGE2 cards. No matter if fossil or not, that's the area where it breaks and gives [this crash](https://gist.githubusercontent.com/starg09/0611c4b5e80e68bc3a0f5b1509cef6b1/raw/7aefb03d7d332e63d24cd29c709b7a3bcc067aad/Excavator_Search_Stage2_Crash)...

flagrama: Interestingly, I think with how Groovy works this line could actually be
```
(potentialFossil.cardTypes.is STAGE2 && bg.gm().getBasicsFromStage2 potentialFossil.name findAll {it.contains "Fossil"})
``` 

from some of the Groovy stuff I've been looking up. Not that I think you should stray _too far_ from what seems to be standard currently.

PR #569, opened 2020-06-05
Cosmic eclipse fixes

flagrama:
Leaving this as a draft for comments and for me to finish later.

### DIFF
--- a/src/tcgwars/logic/impl/gen7/CosmicEclipse.groovy
+++ b/src/tcgwars/logic/impl/gen7/CosmicEclipse.groovy
@@ -2627,7 +2627,7 @@ public enum CosmicEclipse implements LogicCardInfo {
             text "Switch 1 of your opponent’s Benched Pokémon with their Active Pokémon."
             energyCost C, C
             attackRequirement {
-              assert opp.bench
+              assert opp.bench : "Your opponent has no benched Pokémon."
             }
             onAttack {
               sw(opp.active, opp.bench.select())
--- a/src/tcgwars/logic/impl/gen7/CosmicEclipse.groovy
+++ b/src/tcgwars/logic/impl/gen7/CosmicEclipse.groovy
@@ -2875,7 +2875,7 @@ public enum CosmicEclipse implements LogicCardInfo {
               checkLastTurn()
               assert my.deck : "Your deck is empty."
               powerUsed()
-              def tar = my.deck.select(min: 0, max: 1, "Choose 1 card to put on top of your deck.").first()
+              def tar = my.deck.select("Choose 1 card to put on top of your deck.").first()
               my.deck.remove(tar)
               shuffleDeck()
               my.deck.add(0, tar)
--- a/src/tcgwars/logic/impl/gen7/CosmicEclipse.groovy
+++ b/src/tcgwars/logic/impl/gen7/CosmicEclipse.groovy
@@ -2230,7 +2230,7 @@ public enum CosmicEclipse implements LogicCardInfo {
             energyCost P
             onAttack {
               def list = LUtils.selectMultiPokemon(bg.oppClient(), opp.bench, "Opponent used Charming Stamp. Select a Pokémon to have 90 damage dealt to it.", 1)
-              opp.bench.findAll { !list.contains(it) }.each {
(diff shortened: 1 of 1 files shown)


src/tcgwars/logic/impl/gen7/CosmicEclipse.groovy
```diff
@@ -1498,9 +1505,12 @@ public enum CosmicEclipse implements LogicCardInfo {
           move "Direct Dive", {
             text "Discard all Energy from this Pokémon. This attack does 100 damage to 1 of your opponent's Benched Pokémon. (Don't apply Weakness and Resistance for Benched Pokémon.)"
             energyCost W, W, W
+            attackRequirement {
+              assert opp.bench : "Your opponent has no Benched Pokémon."
+            }
```
starg09, line 1510: Might require an actual official ruling (since the only one similar to this is from 12 years ago). But I'd say you can use the attack, discard the Pokémon Energy and not attack due to lack of bench. The discard to me seems like an effect, not a cost. Other cards should be updated based on how this is handled:

* Talonflame (UNM)
* Kingdra (BUS)
* Arceus (DP Promo DP50)
* Hippowdon LV.X (Rising Rivals 107)
* Blastoise (Platinum 2)
* Ampharos (POP7 1, Secret Wonders 1)

From Compendium LV.X:

> Q. Can you use Salamence-EX's "Hydro Wave" attack if the opponent has no Benched Pokemon?
>
> A. Yes. Salamence-EX would simply discard all attached Water Energy and that's all. (Apr 3, 2008 PUI Rules Team)
>
>

flagrama: You're probably right. It still has an effect even if it does no damage. I'll correct this.

src/tcgwars/logic/impl/gen7/CosmicEclipse.groovy
```diff
@@ -2225,7 +2230,7 @@ public enum CosmicEclipse implements LogicCardInfo {
             energyCost P
             onAttack {
               def list = LUtils.selectMultiPokemon(bg.oppClient(), opp.bench, "Opponent used Charming Stamp. Select a Pokémon to have 90 damage dealt to it.", 1)
-              opp.bench.findAll { !list.contains(it) }.each {
+              opp.bench.findAll { list.contains(it) }.each {
                 damage 90 it
               }
             }
```
starg09, line 2236: Couldn't this use

`damage 90, opp.all.oppChoose("Opponent used Charming Stamp. Select a Pokémon to have 90 damage dealt to it.")` 

instead? Also, as it is right now it wouldn't be able to pick the active.

flagrama: You're right. I spent way too much time trying to figure out how this even was trying to do what it was trying to do, and failing to remember what `oppSelect` was called, I missed that.

src/tcgwars/logic/impl/gen7/CosmicEclipse.groovy
```diff
@@ -3707,8 +3707,18 @@ public enum CosmicEclipse implements LogicCardInfo {
             energyCost Y, C, C
             onAttack {
               damage 80
-              // TODO: edit it to include only tag team supporters
-              if(bg.em().retrieveObject("last_supporter_play_turn") == bg.turnCount) damage 80
+              if(bg.em().retrieveObject("last_supporter_play_turn") == bg.turnCount && bg.em().retrieveObject("last_tag_team_supporter_play_turn") == bg.turnCount) {
```
starg09, line 3710: Is the supporter turn check needed? If a tag team supporter was played, a supporter would've been played right?

flagrama: Yes. That was left over from when I was planning on just storing a true value in an effect and then setting it back to null when unregistering it.

flagrama:
Rebased onto current master

flagrama:
I can also break this out into smaller PRs if needed. Most likely by Pokémon type or something else. I hadn't expected this PR to be this large due to the previous sets not being so.

PR #571, opened 2020-06-05
Trev/Dusk GX SM217 - Attempt to fix Pale Moon GX

flagrama:
Pale Moon GX currently affects Pokémon that are ignoring effects from
attacks.


flagrama:
Did not accomplish what I wanted. Now the move does nothing.

PR #577, opened 2020-06-06
Add SWSH Promos to Collection File

luongthomas:
After adding this entry to the Collections File, we'll be able to test the cards out on the dev server


luongthomas:
oh fk lemme remake

PR #579, opened 2020-06-06
[Ready for Review] Various Fixes

starg09:
* Tri Attack-GX on Alolan Muk-GX (BUS 84) can be used regardless of the opponent having bench.
  - This is an important fix (and might need to check around other GX attacks where this may apply), due to it allowing attacks (in cards like this one, a free attack) that enable boosts on cards like Hala.
  - Relevant ruling:
    > Q. If my opponent has no Benched Pokemon, can I still use Alolan Muk-GX's "Tri Hazard-GX" attack?
    >
    > A. You can use it, but it won't have any effect, and it will count as having used your GX attack for the game. (Burning Shadows FAQ; Aug 3, 2017 TPCi Rules Team)

* Fixed "Extra Prize" check on multiple cards being done after KOs instead of before them. Should fix opponent choosing the new active before the extra prizes are taken. Cards affected:
  - Guzzlord (CEC 136)
  - Guzzlord-GX (CIN 63)
  - Whimsicott (GRI 91)
  - Pheromosa & Buzzwole-GX (UNB 1)
  - Mega Sableye & Tyranitar-GX (UNM 126)
* Simplified code for Slowking (DS 28)
* Removed the following unintended text from several GX attacks: _"\nPokémon-GX rule: When your Pokémon-GX is Knocked Out, your opponent takes 2 Prize cards."_
* Fixed some characters with broken encoding in Darkrai Prism Star (UPR 77)
* Made Fearow (UNB 146) use `2.times{[...]}` instead of repeating code.
* Wrapped Prankish from Clefable (RCL 75) around a `targeted`, marking it as an Ability effect to other cards.
* Fixed an assert inside the attack of Lumineon (UNM 40), also made the energy move optional as the card reads.
* Added an ability/trainer `target` to Lycanroc-GX (TEU 82) and Nita (TEU 151)
* Simplified the "Chain Lightning" attack from Electrode (Jungle 2)
* Fixed some attacks and abilities reducing damage from the opponent's Pokémon after applying Weakness & Resistance, instead of before as the text indicates. Affected cards:
  - Slaking (DX 15), Poké-Body "Lazy Aura"
  - Masquerain (DX 39), Poké-Body "Intimidating Pattern"
  - Meganium Delta (DF 4), Attack "Delta Reduction"
  - Togepi Delta (DF 41), Attack "Charm"
  - Kingdra-_ex_ Delta (DF 94), Poké-Body "Extra Smoke"
  - Deoxys Delta (HP 4), Attack "Delta Reduction" _[Changed to a specific delayed effect here, since it's only damage to Deoxys Delta]_
  - Chinchou (PK 49), Attack "Negative Ion"
  - Feraligatr (UF 4), Poké-Body "Intimidating Fang"
  - Granbull (UF 39), Poké-Body "Intimidating Fang" 
* Added a TODO for Charizard Delta (CG 4).
* Made some Pokémon check if they're active before using their Poké-Body/Ability:
  - Dark Sandslash (TRR 18)
  - Qwilfish (TRR 27)
  - Rocket's Hitmonchan-_ex_ (TRR 98)
  - Dragalge (FLI 53)
  - Magmortar (UPR 19)
* Made Kartana-GX (CIN 70) not assert bench being empty. You can shuffle it and lose the game due to not having any Pokémon in play (TODO: Check for more cards doing this unneeded check).
* Minor text fix on Reshiram-GX (DRM 11).
* Fixed "Iron Wings" on Corviknight (SSH 135) not being ignored by shred attacks.
* ~~Galarian Stunfisk (SSH 132) should now ask its owner for which energy to discard, not the opponent.~~ **(Reverted, better implemented in #584)**
  - Side-note: Against cards like Blacephalon-GX or Eldegoss V, this card shouldn't be able to discard any card. But it's last priority already... Would using `after APPLY_ATTACK_DAMAGES` instead of `before [...]` work?

### DIFF
--- a/src/tcgwars/logic/impl/gen3/DeltaSpecies.groovy
+++ b/src/tcgwars/logic/impl/gen3/DeltaSpecies.groovy
@@ -1170,17 +1170,8 @@ public enum DeltaSpecies implements LogicCardInfo {
           attackRequirement {}
           onAttack {
             damage 20
-            if (self.cards.filterByType(ENERGY).findAll {
-              it.asEnergyCard().containsTypePlain(W)
-            }) {
-              damage 20
-            }
-
-            if (self.cards.filterByType(ENERGY).findAll {
-              it.asEnergyCard().containsTypePlain(P)
-            }) {
-              discardDefendingSpecialEnergy(delegate)
-            }
+            if (self.cards.filterByEnergyType(W)) { damage 20 }
+            if (self.cards.filterByEnergyType(P)) { discardDefendingSpecialEnergy(delegate) }
           }
         }
       };
--- a/src/tcgwars/logic/impl/gen3/HolonPhantoms.groovy
+++ b/src/tcgwars/logic/impl/gen3/HolonPhantoms.groovy
@@ -340,7 +340,7 @@ public enum HolonPhantoms implements LogicCardInfo {
             damage 30
             afterDamage{
               delayed{
-                before PROCESS_ATTACK_EFFECTS, {
+                after PROCESS_ATTACK_EFFECTS, {
                   bg.dm().each{
                     if(it.from.owner == self.owner.opposite && it.to == self) {
                       bc "Delta Reduction -30 (before W/R)"
--- a/src/tcgwars/logic/impl/gen3/HolonPhantoms.groovy
+++ b/src/tcgwars/logic/impl/gen3/HolonPhantoms.groovy
@@ -338,7 +338,20 @@ public enum HolonPhantoms implements LogicCardInfo {
           attackRequirement {}
           onAttack {
             damage 30
-            reduceDamageNextTurn(hp(30), thisMove)
+            afterDamage{
+              delayed{
+                before PROCESS_ATTACK_EFFECTS, {
+                  bg.dm().each{
+                    if(it.from.owner == self.owner.opposite && it.to == self) {
+                      bc "Delta Reduction -30 (before W/R)"
+                      it.dmg -= hp(30)
+                    }
+                  }
+                }
+                after SWITCH, self, {unregister()}
+                unregisterAfter 2
+              }
+            }
           }
         }
       };
--- a/src/tcgwars/logic/impl/gen7/UnbrokenBonds.groovy
+++ b/src/tcgwars/logic/impl/gen7/UnbrokenBonds.groovy
@@ -344,7 +344,7 @@ public enum UnbrokenBonds implements LogicCardInfo {
               gxPerform()
               delayed (priority: LAST) {
                 def pcs = defending
-                after KNOCKOUT, pcs, {
+                before KNOCKOUT, pcs, {
                   bg.em().run(new TakePrize(self.owner, pcs))
                   if(self.cards.energySufficient(thisMove.energyCost + C + C+ C+ C+ C+ C+ C)){
                     bg.em().run(new TakePrize(self.owner, pcs))
--- a/src/tcgwars/logic/impl/gen7/UnbrokenBonds.groovy
+++ b/src/tcgwars/logic/impl/gen7/UnbrokenBonds.groovy
@@ -3580,7 +3580,7 @@ public enum UnbrokenBonds implements LogicCardInfo {
             energyCost C, C
             onAttack {
               damage 70
-              flip {afterDamage{discardDefendingEnergy();discardDefendingEnergy()}}
+              flip { afterDamage{ 2.times{ discardDefendingEnergy() } } }
             }
           }
 
--- a/src/tcgwars/logic/impl/gen8/RebelClash.groovy
+++ b/src/tcgwars/logic/impl/gen8/RebelClash.groovy
@@ -2031,10 +2031,10 @@ public enum RebelClash implements LogicCardInfo {
         resistance F, MINUS30
         bwAbility "Counterattack", {
           text "If this Pokemon is your Active Pokemon and is damaged by an opponent’s attack, place 3 damage counters on the attacking Pokemon."
-          delayedA {
+          delayedA (priority: LAST) {
             before APPLY_ATTACK_DAMAGES, {
               if (bg.currentTurn == self.owner.opposite && bg.dm().find({ it.to==self && it.dmg.value }) && self.active) {
-                powerused()
+                bc "Counterattack activates"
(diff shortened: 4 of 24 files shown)


starg09:
Ok, this got a bit big so I'll stop pushing for now 👍

src/tcgwars/logic/impl/gen3/DeltaSpecies.groovy
```diff
@@ -1170,17 +1170,8 @@ public enum DeltaSpecies implements LogicCardInfo {
...
-            }) {
-              damage 20
-            }
-
-            if (self.cards.filterByType(ENERGY).findAll {
-              it.asEnergyCard().containsTypePlain(P)
-            }) {
-              discardDefendingSpecialEnergy(delegate)
-            }
+            if (self.cards.filterByEnergyType(W)) { damage 20 }
```
flagrama, line 1173: How do both the old version and this new version handle Energy that provide multiple types, but only one type at a time?

src/tcgwars/logic/impl/gen3/HolonPhantoms.groovy
```diff
@@ -338,7 +338,20 @@ public enum HolonPhantoms implements LogicCardInfo {
           attackRequirement {}
           onAttack {
             damage 30
-            reduceDamageNextTurn(hp(30), thisMove)
+            afterDamage{
```
flagrama, line 341: What does this do differently than `reduceDamageNextTurn`?

src/tcgwars/logic/impl/gen7/UnbrokenBonds.groovy
```diff
@@ -3580,7 +3580,7 @@ public enum UnbrokenBonds implements LogicCardInfo {
             energyCost C, C
             onAttack {
               damage 70
-              flip {afterDamage{discardDefendingEnergy();discardDefendingEnergy()}}
+              flip { afterDamage{ 2.times{ discardDefendingEnergy() } } }
```
flagrama, line 3583: Interesting Groovy feature. Maybe this could be used for the instances of taking two prize cards you've somewhat touched already too?

src/tcgwars/logic/impl/gen3/DeltaSpecies.groovy
starg09: Should take a single rainbow as sufficient for both effects.

> Q. When using Houndoom's "Black Fire" attack, what happens if I discard an Energy card that provides multiple types of Energy at the same time?
>
> A. If the Energy card you discarded provides all types (for example, Rainbow Energy), then both effects would trigger - the "Black Fire" attack would to 70 damage AND apply the Burned Special Condition to the Defending Pokemon. (DP:Great Encounters FAQ; Feb 21, 2008 PUI Rules Team)

> Q. If you have a Rainbow Energy attached to a Pokemon using one of the Miracle Sphere trainers, can you do both conditions of the card (i.e. both Confused *AND* removing damage for Alpha)? The Miracle Sphere cards specify "if you have xxx Basic Energy cards..." So Rainbow will not work for the extra requirements of the Miracle Spheres, right?
> 
> A. Right. (May 8, 2003 WotC Chat, Q1523 & Q1542)

Couldn't find a specific ruling from EX Era, but those two follow line with each check happening independently, so a single rainbow and all other involved energies being colorless should still trigger all effects.

src/tcgwars/logic/impl/gen3/HolonPhantoms.groovy
starg09: reduceDamage does the reduction after weakness and resistance is applied. Since there's not a static that handles it prior to that, it had to be hardcoded in the card.

src/tcgwars/logic/impl/gen7/UnbrokenBonds.groovy
starg09: Could do indeed! Wanna confirm it working well but if it does will def update those 👍

src/tcgwars/logic/impl/gen3/DeltaSpecies.groovy
flagrama: So "1 at a time" only refers to attack energy cost it seems?

starg09: Yup, it's geared towards the base energy cost printed in the cards. No Rainbow+DCE shenanigans :P

src/tcgwars/logic/impl/gen8/RebelClash.groovy
```diff
@@ -2031,10 +2031,10 @@ public enum RebelClash implements LogicCardInfo {
         resistance F, MINUS30
         bwAbility "Counterattack", {
           text "If this Pokemon is your Active Pokemon and is damaged by an opponent’s attack, place 3 damage counters on the attacking Pokemon."
-          delayedA {
+          delayedA (priority: LAST) {
             before APPLY_ATTACK_DAMAGES, {
               if (bg.currentTurn == self.owner.opposite && bg.dm().find({ it.to==self && it.dmg.value }) && self.active) {
-                powerused()
+                bc "Counterattack activates"
```
flagrama, line 2037: Why this change? Without a check for last turn usage, it should produce this same output.

src/tcgwars/logic/impl/gen8/RebelClash.groovy
```diff
@@ -2029,10 +2031,10 @@ public enum RebelClash implements LogicCardInfo {
...
```
flagrama, line 2037: Why this change? As long as the proper method `powerUsed()` is used, it should provide the same output. The issue is that `powerused()` doesn't exist.

starg09: powerUsed doesn't seem to be used on any passive abilities? Removed it so it works exactly like other "if damaged, put 3 counters on..." abilities, like on Dragon Majesty

flagrama: I suppose. I had added this just because there was no text displayed at all. I figured it'd be easier to use `powerUsed` in any ability that you just need to know it activated and that's what caused additional things in the game log to happen. Either way it works I guess, but I'd at least include the Pokémon's name.

PR #580, opened 2020-06-06
assertAnyPokemonInPlay() static method, isPokemonV() / isPokemonVMAX() PCS methods

starg09:
Placed it in some cards from Cosmic Eclipse as examples.

* ~~Could probably use a "type filter" param, since a couple of cards make use of it (Like N's Resolve in the examples).~~ Already implemented into the draft.
* ~~Another potential param filter could be for "Pokémon ex/LV.X/EX/GX/V/VMAX", since a couple cards search for those specifically (See: Great Catcher CEC 192)~~ Implemented as well, just for EXs/GXs/Vs tho. More can be added later on, if needed, without much trouble.
  - Implemented isPokemonV() and isPokemonVMAX() on PCS for this to work

### DIFF
--- a/src/tcgwars/logic/groovy/TcgStatics.groovy
+++ b/src/tcgwars/logic/groovy/TcgStatics.groovy
@@ -1163,7 +1163,8 @@ class TcgStatics {
    * @param message Additional text to be put at the end of the assert fail warning. If repText is set to true in params, message will override the entirety of said warning.
    *
    */
-  static void assertBench(params=[:], Closure c = null, String message = "") {
+  static void assertBench(params=[:], Closure c = null, String info = "") {
+    def failMessage
     def checkedBench = params.opp ? opp.bench : my.bench
 
     def hasPokeCnt = 0
@@ -1185,34 +1186,37 @@ class TcgStatics {
       }
     )
 
-    def typeString = param.isType ? " ${params.isType.getShortNotation()}" : ""
+    if (!params.repText) {
+      failMessage = info
+    } else {
+      def typeString = param.isType ? " ${params.isType.getShortNotation()}" : ""
 
-    def pokeString = "${typeString}Pokémon"
-    if (hasPokeCnt) (
-      pokeString = "${typeString}"
-      i = 1
-      [
-        (params.hasPokemonEX, "Pokémon-EX"),
-        (params.hasPokemonGX, "Pokémon-GX"),
-        (params.hasPokemonV, "Pokémon V")
-      ].each{
-        if (it[0]) {
-          pokeString += it[1] + (if (i == isPokeCnt) "" else if (i == isPokeCnt-1) " or " else ", ")
-          i += 1
+      def pokeString = "${typeString}Pokémon"
+      if (hasPokeCnt) (
+        pokeString = "${typeString}"
+        i = 1
+        [
+          (params.hasPokemonEX, "Pokémon-EX"),
+          (params.hasPokemonGX, "Pokémon-GX"),
+          (params.hasPokemonV, "Pokémon V")
+        ].each{
+          if (it[0]) {
+            pokeString += it[1] + (if (i == isPokeCnt) "" else if (i == isPokeCnt-1) " or " else ", ")
+            i += 1
+          }
         }
-      }
-    )
+      )
 
+      failMessage = "${params.opp ? "Your opponent doesn't" : "You don't"} have any Benched $pokeString that ${!info.equals("") ? info : "follow the stated condition(s)"}"
+    }
 
-    if (!params.repText) message = "${params.opp ? "Your opponent doesn't" : "You don't"} have any Benched $pokeString that ${!message.equals("") ? message : "follow the stated condition(s)"}"
-
-    assert checkedBench.any{benchFilter} : message
+    assert checkedBench.any{benchFilter} : failMessage
   }
-  static void assertMyBench(params=[:], Closure c = null, String message = "") {
+  static void assertMyBench(params=[:], Closure c = null, String info = "") {
     params.opp = false
     assertBench(params, c, cText)
   }
-  static void assertOppBench(params=[:], Closure c = null, String message = "") {
+  static void assertOppBench(params=[:], Closure c = null, String info = "") {
     params.opp = true
     assertBench(params, c, cText)
   }
--- a/src/tcgwars/logic/groovy/TcgStatics.groovy
+++ b/src/tcgwars/logic/groovy/TcgStatics.groovy
@@ -1146,6 +1146,42 @@ class TcgStatics {
   static boolean isGxPerformed(){
     bg.em().retrieveObject("gx_"+my.owner)
   }
+
+  /* General checks for attacks and abilities */
+
+  //  * checkAnyBench
+  //    - Can be used with no arguments, or with a condition and additional text included.
+  //    - If a closure is given, it'll check for any "it" element in bench returning true.
+  //    - Optional params:
+  //      + opp: If true, checks for the opponent's bench instead of "my" bench.
+  //      + repText: If true, instead of adding cText at the end of the assert it'll be the only thing printed.
+  static void checkAnyBench(params=[:], Closure c, String cText) {
+    def checkedBench = params.opp ? opp.bench : my.bench
+    assert (
+      if (c != null) { checkedBench.any{c(it)} } else { checkedBench }
+    ) : (
+      if (params.repText) { cText } else {
+        (
+          if (params.opp) { "You don't" } else { "Your opponent doesn't" }
+        ) + " have any Benched Pokémon that " + (
+          if (cText) {
+            cText
+          } else {
+            "follow the stated condition(s)."
+          }
+        )
+      }
+    )
+  }
+  static void checkMyBench(params=[:], Closure c, String cText) {
+    params.opp = false
+    checkAnyBench(params, c, cText)
+  }
+  static void checkOppBench(params=[:], Closure c, String cText) {
+    params.opp = true
+    checkAnyBench(params, c, cText)
+  }
+
   static void cantBeHealed(PokemonCardSet defending){
     delayed {
       after EVOLVE, defending, {unregister()}
--- a/src/tcgwars/logic/groovy/TcgStatics.groovy
+++ b/src/tcgwars/logic/groovy/TcgStatics.groovy
@@ -1172,7 +1172,7 @@ class TcgStatics {
     def hasPokeCnt = 0
     [params.hasPokemonEX, params.hasPokemonGX, params.hasPokemonV].each{hasPokeCnt += 1}
 
-    def benchFilter = {
+    def areaFilter = {
       if (filter == null) { true } else {
         (
           if (params.hasType) {it.types.contains(params.hasType)} else {true}
@@ -1215,7 +1215,7 @@ class TcgStatics {
       failMessage = "${params.opp ? "Your opponent doesn't" : "You don't"} have any $pokeString that ${params.info ? params.info : "follow the stated condition(s)"}"
     }
 
-    assert checkedArea.any(benchFilter) : failMessage
+    assert checkedArea.any(areaFilter) : failMessage
   }
 
   static void assertMyPokemon(params=[:], Closure filter = null) {
--- a/src/tcgwars/logic/groovy/TcgStatics.groovy
+++ b/src/tcgwars/logic/groovy/TcgStatics.groovy
@@ -1157,13 +1157,13 @@ class TcgStatics {
    *   + hasType: If set, restricts to benched Pokémon of a single specific type.
    *   + hasPokemonEX/hasPokemonGX/hasPokemonV: Can be expanded if needed. All of these unset will have the method search for any Pokémon no matter what, but if even a single one is set true it'll only filter those that are set true as well.
    *   + opp: If true, checks for the opponent's bench instead of "my" bench.
+   *   + info: If set, it'll replace the end of the failed assert warning with a custom text, instead of the default "follow the stated condition(s)".
+   *   + repText: If true, params.info will override the entirety of the failed assert warning.
    *
    * @param c Additional condition the filtered benched Pokémon must follow. Defaults to true (so any benched Pokémon).
    *
-   * @param info Additional text to be put at the end of the assert fail warning. If repText is set to true in params, info will override the entirety of said warning.
-   *
    */
-  static void assertBench(params=[:], Closure c = null, String info = "") {
+  static void assertBench(params=[:], Closure c = null) {
     def failMessage
     def checkedBench = params.opp ? opp.bench : my.bench
 
@@ -1186,8 +1186,8 @@ class TcgStatics {
       }
     )
 
-    if (!params.repText) {
-      failMessage = info
+    if (params.info + !params.repText) {
+      failMessage = params.info
     } else {
       def typeString = param.hasType ? " ${params.hasType.getShortNotation()}" : ""
 
@@ -1207,16 +1207,16 @@ class TcgStatics {
         }
       )
 
-      failMessage = "${params.opp ? "Your opponent doesn't" : "You don't"} have any Benched $pokeString that ${!info.equals("") ? info : "follow the stated condition(s)"}"
+      failMessage = "${params.opp ? "Your opponent doesn't" : "You don't"} have any Benched $pokeString that ${params.info ? params.info : "follow the stated condition(s)"}"
     }
 
     assert checkedBench.any{benchFilter} : failMessage
   }
-  static void assertMyBench(params=[:], Closure c = null, String info = "") {
+  static void assertMyBench(params=[:], Closure c = null) {
     params.opp = false
     assertBench(params, c, cText)
   }
-  static void assertOppBench(params=[:], Closure c = null, String info = "") {
+  static void assertOppBench(params=[:], Closure c = null) {
     params.opp = true
     assertBench(params, c, cText)
   }
--- a/src/tcgwars/logic/groovy/TcgStatics.groovy
+++ b/src/tcgwars/logic/groovy/TcgStatics.groovy
@@ -1151,7 +1151,7 @@ class TcgStatics {
 
   /**
    *
-   * Does a customized assert with a automated fail message.
+   * Does a customized assert with an automated fail warning.
    *
    * @param params Optional settings that can be added:
    *   + isType: if set, restricts to benched Pokémon of a single specific type.
@@ -1160,7 +1160,7 @@ class TcgStatics {
    *
    * @param c Additional condition the filtered benched Pokémon must follow. Defaults to true (so any benched Pokémon).
    *
-   * @param message Additional text to be put at the end of the assert fail warning. If repText is set to true in params, message will override the entirety of said warning.
+   * @param info Additional text to be put at the end of the assert fail warning. If repText is set to true in params, info will override the entirety of said warning.
    *
    */
   static void assertBench(params=[:], Closure c = null, String info = "") {
--- a/src/tcgwars/logic/groovy/TcgStatics.groovy
+++ b/src/tcgwars/logic/groovy/TcgStatics.groovy
@@ -1157,7 +1157,7 @@ class TcgStatics {
    *   + benched: If true, checks for only Benched Pokémon; otherwise also includes the Active.
    *   + opp: If true, checks for the opponent's bench instead of "my" bench.
    *   + hasType: If set, restricts to benched Pokémon of a single specific type.
-   *   + hasPokemonEX/hasPokemonGX/hasPokemonV: Can be expanded if needed. All of these unset will have the method search for any Pokémon no matter what, but if even a single one is set true it'll only filter those that are set true as well.
+   *   + hasPokemonEX/hasPokemonGX/hasPokemonV/hasPokemonMAX: Can be expanded if needed. All of these unset will have the method search for any Pokémon no matter what, but if even a single one is set true it'll only filter those that are set true as well.
    *   + info: If set, it'll replace the end of the failed assert warning with a custom text, instead of the default "follow the stated condition(s)".
    *   + repText: If true, params.info will override the entirety of the failed assert warning.
    *
@@ -1180,7 +1180,8 @@ class TcgStatics {
           if (!hasPokeCnt) {true} else {
             (params.hasPokemonEX && it.pokemonEX) ||
             (params.hasPokemonGX && it.pokemonGX) ||
-            (params.hasPokemonV && it.pokemonV)
+            (params.hasPokemonV && it.pokemonV) ||
+            (params.hasPokemonVMAX && it.pokemonVMAX)
           }
         ) && (
           filter.call(it)
@@ -1201,7 +1202,8 @@ class TcgStatics {
         [
           (params.hasPokemonEX, "Pokémon-EX"),
           (params.hasPokemonGX, "Pokémon-GX"),
-          (params.hasPokemonV, "Pokémon V")
+          (params.hasPokemonV, "Pokémon V"),
+          (params.hasPokemonVMAX, "Pokémon VMAX")
         ].each{
           if (it[0]) {
             pokeString += it[1] + (if (i == isPokeCnt) "" else if (i == isPokeCnt-1) " or " else ", ")
--- a/src/tcgwars/logic/groovy/TcgStatics.groovy
+++ b/src/tcgwars/logic/groovy/TcgStatics.groovy
@@ -1163,7 +1163,7 @@ class TcgStatics {
    * @param c Additional condition the filtered benched Pokémon must follow. Defaults to true (so any benched Pokémon).
    *
    */
-  static void assertBench(params=[:], Closure c = null) {
+  static void assertBench(params=[:], Closure filter = null) {
     def failMessage
     def checkedBench = params.opp ? opp.bench : my.bench
 
@@ -1171,7 +1171,7 @@ class TcgStatics {
     [params.hasPokemonEX, params.hasPokemonGX, params.hasPokemonV].each{hasPokeCnt += 1}
 
     def benchFilter = (
-      if (c == null) { true } else {
+      if (filter == null) { true } else {
         (
           if (params.hasType) {it.types.contains(params.hasType)} else {true}
         ) && (
@@ -1181,7 +1181,7 @@ class TcgStatics {
             (params.hasPokemonV && it.pokemonV)
           }
         ) && (
-          c(it)
+          filter.call(it)
         )
       }
     )
@@ -1212,13 +1212,13 @@ class TcgStatics {
 
     assert checkedBench.any{benchFilter} : failMessage
   }
-  static void assertMyBench(params=[:], Closure c = null) {
+  static void assertMyBench(params=[:], Closure filter = null) {
     params.opp = false
-    assertBench(params, c)
+    assertBench(params, filter)
   }
-  static void assertOppBench(params=[:], Closure c = null) {
+  static void assertOppBench(params=[:], Closure filter = null) {
     params.opp = true
-    assertBench(params, c)
+    assertBench(params, filter)
   }
 
   static void cantBeHealed(PokemonCardSet defending){
--- a/src/tcgwars/logic/groovy/TcgStatics.groovy
+++ b/src/tcgwars/logic/groovy/TcgStatics.groovy
@@ -1156,7 +1156,7 @@ class TcgStatics {
   //      + isType: if set, restricts to benched Pokémon of a single specific type.
   //      + hasPokemonEX/hasPokemonGX/hasPokemonV: Can be expanded if needed. All of these unset will have the method search for any Pokémon no matter what, but if even a single one is set true it'll only filter those that are set true as well.
   //      + opp: If true, checks for the opponent's bench instead of "my" bench.
-  //      + repText: If true, instead of adding cText at the end of the assert it'll be the only thing printed.
+  //      + repText: If true, cText will override the entire fail message instead of being added at the end.
   static void assertBench(params=[:], Closure c = null, String cText = null) {
     def checkedBench = params.opp ? opp.bench : my.bench
 
--- a/src/tcgwars/logic/groovy/TcgStatics.groovy
+++ b/src/tcgwars/logic/groovy/TcgStatics.groovy
@@ -1220,12 +1220,12 @@ class TcgStatics {
     assert checkedArea.any(areaFilter) : failMessage
   }
 
-  static void assertMyPokemon(params=[:], Closure filter = null) {
+  static void assertMyAll(params=[:], Closure filter = null) {
     params.benched = false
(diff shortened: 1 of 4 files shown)


src/tcgwars/logic/groovy/TcgStatics.groovy
```diff
@@ -1153,6 +1153,42 @@ class TcgStatics {
...
+  //      + opp: If true, checks for the opponent's bench instead of "my" bench.
+  //      + repText: If true, instead of adding cText at the end of the assert it'll be the only thing printed.
+  static void checkAnyBench(params=[:], Closure c, String cText) {
+    def checkedBench = params.opp ? opp.bench : my.bench
+    assert (
+      if (c != null) { checkedBench.any{c(it)} } else { checkedBench }
+    ) : (
+      if (params.repText) { cText } else {
+        (
+          if (params.opp) { "You don't" } else { "Your opponent doesn't" }
```
flagrama, line 1172: This seems backwards to me. Shouldn't it be
```suggestion
          if (params.opp) { "Your opponent doesn't" } else { "You don't" } 
```

starg09:
> I don't agree with N's Resolve, but since it has a ruling I can't really argue against it.

Oh? What exactly do you mean @flagrama? Wasn't trying to change anything about N's resolve, tbh. But instead of such a long time as there already was, having a "types: N" handle that part of the assert fail message :P

flagrama:
> Oh? What exactly do you mean @flagrama? Wasn't trying to change anything about N's resolve, tbh. But instead of keeping as long a method call line as there already was, having a "types: N" handle that part of the assert fail message :P

Nothing wrong with what you wrote there. Would have placed a comment directly on it if that was the case. Just a little rant I guess:

Just the fact that you must have a benched [N] Pokémon since discarding cards is still technically an effect. I guess for the ruling it's the difference between being a prerequisite for playing the card at all and the effect being the text after that prerequisite. Just a bit confusing from just reading the card IMO.

starg09:
> Nothing wrong with what you wrote there. Would have placed a comment directly on it if that was the case. Just a little rant I guess:

Ah, alright :P

> Just the fact that you must have a benched [N] Pokémon since discarding cards is still technically an effect. I guess for the ruling it's the difference between being a prerequisite for playing the card at all and the effect being the text after that prerequisite. Just a bit confusing from just reading the card IMO.

Checking a bit... I agree, weird ruling but oh well lol. Not even JP card having the Benched dragon as a cost/requirement explicitely, but their FAQ responds in the same way as compendium so 🤷

flagrama:
I fear you may be getting to clever with this, but more importantly, I don't think this reads as an assert at all at a glance. Perhaps you could create the logic closures before the assert, and insert them into an assert at the end?

starg09:
> Perhaps you could create the logic closures before the assert, and insert them into an assert at the end?

Hmm, you might have a point. Was thinking it would make it slower, but everything inside the message is not really calculated so I'll move it above 👍

axpendix:
Hey, I think it's better to rename it to `assertBench()` because essentially it runs an `assert` at the end.

starg09:
Ok that should do it, just gotta test it all working fine on the couple edited cards before mass replacing :P

axpendix:
Hey, sorry for not mentioning this earlier but I think renaming the parameters to `hasPokemonGX`, `hasPokemonEX`, `hasType`, `has...` will be more meaningful. 

Also note that we have `PokemonCardSet.[pokemonGX|pokemonEX|EX]` so I think they should be consistent with it.

src/tcgwars/logic/groovy/TcgStatics.groovy
```diff
@@ -1146,6 +1146,72 @@ class TcgStatics {
...
+    assert benchFilter : failMessage
+  }
+  static void assertMyBench(params=[:], Closure c, String cText) {
+    params.opp = false
+    assertBench(params, c, cText)
+  }
+  static void assertOppBench(params=[:], Closure c, String cText) {
+    params.opp = true
+    assertBench(params, c, cText)
+  }
```
axpendix, line 1213: I don't understand why cText is a required parameter here

src/tcgwars/logic/groovy/TcgStatics.groovy
```diff
@@ -1146,6 +1146,72 @@ class TcgStatics {
...
+  /* General checks for attacks and abilities */
+
+  //  * assertBench
+  //    - Can be used with no arguments, or with a condition and additional text included.
+  //    - If a closure is given, it'll check for any "it" element in bench returning true.
+  //    - Optional params:
+  //      + type: if set, restricts to benched Pokémon of a single specific type.
+  //      + isEX/isGX/isV: Can be expanded if needed. All of these unset will have the method search for any Pokémon no matter what, but if even a single one is set true it'll only filter those that are set true as well.
+  //      + opp: If true, checks for the opponent's bench instead of "my" bench.
+  //      + repText: If true, instead of adding cText at the end of the assert it'll be the only thing printed.
```
axpendix, line 1159: I don't get the purpose of `repText`, can you rewrite the explanation it in a better way?

src/tcgwars/logic/groovy/TcgStatics.groovy
starg09: It's not, though the check is done for null inside the method. Could probably use `.equals("")` though.

src/tcgwars/logic/groovy/TcgStatics.groovy
starg09: It allows for a more specific assert fail message, just as a possibility. Should cover way too specific filters better (checking it, it used to have more use before adding the "type" param, since then it'd have said "valid Pokémon that" +"are [N] type." instead of "valid [N] Pokémon"). No immediate use right now but doesn't hurt to keep :P

src/tcgwars/logic/groovy/TcgStatics.groovy
axpendix: In this case the method won't run without a default value assignment. Change the parameter to `String cText = null`

src/tcgwars/logic/groovy/TcgStatics.groovy
axpendix: So it allows to override the assertion fail message completely or partially?

axpendix: Ok just checked the logic, it allows override. But I would do it differently. 

First of all, I don't get the value of a partially overriding `cText`. 
It should be either a) all auto-generated b) overriden manually. 

Auto-generation can worked on to produce better strings with multiple conditions as well, instead of "follow the stated condition(s)" clause. 

If auto-generated message is worked out to provide a good message, there would be no need for a manual override. 

If you think, manual override is still needed, the field name should be `message`. When set, it should override no matter what, so no need for `cText` and `repText` fields, IMO.

starg09: > So it allows to override the assertion fail message completely or partially?

Yeah, that'd be the gist of it. Updated the description a bit, wasn't very clear looking at it again.

src/tcgwars/logic/groovy/TcgStatics.groovy
```diff
@@ -1146,6 +1146,72 @@ class TcgStatics {
...
+      }
+    )
+
+    def failMessage = if (params.repText) { cText } else {
+      "You" + (params.opp ? "r opponent does" : " do") + "n't have any " + (
+        param.type ? params.type.getShortNotation() + " " : ""
+      ) + "Benched $pokeString" + () + (
+        if (c != null) {
+          " that " + (cText ? cText : "follow the stated condition(s)")
+        }
```
axpendix: Please type full strings without splitting them up for english grammar like this. It will be more readable/maintainable. 

Also, please use groovy Gstring feature - use variable interpolation with `$` instead of concetenation with `+`. This will also produce more readable/maintainable code.

Thanks

src/tcgwars/logic/groovy/TcgStatics.groovy
```diff
@@ -1146,6 +1146,72 @@ class TcgStatics {
...
+  /* General checks for attacks and abilities */
+
+  //  * assertBench
+  //    - Can be used with no arguments, or with a condition and additional text included.
+  //    - If a closure is given, it'll check for any "it" element in bench returning true.
+  //    - Optional params:
+  //      + isType: if set, restricts to benched Pokémon of a single specific type.
+  //      + hasPokemonEX/hasPokemonGX/hasPokemonV: Can be expanded if needed. All of these unset will have the method search for any Pokémon no matter what, but if even a single one is set true it'll only filter those that are set true as well.
+  //      + opp: If true, checks for the opponent's bench instead of "my" bench.
+  //      + repText: If true, cText will override the entire fail message instead of being added at the end.
```
axpendix, line 1159: Please use JavaDoc / GroovyDoc style to write comments for methods

axpendix: If you use an IDE they usually autogenerate it when you type `/**` and press return.

src/tcgwars/logic/groovy/TcgStatics.groovy
starg09: Taking a bit to reply, enabled Github's beta redesign and comments don't autoupdate as usual >.>

src/tcgwars/logic/groovy/TcgStatics.groovy
starg09: > Please type full strings without splitting them up for english grammar like this. It will be more readable/maintainable.

Fair point, might be longer but also simpler to follow.

> Also, please use groovy Gstring feature - use variable interpolation with $ instead of concetenation with +. This will also produce more readable/maintainable code.

Initially went with it, but for some reason (I think a recurring error while working on some card implementations) I jumped back into concatenation. Will update it in a bit.

src/tcgwars/logic/groovy/TcgStatics.groovy
axpendix: I quite liked the design change! (but not sure about the push updates yet)

src/tcgwars/logic/groovy/TcgStatics.groovy
starg09: Tried to update this, tho I'm not sure if it follows GroovyDocs 100% in the params section.

src/tcgwars/logic/groovy/TcgStatics.groovy
```diff
@@ -1149,15 +1149,21 @@ class TcgStatics {
...
+   *   + isType: if set, restricts to benched Pokémon of a single specific type.
+   *   + hasPokemonEX/hasPokemonGX/hasPokemonV: Can be expanded if needed. All of these unset will have the method search for any Pokémon no matter what, but if even a single one is set true it'll only filter those that are set true as well.
+   *   + opp: If true, checks for the opponent's bench instead of "my" bench.
+   *
+   * @param c Additional condition the filtered benched Pokémon must follow. Defaults to true (so any benched Pokémon).
+   *
+   * @param message Additional text to be put at the end of the assert fail warning. If repText is set to true in params, message will override the entirety of said warning.
+   *
+   */
+  static void assertBench(params=[:], Closure c = null, String message = "") {
```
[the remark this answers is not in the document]
starg09: Hmm, I put it like that based on other commands using info as an explicit parameter. Which looking at it might be a better name for it since it won't be a full message 99% of the time.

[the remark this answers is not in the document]
starg09: I just pushed a correction on that, also ignoring all the type and pokemon logic if the warning is fully overrode by `repText: true`. Will update the doc in a bit.

src/tcgwars/logic/groovy/TcgStatics.groovy
```diff
@@ -1185,34 +1186,37 @@ class TcgStatics {
       }
     )
 
-    def typeString = param.isType ? " ${params.isType.getShortNotation()}" : ""
+    if (!params.repText) {
+      failMessage = info
+    } else {
+      def typeString = param.isType ? " ${params.isType.getShortNotation()}" : ""
```
bekce, line 1192: Should be `hasType` to be consistent

starg09:
Just leaving it noted. The update for `isPokemonV` follows this ruling:

> **Fighting Fangs**    _(Boltund - Sword & Shield)_
>
> Q. Are Pokemon VMAX considered as "Pokemon V" for the purposes of Boltund's "Fighting Fangs" attack?
>
> A. Yes, Pokemon VMAX are always considered to be Pokemon V. (May 7, 2020 TPCi Rules Team)

starg09:
@axpendix just to confirm: Is `it.pokemonV` and `it.pokemonVMAX` something that'll work with those additions in `PokemonCardSet.java` alone, or do they need to be added somewhere else?

axpendix:
> @axpendix just to confirm: Is `it.pokemonV` and `it.pokemonVMAX` something that'll work with those additions in `PokemonCardSet.java` alone, or do they need to be added somewhere else?

They will work

src/tcgwars/logic/groovy/TcgStatics.groovy
```diff
@@ -1146,6 +1146,81 @@ class TcgStatics {
...
+    assert checkedBench.any{benchFilter} : failMessage
+  }
+  static void assertMyBench(params=[:], Closure c = null, String info = "") {
+    params.opp = false
+    assertBench(params, c, cText)
+  }
+  static void assertOppBench(params=[:], Closure c = null, String info = "") {
+    params.opp = true
+    assertBench(params, c, cText)
+  }
```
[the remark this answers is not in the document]
axpendix: You are correct in the sense that many methods accept info as a positional parameter but I think that should also change in the long term as it often causes mistakes during development. This one, as you said, won't have an immediate use case so it's best if we start with it.

src/tcgwars/logic/groovy/TcgStatics.groovy
```diff
@@ -1146,6 +1146,81 @@ class TcgStatics {
...
   }
+
+  /* General checks for attacks and abilities */
+
+  /**
+   *
+   * Does a customized assert with an automated fail warning.
+   *
+   * @param params Optional settings that can be added:
+   *   + isType: if set, restricts to benched Pokémon of a single specific type.
```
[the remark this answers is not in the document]
starg09: Yeah, being a single made me go with is but has should work fine too.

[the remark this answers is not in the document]
axpendix: Thank you!

src/tcgwars/logic/groovy/TcgStatics.groovy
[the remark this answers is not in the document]
starg09: Aight, will update in a bit 👍

src/tcgwars/logic/groovy/TcgStatics.groovy
```diff
@@ -1146,6 +1146,81 @@ class TcgStatics {
...
+
+    assert checkedBench.any{benchFilter} : failMessage
+  }
+  static void assertMyBench(params=[:], Closure c = null) {
+    params.opp = false
+    assertBench(params, c, cText)
+  }
+  static void assertOppBench(params=[:], Closure c = null) {
+    params.opp = true
+    assertBench(params, c, cText)
```
axpendix, line 1221: `cText` won't resolve

src/tcgwars/logic/groovy/TcgStatics.groovy
```diff
@@ -1146,6 +1146,81 @@ class TcgStatics {
...
+        ) && (
+          c(it)
+        )
+      }
+    )
+
+    if (params.info + !params.repText) {
+      failMessage = params.info
+    } else {
+      def typeString = param.hasType ? " ${params.hasType.getShortNotation()}" : ""
```
axpendix, line 1192:
```suggestion
      def typeString = param.hasType ? " ${params.hasType}" : ""
```

src/tcgwars/logic/groovy/TcgStatics.groovy
```diff
@@ -1146,6 +1146,81 @@ class TcgStatics {
...
+   *   + hasType: If set, restricts to benched Pokémon of a single specific type.
+   *   + hasPokemonEX/hasPokemonGX/hasPokemonV: Can be expanded if needed. All of these unset will have the method search for any Pokémon no matter what, but if even a single one is set true it'll only filter those that are set true as well.
+   *   + opp: If true, checks for the opponent's bench instead of "my" bench.
+   *   + info: If set, it'll replace the end of the failed assert warning with a custom text, instead of the default "follow the stated condition(s)".
+   *   + repText: If true, params.info will override the entirety of the failed assert warning.
+   *
+   * @param c Additional condition the filtered benched Pokémon must follow. Defaults to true (so any benched Pokémon).
+   *
+   */
+  static void assertBench(params=[:], Closure c = null) {
```
axpendix, line 1166:
```suggestion
  static void assertBench(params=[:], Closure filter = null) {
```

src/tcgwars/logic/groovy/TcgStatics.groovy
```diff
@@ -1146,6 +1146,81 @@ class TcgStatics {
...
+    def failMessage
+    def checkedBench = params.opp ? opp.bench : my.bench
+
+    def hasPokeCnt = 0
+    [params.hasPokemonEX, params.hasPokemonGX, params.hasPokemonV].each{hasPokeCnt += 1}
+
+    def benchFilter = (
+      if (c == null) { true } else {
```
axpendix, line 1174:
```suggestion
      if (filter == null) { true } else {
```

src/tcgwars/logic/groovy/TcgStatics.groovy
```diff
@@ -1146,6 +1146,81 @@ class TcgStatics {
...
+        (
+          if (params.hasType) {it.types.contains(params.hasType)} else {true}
+        ) && (
+          if (!hasPokeCnt) {true} else {
+            (params.hasPokemonEX && it.pokemonEX) ||
+            (params.hasPokemonGX && it.pokemonGX) ||
+            (params.hasPokemonV && it.pokemonV)
+          }
+        ) && (
+          c(it)
```
axpendix, line 1184:
```suggestion
          filter.call(it)
```

src/tcgwars/logic/groovy/TcgStatics.groovy
```diff
@@ -1146,6 +1146,81 @@ class TcgStatics {
...
+            pokeString += it[1] + (if (i == isPokeCnt) "" else if (i == isPokeCnt-1) " or " else ", ")
+            i += 1
+          }
+        }
+      )
+
+      failMessage = "${params.opp ? "Your opponent doesn't" : "You don't"} have any Benched $pokeString that ${params.info ? params.info : "follow the stated condition(s)"}"
+    }
+
+    assert checkedBench.any{benchFilter} : failMessage
```
axpendix, line 1213:
```
benchFilter = {
...
}

assert checkedBench.any(benchFilter) : failMessage
```

src/tcgwars/logic/groovy/TcgStatics.groovy
```diff
@@ -1146,6 +1146,81 @@ class TcgStatics {
...
+  }
+  static void assertMyBench(params=[:], Closure c = null) {
+    params.opp = false
+    assertBench(params, c)
+  }
+  static void assertOppBench(params=[:], Closure c = null) {
+    params.opp = true
+    assertBench(params, c)
+  }
```
starg09, line 1222:
```suggestion
  static void assertMyBench(params=[:], Closure filter = null) {
    params.opp = false
    assertBench(params, filter)
  }
  static void assertOppBench(params=[:], Closure filter = null) {
    params.opp = true
    assertBench(params, filter)
  }
```

src/tcgwars/logic/groovy/TcgStatics.groovy
starg09: Just to confirm: You're suggesting benchFilter does not require a def? not sure I'm fully following here.

src/tcgwars/logic/groovy/TcgStatics.groovy
starg09: Also just to confirm: Would the Type passed as param be printed as [x] even without short notation? Checked a couple cards for reference but they all had hardcoded texts, The one case I saw without them was Porygon from TeamRocketNG, and it printed R as "FIRE", same with all others.

src/tcgwars/logic/groovy/TcgStatics.groovy
axpendix: No, it should be def again. what i meant was the curly vs square bracket change.

starg09: ahh, ok haha

starg09, line 1213:
```suggestion
    assert checkedBench.any(benchFilter) : failMessage
```

src/tcgwars/logic/groovy/TcgStatics.groovy
axpendix: yes that will be good, that's what I want, print in full letters. gives greater visibility.

starg09: Hmm, ok then 👍 

Personally more of a fan of short since it's used in so many places, but is not necessary at all.

starg09:
Turning to draft for one change I thought while working on some Darkness Ablaze cards. This could very easily cover cases checking for any Pokémon, active included (right now, Rose looking for a VMAX in play by the one using it). Will update in a bit.

axpendix:
Closing brackets are wrong

axpendix:
Closing brackets are wrong

src/tcgwars/logic/groovy/TcgStatics.groovy
```diff
@@ -1146,6 +1146,81 @@ class TcgStatics {
...
+   *
+   */
+  static void assertBench(params=[:], Closure filter = null) {
+    def failMessage
+    def checkedBench = params.opp ? opp.bench : my.bench
+
+    def hasPokeCnt = 0
+    [params.hasPokemonEX, params.hasPokemonGX, params.hasPokemonV].each{hasPokeCnt += 1}
+
+    def benchFilter = (
```
axpendix, line 1173: No this wont work like that. The closure should beiterated over each bench pokemon.

starg09:
> Closing brackets are wrong

Yup, noticed while fixing the message section :P

src/tcgwars/logic/groovy/TcgStatics.groovy
starg09: Ok, not sure what went through my head on those >.>

src/tcgwars/logic/groovy/TcgStatics.groovy
```diff
@@ -1146,6 +1146,102 @@ class TcgStatics {
...
+  static void assertMyBenched(params=[:], Closure filter = null) {
+    params.benched = true
+    params.opp = false
+    assertAnyPokemonInPlay(params, filter)
+  }
+  static void assertOppBenched(params=[:], Closure filter = null) {
+    params.benched = true
+    params.opp = true
+    assertAnyPokemonInPlay(params, filter)
+  }
```
axpendix, line 1243: Better rewritten as

src/tcgwars/logic/groovy/TcgStatics.groovy
```diff
@@ -1146,6 +1146,102 @@ class TcgStatics {
...
+        ) && (
+          filter.call(it)
+        )
+      }
+    }
+
+    if (params.info + !params.repText) {
+      failMessage = params.info
+    } else {
+      def benchedString = (param.benched ? "Benched " : "")
```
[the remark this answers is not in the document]
starg09: Now the general method can handle asserts for `my.all` and `opp.all`. The bottom 4 methods should be used preferably, which force the benched and opp params.

src/tcgwars/logic/groovy/TcgStatics.groovy
```diff
@@ -1146,6 +1146,102 @@ class TcgStatics {
...
+          if (it[0]) {
+            pokeString += it[1] + (if (i == isPokeCnt) "" else if (i == isPokeCnt-1) " or " else ", ")
+            i += 1
+          }
+        }
+      } else {
+        pokeString += "Pokémon"
+      }
+
+      failMessage = "${params.opp ? "Your opponent doesn't" : "You don't"} have any $pokeString that ${params.info ? params.info : "follow the stated condition(s)"}"
```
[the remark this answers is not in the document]
starg09: What part of it specifically? The part about Pokémon is concatenated so it's written properly, rest uses gstring as asked :/

src/tcgwars/logic/groovy/TcgStatics.groovy
starg09: Was there a code suggestion here? Didn't get submitted

axpendix:
```suggestion
  static void assertMyAll(params=[:], Closure filter = null) {
    params.benched = false
    params.opp = false
    assertAnyPokemonInPlay(params, filter)
  }
  static void assertOppAll(params=[:], Closure filter = null) {
    params.benched = false
    params.opp = true
    assertAnyPokemonInPlay(params, filter)
  }

  static void assertMyBenched(params=[:], Closure filter = null) {
    params.benched = true
    params.opp = false
    assertAnyPokemonInPlay(params, filter)
  }
  static void assertOppBenched(params=[:], Closure filter = null) {
    params.benched = true
    params.opp = true
    assertAnyPokemonInPlay(params, filter)
  }
```

starg09: Hmm, don't really like how it sounds but it's probably better than My/Opp Pokemon (which is a bit too vague, and PokemonInPlay for these would be a bit too much)

src/tcgwars/logic/groovy/TcgStatics.groovy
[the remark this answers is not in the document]
axpendix: The areaFilter won't work correctly. Reiterate on the logic.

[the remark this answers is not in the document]
starg09: Ah, a different section of code got marked. All I saw here was the `failMessage` line and final 2 of the `pokeString`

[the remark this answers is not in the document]
axpendix: `if (filter == null) { true } ` this is incorrect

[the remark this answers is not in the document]
axpendix: > Ah, a different section of code got marked. All I saw here was the `failMessage` line and final 2 of the `pokeString`

I am sorry yeah I must have clicked at wrong line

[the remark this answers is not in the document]
starg09: Just pushed a rewrite of this, tested in a vacuum compiler seemed to return true and false properly now.

src/tcgwars/logic/groovy/TcgStatics.groovy
```diff
@@ -1146,6 +1146,100 @@ class TcgStatics {
...
```
axpendix, line 1215: Update this message as well/ it doesn't make any use of `typeString` or `benchString` !!

axpendix: Also the way it uses `params.info` is wrong. It shouldn't use it

starg09: Hmm, left them concatenated into pokeString since otherwise it'd be `${benchString}${params.benched && params.hasType ? " " ; ""}${typeString}`. Can move them down here I guess?

starg09: I take this back, actually solved that earlier. `${benchString}${typeString}` doesn't look **_that_** bad

starg09: And final push looks even better (and apparently working well).

src/tcgwars/logic/groovy/TcgStatics.groovy
```diff
@@ -1146,6 +1146,99 @@ class TcgStatics {
...
+          if (it[0]) {
+            pokeString += it[1] + (if (i == isPokeCnt) "" else if (i == isPokeCnt-1) " or " else ", ")
+            i += 1
+          }
+        }
+      } else {
+        pokeString += "Pokémon"
+      }
+
+      failMessage = "${params.opp ? "Your opponent doesn't" : "You don't"} have any ${benchString + typeString + pokeString} that ${params.info ? params.info : "follow the stated condition(s)"}"
```
axpendix, line 1214: params.info still here...

axpendix, line 1214: params.info still here...

starg09: Missed the params.info part earlier... Are you suggesting to move it to it's own variable as with benchString/typeString/pokeString? Or just removing the middle value and make it `params.info ?: "follow[...]"`?

axpendix: just take a look at the beginning of this if/else clause and you'll see why it has to be removed

starg09: Ahh, ok there was a typo on the if itself. It's meant to be and && not a +.

If params.info is not set, it should generate the pre-formatted failMessage no matter what. If it's set _and_ params.repText is enabled then params.info is used alone, otherwise it replaces the final section of the pre-formatted message.

src/tcgwars/logic/groovy/TcgStatics.groovy
```diff
@@ -1146,6 +1146,99 @@ class TcgStatics {
...
+          !hasPokeCnt || (
+              (params.hasPokemonEX && it.pokemonEX) ||
+              (params.hasPokemonGX && it.pokemonGX) ||
+              (params.hasPokemonV && it.pokemonV) ||
+              (params.hasPokemonVMAX && it.pokemonVMAX)
+          )
+      ) && (
+          filter == null || filter.call(it)
+      )
+    )
```
axpendix, line 1188:
```suggestion
    def areaFilter = {
      (
          !params.hasType || it.types.contains(params.hasType)
      ) && (
          !hasPokeCnt || (
              (params.hasPokemonEX && it.pokemonEX) ||
              (params.hasPokemonGX && it.pokemonGX) ||
              (params.hasPokemonV && it.pokemonV) ||
              (params.hasPokemonVMAX && it.pokemonVMAX)
          )
      ) && (
          filter == null || filter.call(it)
      )
    }
```

axpendix:
This was a long run!

starg09:
It sure was, haha
Hopefully next PRs like this go a bit smoother 😅

axpendix:
There are errors. I am fixing them

starg09:
Oh ok I'll close my PR then, not sure if it fixed build though.

PR #582, opened 2020-06-07
Implementations

Grayahm:
Implemented:

- Snowpoint Temple
- Azelf Lv.X
- Mesprit Lv.X
- Regigigas Lv.X
- Dialga G
- Dialga G Lv.X

### DIFF
--- a/src/tcgwars/logic/impl/gen4/LegendsAwakened.groovy
+++ b/src/tcgwars/logic/impl/gen4/LegendsAwakened.groovy
@@ -3125,6 +3125,7 @@ public enum LegendsAwakened implements LogicCardInfo {
             energyCost ()
             attackRequirement {}
             onAttack {
+              assert my.bench : "Your bench is empty"
               my.bench.each{
                 heal 30, it
               }
@@ -3136,7 +3137,7 @@ public enum LegendsAwakened implements LogicCardInfo {
             attackRequirement {my.all.findAll{it.name == ("Uxie Lv.X")} && my.all.findAll{it.name == ("Uxie Lv.X")}}
             onAttack {
               damage 200
-              discardAllEnergySelf(null)
+              discardAllEnergySelf()
             }
           }
           move "", {
@@ -3213,10 +3214,12 @@ public enum LegendsAwakened implements LogicCardInfo {
             actionA {
               checkNoSPC()
               checkLastTurn()
+              assert bg.em().retrieveObject("Trade_Off") != bg.turnCount : "You cannot use more than one Trade Off per turn"
               assert my.deck : "Your deck is empty"
+              bg.em().storeObject("Trade_Off",bg.turnCount)
               powerUsed()
               def sel = my.deck.subList(0,2).select("Choose 1 card to put into your hand")
-              my.deck.subList(0,2).getExcludedList(sel).moveTo(suppressLog: true, my.deck)
+              my.deck.subList(0,2).getExcludedList(sel).moveTo(hidden: true, my.deck)
               sel.moveTo(hidden: true, my.hand)          
             }
           }
@@ -3226,7 +3229,7 @@ public enum LegendsAwakened implements LogicCardInfo {
             attackRequirement {}
             onAttack {
               damage 60
-              cantAttackNextTurn self
+              cantUseAttack thisMove, self
             }
           }
           move "", {
--- a/src/tcgwars/logic/impl/gen4/LegendsAwakened.groovy
+++ b/src/tcgwars/logic/impl/gen4/LegendsAwakened.groovy
@@ -2980,10 +2980,16 @@ public enum LegendsAwakened implements LogicCardInfo {
         return basicTrainer (this) {
           text "This card stays in play when you play it. Discard this card if another Stadium card comes into play. If another card with the same name is in play, you can’t play this card.\nEach Pokémon that isn’t an Evolved Pokémon in play (both your and your opponent’s gets +20 HP."
           onPlay {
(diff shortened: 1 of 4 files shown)


src/tcgwars/logic/impl/gen4/LegendsAwakened.groovy
```diff
@@ -3112,15 +3125,18 @@ public enum LegendsAwakened implements LogicCardInfo {
             energyCost ()
             attackRequirement {}
```
flagrama, line 3126: I'd probably put a `assert my.bench` here as without a benched Pokémon this attack would have no effect.

src/tcgwars/logic/impl/gen4/LegendsAwakened.groovy
```diff
@@ -3112,15 +3125,18 @@ public enum LegendsAwakened implements LogicCardInfo {
...
           move "Supreme Blast", {
-            text "200 damage. in play, this attack does nothing. Discard all Energy attached to Mesprit."
+            text "200 damage. If you don't have Uxie LV.X and Azelf LV.X in play, this attack does nothing. Discard all Energy attached to Mesprit."
             energyCost P, P
-            attackRequirement {}
+            attackRequirement {my.all.findAll{it.name == ("Uxie Lv.X")} && my.all.findAll{it.name == ("Uxie Lv.X")}}
             onAttack {
-              damage 0
+              damage 200
+              discardAllEnergySelf(null)
```
flagrama, line 3139: The method is `discardAllSelfEnergy()` and you don't need to include the `null` as it has null as the default value.

src/tcgwars/logic/impl/gen4/LegendsAwakened.groovy
```diff
@@ -3195,14 +3211,22 @@ public enum LegendsAwakened implements LogicCardInfo {
           pokePower "Trade Off", {
             text "Once during your turn , you may look at the top 2 cards of your deck, choose 1 of them, and put it into your hand. Put the other card on the bottom of your deck. This power can’t be used if Uxie is affected by a Special Condition. You can’t use more than 1 Trade Off Poké-Power each turn."
             actionA {
+              checkNoSPC()
+              checkLastTurn()
+              assert my.deck : "Your deck is empty"
+              powerUsed()
+              def sel = my.deck.subList(0,2).select("Choose 1 card to put into your hand")
+              my.deck.subList(0,2).getExcludedList(sel).moveTo(suppressLog: true, my.deck)
```
flagrama, line 3219: I don't see `suppressLog` anywhere else in the codebase. Are you sure `hidden` won't do what you want here like it does when moving to your hand?

src/tcgwars/logic/impl/gen4/LegendsAwakened.groovy
```diff
@@ -3195,14 +3211,22 @@ public enum LegendsAwakened implements LogicCardInfo {
           pokePower "Trade Off", {
```
flagrama, line 3211: This power isn't fully implemented. You can't use more than 1 Trade Off per turn, so you have to make sure other Pokémon with the same power can't use it. Check the Aegislash in RebelClash for an example on bwAbility.

src/tcgwars/logic/impl/gen4/LegendsAwakened.groovy
```diff
@@ -3195,14 +3211,22 @@ public enum LegendsAwakened implements LogicCardInfo {
...
             }
           }
           move "Zen Blade", {
             text "60 damage. Uxie can’t use Zen Blade during your next turn."
             energyCost C, C
             attackRequirement {}
             onAttack {
-              damage 0
+              damage 60
+              cantAttackNextTurn self
```
flagrama, line 3229: According to the text, Uxie can still attack, it just can't use this same attack.
```suggestion
              cantUseAttack thisMove, self
```

src/tcgwars/logic/impl/gen4/LegendsAwakened.groovy
Grayahm: As in `.moveTo(hidden, my.deck)` ?

src/tcgwars/logic/impl/gen4/LegendsAwakened.groovy
Grayahm: Pretty sure I fixed this correctly now

flagrama: Yeah, I think it looks good now.

src/tcgwars/logic/impl/gen4/LegendsAwakened.groovy
```diff
@@ -3112,15 +3125,19 @@ public enum LegendsAwakened implements LogicCardInfo {
...
           move "Supreme Blast", {
-            text "200 damage. in play, this attack does nothing. Discard all Energy attached to Mesprit."
+            text "200 damage. If you don't have Uxie LV.X and Azelf LV.X in play, this attack does nothing. Discard all Energy attached to Mesprit."
             energyCost P, P
-            attackRequirement {}
+            attackRequirement {my.all.findAll{it.name == ("Uxie Lv.X")} && my.all.findAll{it.name == ("Uxie Lv.X")}}
             onAttack {
-              damage 0
+              damage 200
+              discardAllEnergySelf()
```
flagrama, line 3140: The order of Energy and Self is still incorrect. `discardAllEnergySelf` may make more sense using english rules, but `discardAllSelfEnergy` is actually in the codebase.

Grayahm:
Think everything is fixed

PR #583, opened 2020-06-08
EX Deoxys set fixes

starg09:
Open to feedback, multiple minor fixes, but some are more substantial.

Will undraft it once I finish checking the other 30~ cards in the set.

* Regarding Master Ball (EX 88):
  - This is a weird one. On one hand, the EX Compendium has a 2005 ruling saying that picking a Pokémon card found among the top 7 cards was mandatory:
    > Q. The wording on the new Deoxys verison of "Master Ball" says, "...Choose a Basic Pokémon or Evolution card...". However, the wording on the older versions say, "...you MAY choose..." Does this mean now that if you draw a Pokémon you MUST take it, or should it still be a choice?
    >
    > A. Sorry, you no longer get to choose. You MUST take a Pokémon if you draw at least one. (Apr 14, 2005 PUI Rules Team)
  - **_However,_** deck content is private knowledge, and future rulings from the DP Era (while this card was still legal, alongside a Croconaw from MT with similar oddly-written text) went against this concept:
    > Q. If I use Pokemon Communicator to put a Pokemon into my deck, am I allowed to voluntarily fail and choose to take no Pokemon out of my deck?
    > 
    > A. The contents of your deck is not public knowledge so the deck search can fail, even if you've just put a pokemon into the deck. (Feb 4, 2010 PUI Rules Team)
  - More specifically, though not using this as a main source, on the "Ask the Rules Team" from the PokéGym forums, there are multiple instances ([1 in particular](https://pokegym.net/community/index.php?threads/can-we-not-take-what-isnt-public-knoledge.79279/); [1 regarding Master Ball, Croconaw and deck searches in general](https://pokegym.net/community/index.php?threads/croconaw-mt-master-ball-question.82120/)) saying these searches can fail:
    > **Blaziken 1111**> Sorry if this has already been ruled but I looked everywhere and couldn't find a rulling. Say I play master ball and there is a pokemon. I don't want to take anything for what ever reason. Since it isn't public knoledge do I have to take it. Is this cheating or is this leagle. _(Aug 3, 2008)_
    >
    > **PokePop**> It's legal. What is in your deck is private. _(Aug 6, 2008)_
  - Based on all this, the commit makes Master Ball behave like current Great Ball, just like it worked in the Gym Challenge and Expedition prints. The player using the trainer can opt not to pick a Pokémon even if one is present, since deck is private knowledge.

### DIFF
--- a/src/tcgwars/logic/impl/gen3/Deoxys.groovy
+++ b/src/tcgwars/logic/impl/gen3/Deoxys.groovy
@@ -3010,7 +3010,7 @@ public enum Deoxys implements LogicCardInfo {
             onAttack {
               def pcs = defending
               if(opp.bench){
-                if(confirm("Switch 1 of your opponent’s Benched Pokémon with the Defending Pokémon.")){
+                if(confirm("Switch 1 of your opponent’s Benched Pokémon with the Defending Pokémon before doing damage?")){
                   pcs = opp.bench.select()
                   sw defending, pcs
                 }
@@ -3027,7 +3027,10 @@ public enum Deoxys implements LogicCardInfo {
               damage 60
               if(confirm("Discard a [D] Energy attached to Sharpedo ex? You will do 20 more damage and discard 1 Energy card attached to the Defending Pokémon.")){
                 damage 20
-                discardDefendingEnergy()
+                afterDamage{
+                  discardSelfEnergy D
+                  discardDefendingEnergy()
+                }
               }
             }
           }
--- a/src/tcgwars/logic/impl/gen3/Deoxys.groovy
+++ b/src/tcgwars/logic/impl/gen3/Deoxys.groovy
@@ -568,7 +568,9 @@ public enum Deoxys implements LogicCardInfo {
             onAttack {
               damage 30
               afterDamage {
-                opp.hand.oppSelect(count: opp.hand.size() - 4, "Select cards to discard until you end up with ${opp.hand.size() - 4} left").discard()
+                if(opp.hand.size() >= 5){
+                  opp.hand.oppSelect(count: opp.hand.size() - 4, "Select cards to discard until you end up with 4 left").discard()
+                }
               }
             }
           }
--- a/src/tcgwars/logic/impl/gen3/Deoxys.groovy
+++ b/src/tcgwars/logic/impl/gen3/Deoxys.groovy
@@ -1728,7 +1728,8 @@ public enum Deoxys implements LogicCardInfo {
       case BAGON_52:
         return basic (this, hp:HP050, type:COLORLESS, retreatCost:1) {
           weakness COLORLESS
-          resistance FIGHTING
+          resistance FIRE, MINUS30
+          resistance FIGHTING, MINUS30
           move "Singe", {
             text "Flip a coin. If heads, the Defending Pokémon is now Burned."
             energyCost R
--- a/src/tcgwars/logic/impl/gen3/Deoxys.groovy
+++ b/src/tcgwars/logic/impl/gen3/Deoxys.groovy
@@ -1272,10 +1272,7 @@ public enum Deoxys implements LogicCardInfo {
             attackRequirement {}
             onAttack {
               damage 10
-              if(opp.hand){
-                opp.hand.select(hidden:true,"Select one card").showToMe("Selected card.").moveTo(opp.deck)
-                shuffleDeck(null,TargetPlayer.OPPONENT)
-              }
+              afterDamage { astonish() }
             }
           }
 
--- a/src/tcgwars/logic/impl/gen3/Deoxys.groovy
+++ b/src/tcgwars/logic/impl/gen3/Deoxys.groovy
@@ -2690,7 +2690,7 @@ public enum Deoxys implements LogicCardInfo {
               checkNoSPC()
               assert self.active : "$self is not your Active Pokémon"
               powerUsed()
-              directDamage 10, opp.all.select()
+              directDamage 10, opp.all.select(), SRC_ABILITY
             }
           }
           move "Cross Attack", {
--- a/src/tcgwars/logic/impl/gen3/Deoxys.groovy
+++ b/src/tcgwars/logic/impl/gen3/Deoxys.groovy
@@ -846,7 +846,7 @@ public enum Deoxys implements LogicCardInfo {
               checkNoSPC()
               assert my.deck : "There is no card in your deck"
               powerUsed()
-              def tar = my.deck.search(max:1,"Select 1 card",{true}).each{my.deck.remove(it)}
+              def tar = my.deck.select("Select 1 card").each{my.deck.remove(it)}
               shuffleDeck()
               my.deck.addAll(0, tar)
             }
--- a/src/tcgwars/logic/impl/gen3/Deoxys.groovy
+++ b/src/tcgwars/logic/impl/gen3/Deoxys.groovy
@@ -3133,9 +3133,9 @@ public enum Deoxys implements LogicCardInfo {
             text "Whenever you attach a [D] Energy card from your hand to Rocket’s Raikou ex, you may choose 1 of the Defending Pokémon and switch it with 1 of your opponent’s Benched Pokémon. Your opponent chooses the Benched Pokémon to switch. This power can’t be used if Rocket’s Raikou ex is affected by a Special Condition."
             delayedA {
               after ATTACH_ENERGY, self, {
-                if(!self.specialConditions && ef.reason == PLAY_FROM_HAND && ef.card instanceof BasicEnergyCard && ef.card.basicType == D)
+                if(!self.specialConditions && ef.reason == PLAY_FROM_HAND && ef.card.asEnergyCard().containsType(D))
                   if(self.owner.opposite.pbg.bench){
-                    if(confirm("Switch the defending pokemon with 1 of your opponent’s Benched Pokémon? (Your opponent chooses the Benched Pokémon to switch)")) sw self.owner.opposite.pbg.active, self.owner.opposite.pbg.bench.oppSelect("Select the new active Pokémon.")
+                    if(confirm("Make your opponent switch the Defending Pokémon with 1 of your opponent’s Benched Pokémon? (Your opponent chooses the Benched Pokémon to switch)")) sw self.owner.opposite.pbg.active, self.owner.opposite.pbg.bench.select("Select the new active Pokémon.", self.owner.opposite)
                   }
               }
             }
--- a/src/tcgwars/logic/impl/gen3/Deoxys.groovy
+++ b/src/tcgwars/logic/impl/gen3/Deoxys.groovy
@@ -1684,7 +1684,7 @@ public enum Deoxys implements LogicCardInfo {
               before APPLY_ATTACK_DAMAGES, {
                 bg.dm().each{
                   if(self.active && it.from.active && it.to.active) {
-                    bc "Core Guard -10"
+                    bc "Vigorous Aura +10"
                     it.dmg += hp(10)
                   }
                 }
--- a/src/tcgwars/logic/impl/gen3/Deoxys.groovy
+++ b/src/tcgwars/logic/impl/gen3/Deoxys.groovy
@@ -1658,8 +1658,8 @@ public enum Deoxys implements LogicCardInfo {
               def pcs = opp.all.select("Select the Pokémon to attack.")
               damage 20, pcs
               afterDamage{
-                if(pcs.cards.energyCount(C)){
-                  flip {pcs.cards.filterByType(ENERGY).oppSelect("Select the Energy to discard.").discard()}
+                if (pcs.cards.energyCount(C)) {
+                  flip { targeted (pcs, ATTACK) { pcs.cards.filterByType(ENERGY).oppSelect("Select the Energy to discard.").discard() } }
                 }
               }
             }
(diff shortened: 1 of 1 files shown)


src/tcgwars/logic/impl/gen3/Deoxys.groovy
```diff
@@ -297,11 +315,7 @@ public enum Deoxys implements LogicCardInfo {
             energyCost R, C, C
             attackRequirement {}
             onAttack {
-              multiSelect(opp.all, 2).each{
-                targeted(it){
-                  damage 30, it
-                }
-              }
+              multiSelect(opp.all, 2).each{ damage 30, it }
```
flagrama, line 318: I think this can even be simplified down to this:
```suggestion
              damage 30, multiSelect(opp.all, 2)
```
due to the damage override that takes a list of PCS

src/tcgwars/logic/impl/gen3/Deoxys.groovy
```diff
@@ -547,8 +567,10 @@ public enum Deoxys implements LogicCardInfo {
             attackRequirement {}
             onAttack {
               damage 30
-              if(opp.hand.size() >= 5){
-                opp.hand.oppSelect(max : opp.hand.size() - 4,min : opp.hand.size() - 4,"Select the card to discard").discard()
+              afterDamage {
+                while (opp.hand.size() >= 5){
```
flagrama, line 571: I think changing this to a while loop misrepresents what it's doing. There's no way to actually make this loop as with count you _have_ to choose that number of cards, do you not?

src/tcgwars/logic/impl/gen3/Deoxys.groovy
starg09: Do you have any examples of it being used that way already? Looks the same way to me and I'm up for trying (looks cleaner as well), but I'd rather test it somewhere else as a single PR, and then do a mass correction across sets :P

src/tcgwars/logic/impl/gen3/Deoxys.groovy
starg09: Oh, actually thought of replacing it with a single card discard each time... And then thought it didn't add anything meaningful to anyone so I undid it, but left the while by mistake lmao

Will correct in a bit, good catch!

src/tcgwars/logic/impl/gen3/Deoxys.groovy
flagrama: Quite a few it seems. https://gist.github.com/flagrama/1bfe737d82f9a1a9deedd9eb06a7c2d4 though none that explicitly use multiSelect.

flagrama: I don't think it's currently being used, no.

src/tcgwars/logic/impl/gen3/Deoxys.groovy
```diff
@@ -931,15 +948,7 @@ public enum Deoxys implements LogicCardInfo {
...
-              delayed {
-                before PLAY_TRAINER, {
-                  if (ef.cardToPlay.cardTypes.is(SUPPORTER) && bg.currentTurn == self.owner.opposite) {
-                    wcu "Limitation prevents playing this card"
-                    prevent()
-                  }
-                }
-                unregisterAfter 2
-              }
+              opponentCantPlaySupporterNextTurn(delegate)
```
flagrama, line 951: I've got concerns about this method actually.
1. It doesn't check if it was played from hand, but I'm not sure that is an issue.
1. It explicitly says "**Bother** prevents playing supporters" 
1. Looking at it, it unregisters if this Pokémon (Sableye here) Evolves or Switches. I'm not sure evolve even matters considering I have no knowledge of the opponent being able to evolve your Pokémon, but is it actually supposed to unregister on Sableye switching?

src/tcgwars/logic/impl/gen3/Deoxys.groovy
```diff
@@ -1259,10 +1270,7 @@ public enum Deoxys implements LogicCardInfo {
             attackRequirement {}
             onAttack {
               damage 10
-              if(opp.hand){
-                opp.hand.select(hidden:true,"Select one card").showToMe("Selected card.").moveTo(opp.deck)
-                shuffleDeck(null,TargetPlayer.OPPONENT)
-              }
+              afterDamage { astonish() }
```
flagrama, line 1273: Just leaving this as a note of sorts. `astonish()` explicitly mentions **Astonish**

src/tcgwars/logic/impl/gen3/Deoxys.groovy
```diff
@@ -1283,12 +1291,7 @@ public enum Deoxys implements LogicCardInfo {
             text "Look at the top 5 cards of either player’s deck and put them back on top of that player’s deck in any order."
             energyCost C
             attackRequirement {}
```
flagrama, line 1293: Probably a good idea to remove this too, as `rearrangeEitherPlayersDeck` explicitly defines one.

src/tcgwars/logic/impl/gen3/Deoxys.groovy
```diff
@@ -1343,6 +1346,7 @@ public enum Deoxys implements LogicCardInfo {
...
             onAttack {
               damage 40
+              //TODO: Could this choice be done regardless of having [L] Energy attached or not? May want the self damage for some reason.
```
flagrama, line 1349: I believe you need to meet the requirement of an "if you do" effect before you can use it. In this case, having an [L] Energy attached.

starg09:
> More review now that I'm off work. And you keep adding more changes, so no idea if all of this is still relevant

Yeah, definitely relevant. Save for a couple little details I'm going through this one linearly, so anything I didn't catch then is gonna need changes 😛

src/tcgwars/logic/impl/gen3/Deoxys.groovy
```diff
@@ -547,8 +567,8 @@ public enum Deoxys implements LogicCardInfo {
             attackRequirement {}
             onAttack {
               damage 30
-              if(opp.hand.size() >= 5){
-                opp.hand.oppSelect(max : opp.hand.size() - 4,min : opp.hand.size() - 4,"Select the card to discard").discard()
+              afterDamage {
+                opp.hand.oppSelect(count: opp.hand.size() - 4, "Select cards to discard until you end up with ${opp.hand.size() - 4} left").discard()
```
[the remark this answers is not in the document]
starg09: You're correct, edited 👍

src/tcgwars/logic/impl/gen3/Deoxys.groovy
starg09: > It doesn't check if it was played from hand, but I'm not sure that is an issue.

Might need to be tested against Impersonate, Download or Telepass. But those shouldn't be detected as supporters in the first place.

> It explicitly says "Bother prevents playing supporters"

Yeah, that needs a fix so other attacks can use it (same with GRI's Sableye). But probably not the PR to put it in. Will put a TODO as a reminder.

> Looking at it, it unregisters if this Pokémon (Sableye here) Evolves or Switches. I'm not sure evolve even matters considering I have no knowledge of the opponent being able to evolve your Pokémon, but is it actually supposed to unregister on Sableye switching?

Again, agree but should be fixed on a specific statics edit. It's an effect on the player so should remain even if Sableye is switched/evolved(?)/KO'd.

src/tcgwars/logic/impl/gen3/Deoxys.groovy
starg09: More fixes for statics! Hadn't noticed this one though, thanks for pointing it out (there's multiple non-astonish attacks using it already across sets)

src/tcgwars/logic/impl/gen3/Deoxys.groovy
starg09: Yeah, it's a weird wording but probably intended the way you're saying. Can't find a single question about this either, so might just leave it like this.

src/tcgwars/logic/impl/gen3/Deoxys.groovy
starg09: Same, should be corrected in the static.

flagrama: What's to be corrected in the static? This is purely so someone doesn't attempt to add the attackRequirements here before reading what this method actually does, since it already handles it.

starg09:
Ok, that's all cards.

There are 3 attacks (Sableye 23, Lotad 35 and Lunatone 36) that now use static methods, but those methods (even with generic names, save for astonish) print a specific attack in the bc, instead of `$thisMove`. Will make a separate PR later on for fixing those, but leaving it as a note.

src/tcgwars/logic/impl/gen3/Deoxys.groovy
flagrama: Ah, based off your latest commit, I see what you meant, but I think you misunderstood what I meant. I chose the line this comment is on very specifically :laughing:

starg09: Ahh, ok yeah I get what you mean. Will fix it in a bit!

src/tcgwars/logic/impl/gen3/Deoxys.groovy
```diff
@@ -1806,7 +1806,7 @@ public enum Deoxys implements LogicCardInfo {
             attackRequirement {}
             onAttack {
               damage 30
-              noWrDamage(10, self)
+              damage 10, self //TODO: Check this not doing W/R
```
flagrama, line 1809: Where does your thought that this doesn't need to explicitly ignore weakness and resistance come from?

src/tcgwars/logic/impl/gen3/Deoxys.groovy
```diff
@@ -2889,7 +2889,7 @@ public enum Deoxys implements LogicCardInfo {
             onAttack {
               damage 80
               delayed {
-                before APPLY_ATTACK_DAMAGES, {
+                after PROCESS_ATTACK_EFFECTS, {
```
flagrama, line 2894: Interesting. Remembering to verify the order of effects in EffectType.java seems like a good idea when doing this stuff I guess.

src/tcgwars/logic/impl/gen3/Deoxys.groovy
```diff
@@ -2939,21 +2939,20 @@ public enum Deoxys implements LogicCardInfo {
...
-                    if(!src) break;
-                    def card=src.cards.select("Card to move",cardTypeFilter(ENERGY)).first()
-                    energySwitch(src, self, card)
-                  }
+            onActivate {r->
+              if (r==PLAY_FROM_HAND && confirm("Use Dragon Boost to move any number of basic Energy cards attached to your Pokémon to Rayquaza ex?")) {
+                powerUsed()
+                while(1){
+                  def pl=(my.all.findAll {it.cards.filterByType(BASIC_ENERGY) && it != self})
+                  if(!pl) break;
```
flagrama, line 2953: Since this isn't obviously a list by name, my preference would be to specifically check for `pl.empty()` instead of checking for the falsy value.

src/tcgwars/logic/impl/gen3/Deoxys.groovy
```diff
@@ -3133,9 +3133,9 @@ public enum Deoxys implements LogicCardInfo {
...
             delayedA {
               after ATTACH_ENERGY, self, {
-                if(!self.specialConditions && ef.reason == PLAY_FROM_HAND && ef.card instanceof BasicEnergyCard && ef.card.basicType == D)
+                if(!self.specialConditions && ef.reason == PLAY_FROM_HAND && ef.card.asEnergyCard().containsType(D))
                   if(self.owner.opposite.pbg.bench){
-                    if(confirm("Switch the defending pokemon with 1 of your opponent’s Benched Pokémon? (Your opponent chooses the Benched Pokémon to switch)")) sw self.owner.opposite.pbg.active, self.owner.opposite.pbg.bench.oppSelect("Select the new active Pokémon.")
+                    if(confirm("Make your opponent switch the Defending Pokémon with 1 of your opponent’s Benched Pokémon? (Your opponent chooses the Benched Pokémon to switch)")) sw self.owner.opposite.pbg.active, self.owner.opposite.pbg.bench.select("Select the new active Pokémon.", self.owner.opposite)
```
flagrama, line 3140: This line is way too long to abuse the fact that an if statement doesn't require braces IMO.

src/tcgwars/logic/impl/gen3/Deoxys.groovy
starg09: ...Huh, would've sworn there were some psychic weak to psychic (or [L] weak to [L]) that broke this rule... Apparently not, will revert this (In my defense, last weak-to-self card with an attack like this is from 2015 😝 )

src/tcgwars/logic/impl/gen3/Deoxys.groovy
starg09: > Remembering to verify the order of effects in EffectType.java seems like a good idea when doing this stuff I guess.

...Yes. That's exactly what I did~~, and not just follow what worked correctly in other cards blindly. 👀~~

src/tcgwars/logic/impl/gen3/Deoxys.groovy
starg09: Sounds good 👍

src/tcgwars/logic/impl/gen3/Deoxys.groovy
starg09: There, also not _that_ long anymore 👍

src/tcgwars/logic/impl/gen3/Deoxys.groovy
axpendix: `damage 30, multiSelect(opp.all, 2)` would not work :)

PR #586, opened 2020-06-09
Made some EX-era cards look for that time's definition of "Basic Pokémon" [in play], also a similar thing but for SM-era cards and "Evolution Pokémon"

starg09:
Should correct issues with "treat that card as a Basic Pokémon", namely:
* Holon Fossil (HP 86)
* Anorith (LM 29)
* Strange Cave (LM 77)
* Omanyte (PK 56)

Many cards across the EX-era files were coded using `pcs.basic`, while many others used `!pcs.evolution`. Checking around, the correct definition for these attacks would be "Basic Pokémon [in Play]: Not an evolved card". There's no need to add/remove "BASIC" or "STAGE 1" cardTypes, the bottom intent of that text in the above cards' effect is not considering them as evolved. On the opposite side of the spectrum, Baby cards using "Baby Evolution" make the Basic put on top of them count as an evolved Pokémon, even though being a Basic Pokémon _card_:

> Q. Say I use my Pichu's "Baby Evolution" power to evolve Pichu into Pikachu, which is now an "evolved Basic Pokémon". Can I then play [Rare Candy](https://pkmncards.com/card/rare-candy-emerald-em-83/) on my Pikachu to evolve it into Raichu or Raichu-EX during the same turn?
>
> A. Since Pikachu would be considered an "Evolved Pokémon" at that point, Rare Candy cannot be used on that Pikachu. Rare Candy can only be used on "non-Evolved" Pokémon. (Apr 14, 2005 PUI Rules Team)

DP-era onwards, the term "Basic Pokémon" is now used for actual "BASIC" cards on top, so those should indeed work with `pcs.basic` as they already do:

> Q. If I have Uxie, Azelf, and Mesprit with the "Downer Material" Poke-BODY in play, and my opponent has a Snorlax evolved from a Munchlax in play, do they have to pay more for the evolved Basic Pokemon's attacks?
> 
> A. Yes, they do. While Snorlax is evolved it is still a Basic Pokemon card, and that is what Downer Material is looking at. (May 27, 2010 PUI Rules Team)

Likewise, checks for what in EX-era were called "Basic Pokémon [in Play]" now use the phrase "that isn't an evolved Pokémon". See: Mewtwo LV.X's Psybarrier.

As a reminder to myself, last commit includes changes for the EX Deoxys set in particular. There's another PR (#583) with changes for that set, so assuming both are good to merge it might be needed to rebase whichever is accepted last :P 

---

Also, just pushed updates for SM-era sets regarding "Evolution Pokémon". This term was added around the BW/XY era, with cards like craddily placing Stage 1 and 2 Pokémon in play. More notorious example being Breakpoint's Frogadier with "Shadow Duplicates":

> Glaceon-EX's "Crystal Ray" attack makes it immune to damage done by attacks from "Evolution Pokemon". This is a new term that means any Pokemon that is an evolution card. So a Stage 1 Frogadier brought to the Bench by "Water Duplicates" would be affected by this. Basic Pokemon and Restored Pokemon are not affected. (Fates Collide FAQ; May 5, 2016 TPCi Rules Team)

The cards updated all refer to "Evolution Pokémon", and so they should use the method `pcs.realEvolution`, instead of the current `pcs.evolution`. Cards from this era looking for "Basic Pokémon" all look for `pcs.basic` already, so no need for changes there.

### DIFF
--- a/src/tcgwars/logic/impl/gen3/CrystalGuardians.groovy
+++ b/src/tcgwars/logic/impl/gen3/CrystalGuardians.groovy
@@ -224,7 +224,7 @@ public enum CrystalGuardians implements LogicCardInfo {
           attackRequirement {}
           onAttack {
             damage 30
-            if (defending.basic) {
+            if (!defending.evolution) {
               apply CONFUSED
             }
           }
@@ -2028,17 +2028,18 @@ public enum CrystalGuardians implements LogicCardInfo {
           onActivate {
             eff1 = delayed{
               before CHECK_ATTACK_REQUIREMENTS, {
-                if(self.active && ef.attacker.basic) {
+                if(self.active && !ef.attacker.evolution) {
                   wcu "Intimidating Armor prevents attack"
                   prevent()
                 }
               }
             }
             eff2 = getter IS_ABILITY_BLOCKED, { Holder h->
-              if (self.active && h.effect.target.basic && h.effect.target.owner == self.owner.opposite && (h.effect.ability instanceof PokePower || h.effect.ability instanceof PokeBody)) {
+              if (self.active && !h.effect.target.evolution && h.effect.target.owner == self.owner.opposite && (h.effect.ability instanceof PokePower || h.effect.ability instanceof PokeBody)) {
                 h.object=true
               }
             }
+            // TODO: Is eff3 needed here?
             eff3 = getter IS_GLOBAL_ABILITY_BLOCKED, {Holder h->
               if (self.active && (h.effect.target as Card).cardTypes.is(BASIC) && h.effect.target.owner == self.owner.opposite) {
                 h.object=true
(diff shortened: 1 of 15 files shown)


flagrama:
I think I see. "Unevolved" didn't exist as a term in the EX era and earlier, so unevolved Pokémon were to be treated as "Basic" Pokémon, but the modern term used since the LvX era is "unevolved" instead, which frees Basic from having dual meaning? (Based off of searching "unevolved" in the different compendiums)

starg09:
> but the modern term used since the LvX era is "unevolved" instead, which frees Basic from having dual meaning?

Ahh, felt I was missing something. Didn't put any modern cards using "unevolved" as example, haha. But yeah, that'd be the key difference :)

PR #595, opened 2020-06-09
Cosmic Eclipse Fixes Fixes

flagrama:
This PR is to fix any issues I find while testing my CEC fixes. Comments on whether something will work as expected or not are appreciated.


flagrama:
This should fix the things I broke, except for Sylveon. Want to handle that one separately.

PR #602, opened 2020-06-10
New special energy fix

ufodynasty:
The new gen of special energys shouldn't provide energy unless they are attached to the correct type of pokemon. I wasn't sure before but powerful C energy shows why this is important. Leaving as a draft because merge conflicts and because im not sure if the engine will like energy with empty lists.

### DIFF
--- a/src/tcgwars/logic/impl/gen8/DarknessAblaze.groovy
+++ b/src/tcgwars/logic/impl/gen8/DarknessAblaze.groovy
@@ -3356,7 +3356,7 @@ public enum DarknessAblaze implements LogicCardInfo {
         }
       };
       case HEAT_FIRE_ENERGY_184:
-      return specialEnergy (this, [[C]]) {
+      return specialEnergy (this, [[]]) {
         text "This card provides 1 [R] Energy while it’s attached to a Pokémon. If this card is attached to a [R] Pokémon, its maximum HP is increased by 20."
         onPlay {reason->
         }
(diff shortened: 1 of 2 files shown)


starg09:
Oh, had sent a PR implementing the new energies. Just added a commit, will initially push them with empty lists and see how the engine reacts!

flagrama:
Where is it specified that it has to be attached to the correct type of Pokémon? The card itself says it always provides energy so long as it is attached to any Pokémon, but only the proper type of Pokémon get the bonus effects.

starg09:
> Where is it specified that it has to be attached to the correct type of Pokémon?

It can be attached to any Pokémon for the recent ones, and give their type of energy when attached. But it is true that on deck, discard and similar not-in-play areas it shouldn't count as a [C] energy (like how XY energies didn't; as an example, Burning [R] Energy not being able to be brought back by blaziken's Firestarter).

Keeping them as [C] potentially wouldn't cause any issue, but it'd be more correct if they didn't provide any energy at all, until they're attached.

flagrama:
The OP here indicates this removes the ability to give a type when attached to the wrong type of Pokémon though. If it works as you say it should, then it shouldn't be an issue, although the colorless one will need to have a provided energy override as well I believe.

starg09:
> The OP here indicates this removes the ability to give a type when attached to the wrong type of Pokémon though.

Oh, yeah it does. I agree with you on that, they do provide when attached to a different type Pokémon, just not when not in play.

PR #612, opened 2020-06-10
EX Emerald Fixes

starg09:
Also a Memory Berry (CG) fix I already had lol

### DIFF
--- a/src/tcgwars/logic/impl/gen3/CrystalGuardians.groovy
+++ b/src/tcgwars/logic/impl/gen3/CrystalGuardians.groovy
@@ -1925,8 +1925,8 @@ public enum CrystalGuardians implements LogicCardInfo {
       case MEMORY_BERRY_80:
       return pokemonTool (this) {
         text "The Pokémon this card is attached to can use any attack from its Basic Pokémon or its Stage 1 Evolution card. (You still have to pay for that attack's Energy cost.) If that Pokémon attacks, discard this card at the end of the turn."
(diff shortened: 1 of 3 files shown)


src/tcgwars/logic/impl/gen3/Emerald.groovy
```diff
@@ -459,15 +459,24 @@ public enum Emerald implements LogicCardInfo {
           resistance WATER, MINUS30
           pokeBody "Green Essence", {
             text "As long as Sceptile is in play, each of your Active Pokémon that has [G] Energy attached to it can’t be affected by any Special Conditions."
+            def greenEssence={
```
flagrama, line 462: Why define a closure that only gets used once?

starg09: Honestly, was basing it on Dragon Majesty's Phione, which does the same. Don't see harm in it being written this way, but you do have a point.

src/tcgwars/logic/impl/gen3/CrystalGuardians.groovy
```diff
@@ -1938,8 +1938,11 @@ public enum CrystalGuardians implements LogicCardInfo {
             }
           }
           eff2 = delayed {
+            before ATTACK_MAIN, {
+              flag = (ef.attacker == self)
```
luongthomas, line 1942: I think you also need to define `flag`

PR #616, opened 2020-06-11
Roselia LM fixed

Grayahm:
Fixed Roselia

### DIFF
--- a/src/tcgwars/logic/impl/gen3/LegendMaker.groovy
+++ b/src/tcgwars/logic/impl/gen3/LegendMaker.groovy
@@ -1611,7 +1611,7 @@ public enum LegendMaker implements LogicCardInfo {
           onAttack {
             def pcs = defending
             if (opp.bench) {
-              if (confirm("Switch 1 of your opponent’s Benched Pokémon with the Defending Pokémon?")) {
+              if (confirm("Switch 1 of your opponent’s Benched Pokémon with the Defending Pokémon.")) {
                 pcs = opp.bench.select()
                 sw defending, pcs
               }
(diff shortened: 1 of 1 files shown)


src/tcgwars/logic/impl/gen3/LegendMaker.groovy
```diff
@@ -1611,12 +1611,12 @@ public enum LegendMaker implements LogicCardInfo {
           onAttack {
             def pcs = defending
             if (opp.bench) {
-              if (confirm("Switch 1 of your opponent’s Benched Pokémon with the Defending Pokémon.")) {
-                pcs = opp.bench.oppSelect()
+              if (confirm("Switch 1 of your opponent’s Benched Pokémon with the Defending Pokémon?")) {
```
flagrama, line 1614: The text implies this should not be a choice.

Grayahm: Fixed

flagrama: No, I mean you shouldn't use a `confirm` here. The text should go on the `select` instead, and the confirm should be removed.

Grayahm: Oh there was more wrong with this card than I thought haha. That should do it.

luongthomas:
Changes look good, but I believe you need to rebase your branch

src/tcgwars/logic/impl/gen3/LegendMaker.groovy
```diff
@@ -1610,13 +1610,12 @@ public enum LegendMaker implements LogicCardInfo {
...
           onAttack {
             def pcs = defending
-
-            if (opp.bench && confirm("Switch 1 of your opponent’s Benched Pokémon with the Defending Pokémon.")) {
-              pcs = opp.bench.select("Select a Benched Pokémon to become the new Active Pokémon")
-              sw defending, pcs
+            if (opp.bench) {
+                pcs = opp.bench.select("Switch 1 of your opponent’s Benched Pokémon with the Defending Pokémon.")
+                sw defending, pcs
+              }
```
flagrama, line 1616: I sees it.

PR #620, opened 2020-06-11
Fix build

ufodynasty:


flagrama:
This isn't whirlwind. You're confusing double battle text for effect text :P

PR #635, opened 2020-06-12
Fixed Lunatone LM's Sol Shade

Grayahm:


Grayahm:
How does this look?

`getterA (IS_ABILITY_BLOCKED) { Holder h->
              if (my.all.find{it.name == 'Solrock'} && h.effect.target.types.contains(R) && !h.effect.target.topPokemonCard.cardTypes.is(EX) && h.effect.ability instanceof PokePower) {
                h.object=true
              }
            }`

ufodynasty:
It looks better, however on solrock, `}
          delayedA{
            after PLAY_CARD, {
              new CheckAbilities().run(bg)
            }
          }`
was used as the ability didn't immediately activate when any lunatone was put into play. For future reference, you can commit to the branch you  were working on in order to update a pr.

src/tcgwars/logic/impl/gen3/LegendMaker.groovy
```diff
@@ -861,14 +861,12 @@ public enum LegendMaker implements LogicCardInfo {
         weakness P
         pokeBody "Sol Shade", {
           text "As long as you have Solrock in play, each player's [R] Pokémon (excluding Pokémon-ex) can't use any Poké-Powers."
-          delayedA {
             getterA (IS_ABILITY_BLOCKED) { Holder h->
```
ufodynasty, line 864: Also perhaps fix your indentation?

src/tcgwars/logic/impl/gen3/LegendMaker.groovy
```diff
@@ -861,14 +861,12 @@ public enum LegendMaker implements LogicCardInfo {
...
-              if (my.all.find{it.name == 'Solrock'} && h.effect.target.types.contains(R) && !h.effect.target.EX && h.effect.ability instanceof PokePower) {
+              if (my.all.find{it.name == 'Solrock'} && h.effect.target.types.contains(R) && !h.effect.target.topPokemonCard.cardTypes.is(EX) && h.effect.ability instanceof PokePower) {
```
ufodynasty, line 865: .EX has the same functionality as topPokemonCard.cardTypes.is(EX). I prefer the the former as it keeps line length down but either is acceptable

Grayahm:
Okay I think I fixed all of your suggestions, let me know though.

PR #637, opened 2020-06-12
CELEBI_SM224 Fixes

flagrama:
`[]` is an ArrayList by default. Explicitly create an empty `CardList`.
Pokémon was still in the same position in deck after the selection. Try to explicitly move it to the new `CardList` instead of just adding to the list. May need to add `hidden` key to this.


flagrama:
I just mean for the move to the first list. I think without hidden on that first one, there will be some game log spam. I totally agree with the full list being moved to hand, it should not be hidden.

PR #639, opened 2020-06-12
Holon Scientist fix

ufodynasty:


src/tcgwars/logic/impl/gen3/DeltaSpecies.groovy
```diff
@@ -2887,14 +2887,14 @@ public enum DeltaSpecies implements LogicCardInfo {
...
+          if (opp.hand.size() > my.hand.size()-1) {
             def numToDraw = opp.hand.size() - my.hand.size() + 1
             draw numToDraw
           }
         }
         playRequirement{
-          def hand = my.hand.getExcludedList(thisCard).size() >= 1
-          assert (hand || my.discard.filterByType(SUPPORTER)) : "One other card in hand is required to play this card."
+          assert (my.hand.getExcludedList(thisCard))
+          assert (my.hand.size()-2 < opp.hand.size())
```
flagrama, line 2897: Could you add some descriptive text to these?

src/tcgwars/logic/impl/gen3/DeltaSpecies.groovy
```diff
@@ -2887,14 +2887,14 @@ public enum DeltaSpecies implements LogicCardInfo {
           def toDiscard = my.hand.getExcludedList(thisCard).select(count:1, "Discard a card to discard.")
           toDiscard.discard()
 
-          if (opp.hand.size() > my.hand.size()) {
+          if (opp.hand.size() > my.hand.size()-1) {
```
flagrama, line 2890: Is this `-1` technically this card? If so
`my.hand.getExcludedList(thisCard).size()` might be a little bit more verbose, but has more clarity.

PR #640, opened 2020-06-12
Re-Enable "Acid Sampler" on Victreebel (FRLG 17), and add a TODO

starg09:


src/tcgwars/logic/impl/gen3/FireRedLeafGreen.groovy
```diff
@@ -650,14 +650,15 @@ public enum FireRedLeafGreen implements LogicCardInfo {
...
-//                  directDamage 10, self.owner.opposite.pbg.active
-//                }
-//              }
-//            }
+            // The "stops working" is only for 2v2 format, works fine in singles.
+            // TODO: Handle same time effects like Butterfree (FRLG 2)
+            delayedA {
+              before BEGIN_TURN, {
+                if(self.active){
+                  directDamage 10, self.owner.opposite.pbg.active
```
luongthomas, line 658: Should we also add the Source here too?

starg09: ahh, yeah good catch!

On that topic, I think I read somewhere that Source.POKEMONPOWER didn't work? Is that a thing? Cos a lot of Poke-Power/Bodies eventually need to be marked as that and not SRC_ABILITY :P

PR #643, opened 2020-06-12
EX Unseen Forces Fixes

starg09:


src/tcgwars/logic/impl/gen3/UnseenForces.groovy
```diff
@@ -421,16 +418,18 @@ public enum UnseenForces implements LogicCardInfo {
...
+                before BETWEEN_TURNS, {
+                  if (bg.currentTurn == self.owner.opposite) {
+                    bc "Spiky Shell activates"
+                    directDamage 30, defending
+                  }
                 }
+                unregisterAfter 2
+                after SWITCH, defending, {unregister()}
+                after EVOLVE, defending, {unregister()}
               }
```
axpendix, line 432: `targeted` wrap is not needed for effects that already target internally.  in this case `directDamage 30, defending` will target correctly.

starg09: Edited it based on a similar interaction that was patched recently (Trevenant&Dusknoir-GX). Right now on Dev, a Pokémon preventing all effects of attack will still be hit by the delayed directDamage. the targeted should prevent the delayed effect being placed in the first place.

starg09:
Haven't checked the Unown yet, but based on my experience with those across 8 cards in DP so far, and there being 27 here... Yeah I can check those another day lol

This PR's ready for review 👍

PR #647, opened 2020-06-12
Pow! Hand Extension fix

Grayahm:
Effect of card is no longer blocked by cards that prevents effects of attacks


src/tcgwars/logic/impl/gen3/TeamRocketReturns.groovy
```diff
@@ -2542,17 +2542,17 @@ public enum TeamRocketReturns implements LogicCardInfo {
           }
         };
       case R_ENERGY_95:
-        return specialEnergy (this, [[]]) {
+        return specialEnergy (this, [[C]]) {
```
luongthomas, line 2545: Can I ask the reasoning for adding the default colorless value?

Grayahm: Hmm I didn't make those changes, this branch must be out of date. Let me start a new one.

PR #652, opened 2020-06-12
Energymons

ufodynasty:


flagrama:
Do these show themselves as the correct energy type now too?

PR #665, opened 2020-06-13
ROWL_11 DAB - Fix

flagrama:
Information gathering was a bust. This should work in theory though.


flagrama:
Also moved the globalAbility to Bird Keeper since there are two other Pokémon that have the Sky Circus ability too.

PR #666, opened 2020-06-13
Fixed Dark Electrode's Energy Bomb

Grayahm:
Energy Bomb now moves energy after damage has been done.

### DIFF
--- a/src/tcgwars/logic/impl/gen3/TeamRocketReturns.groovy
+++ b/src/tcgwars/logic/impl/gen3/TeamRocketReturns.groovy
@@ -304,6 +304,7 @@ public enum TeamRocketReturns implements LogicCardInfo {
             attackRequirement {}
             onAttack {
               damage 30
+              afterDamage {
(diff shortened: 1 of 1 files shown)


src/tcgwars/logic/impl/gen3/TeamRocketReturns.groovy
```diff
@@ -304,6 +304,7 @@ public enum TeamRocketReturns implements LogicCardInfo {
             attackRequirement {}
             onAttack {
               damage 30
+              afterDamage {
               if(my.bench){
```
flagrama, line 308: Could you fix the indenting of everything inside of your new closure?

Grayahm: I need an IDE lol

flagrama: It's highly recommended.

PR #689, opened 2020-06-14
Darkest Impulse fix

Grayahm:


flagrama:
Generally `def` is used rather than explicit types in the codebase.

Grayahm:
Got it

PR #701, opened 2020-06-15
Surprise Time Machine fix

Grayahm:
No longer crashes game when player fails search.


luongthomas:
Looks like the conflicts needs to be resolved

PR #709, opened 2020-06-16
Fix attempt at Pow! Hand Extension (TRR) and Alakazam (MT)

starg09:
Unneeded, but maybe one param at the start might fix it?

### DIFF
--- a/src/tcgwars/logic/impl/gen3/TeamRocketReturns.groovy
+++ b/src/tcgwars/logic/impl/gen3/TeamRocketReturns.groovy
@@ -2387,7 +2387,7 @@ public enum TeamRocketReturns implements LogicCardInfo {
               if(choice){
                 sw opp.active, opp.bench.select(), TRAINER_CARD
               }else{
-                moveEnergy(opp.active, opp.bench.select("Select the pokemon getting the Energy"), TRAINER_CARD)
+                moveEnergy(basic: false, opp.active, opp.bench.select("Select the pokemon getting the Energy"), TRAINER_CARD)
               }
             }
           }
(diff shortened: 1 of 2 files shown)


src/tcgwars/logic/impl/gen3/TeamRocketReturns.groovy
```diff
@@ -2387,7 +2387,7 @@ public enum TeamRocketReturns implements LogicCardInfo {
               if(choice){
                 sw opp.active, opp.bench.select(), TRAINER_CARD
               }else{
-                moveEnergy(opp.active, opp.bench.select("Select the pokemon getting the Energy"), TRAINER_CARD)
+                moveEnergy(basic: false, opp.active, opp.bench.select("Select the pokemon getting the Energy"), TRAINER_CARD)
```
flagrama, line 2390: I'd just put `[:]` if you're just trying to ensure the first parameter is the empty param map that it is supposed to default to. It's definitely seeing a PCS instead of an empty map with that error.

starg09: Yeah, Was between that and what I put; went with this purely due to the static handling basic as a boolean instead of in the middle of code, so it looked cleaner in my eyes :P

~~Can always try the empty map once this fails.~~

flagrama: A complete workaround could be checking if `param instanceof PcsList` and rearranging the parameters to the proper values if so in the static as well.

starg09: I agree, though I would put it as a PR in itself. Will work on it later, and include a revert on this for whenever that's merged.

PR #710, opened 2020-06-16
Darkness ablaze [W]

flagrama:


src/tcgwars/logic/impl/gen8/DarknessAblaze.groovy
```diff
@@ -1092,6 +1092,34 @@ public enum DarknessAblaze implements LogicCardInfo {
           attackRequirement {}
           onAttack {
             damage 130
+            afterDamage {
```
flagrama, line 1095: Is this mess at all accurate? I'm trying to emulate what a retreat cost would do, so if you return an energy that gives 3 RAINBOW, it should satisfy the 2 [W] requirement. If this is incorrect I can use what some other cards in the code base do.

src/tcgwars/logic/impl/gen8/DarknessAblaze.groovy
```diff
@@ -1092,6 +1092,34 @@ public enum DarknessAblaze implements LogicCardInfo {
...
+                      if (it.contains W) {
+                        finalCount += 1
+                      }
+                    }
+                  }
+                  else {
+                    finalCount += 1
+                  }
+                  energies.add(energy)
+                  if (finalCount >= count) return
```
starg09, line 1116: Asking from ignorance: is return a way to break this loop? In case this doesn't build, maybe something along the lines of:

```
def energyRange = 1..count
for (n in energyRange) {
	//whatever
  if (finalCount >= count) break
}
```

flagrama: Return exits a closure. Break will break out of the switch statement.

src/tcgwars/logic/impl/gen8/DarknessAblaze.groovy
```diff
@@ -1033,7 +1024,19 @@ public enum DarknessAblaze implements LogicCardInfo {
...
+                  if (it.to == self && bg.currentTurn==self.owner.opposite && it.notNoEffect && it.dmg.value) {
+                    bc "Reflect -30"
+                    it.dmg -= hp 30
+                  }
+                }
+              }
+              unregisterAfter 2
+              after EVOLVE, self, {unregister()}
+              after SWITCH, self, {unregister()}
+            }
```
starg09, line 1039: This could be simplified into

```suggestion
            reduceDamageNextTurn(hp(30), thisMove)
```

src/tcgwars/logic/impl/gen8/DarknessAblaze.groovy
```diff
@@ -1051,6 +1054,16 @@ public enum DarknessAblaze implements LogicCardInfo {
         bwAbility "Shuffle Dance", {
           text "Once during your turn, you may choose 1 of your opponent’s face-down Prize cards and switch it with the top card from their deck. (Both cards remain face-down.)"
           actionA {
+            checkLastTurn()
+            assert opp.deck : "Your opponent's deck is empty"
+            powerUsed()
+
+            def info = "Select opponent's Prize card to switch with the top card from their deck."
```
starg09, line 1061: Just a comment but liking this info variable, might steal the idea on future implementations I make :P

src/tcgwars/logic/impl/gen8/DarknessAblaze.groovy
flagrama: Actually, break might work fine in a normal loop like this. It's other closures that it would break out of the switch statement I think. Guess we'll see how it reacts.

src/tcgwars/logic/impl/gen8/DarknessAblaze.groovy
```diff
@@ -1280,9 +1343,11 @@ public enum DarknessAblaze implements LogicCardInfo {
         move "Deep Sea Swirl", {
           text "Shuffle your hand into your deck, then draw 8 cards."
           energyCost C
+          // TODO: Does this need deck to not be empty?
```
starg09, line 1346: Hmm... On current implementation of cards, I'd say you could add an assert for `my.deck || my.hand` (Not 100% sure it'd be required tho). But you can definitely use the attack if you have hand, since you'd shuffle that into the deck and then draw the 8 (or the whole hand again).

src/tcgwars/logic/impl/gen8/DarknessAblaze.groovy
```diff
@@ -1339,7 +1411,12 @@ public enum DarknessAblaze implements LogicCardInfo {
         weakness L
         bwAbility "Primal Law", {
           text "If this Pokémon is your Active Pokémon, your opponent can’t play any Pokémon from their hand to evolve their Pokémon."
-          actionA {
+          delayedA {
+            before EVOLVE, {
+              if (self.active && bg.currentTurn == self.owner.opposite && ef.evolutionCard.player.pbg.hand.contains(ef.evolutionCard)) {
+                prevent()
+              }
+            }
```
starg09, line 1419: This might need a check on it only covering "played from hand". Looks good besides that.

src/tcgwars/logic/impl/gen8/DarknessAblaze.groovy
```diff
@@ -1360,6 +1437,19 @@ public enum DarknessAblaze implements LogicCardInfo {
...
+                  if (it.to == self && bg.currentTurn==self.owner.opposite && it.notNoEffect && it.dmg.value) {
+                    bc "Reflect -60"
+                    it.dmg -= hp 60
+                  }
+                }
+              }
+              unregisterAfter 2
+              after EVOLVE, self, {unregister()}
+              after SWITCH, self, {unregister()}
+            }
```
starg09, line 1452: Same suggestion as earlier.

```suggestion
            reduceDamageNextTurn(hp(60), thisMove)
```

src/tcgwars/logic/impl/gen8/DarknessAblaze.groovy
flagrama: That's what that last condition is. Relevant rulings if you're talking about something like Rare Candy

> Ancient Powerげんしのちから
Each player can't play any Pokémon from his or her hand to evolve his or her Pokémon.

> Q. Does Archeops' "Ancient Power" Ability stop attacks that evolve Pokemon, such as Karrablast's "Mysterious Evolution"?
A. Ancient Power only prevents you from playing Pokemon from your hand to evolve other Pokemon. Mysterious Evolution's effect is not stopped because it evolves the Pokemon from a card in the deck, not a card in your hand. (Noble Victories FAQ; Nov 17, 2011 PUI Rules Team)

> Q. Does Rare Candy override evolution-preventing effects such as Archeops' "Ancient Power" Ability?
A. Ancient Power does not allow you to use a Pokemon's attack or play a Trainer card to evolve your Benched or Active Pokemon, unless the Evolution card you would be putting into play comes from somewhere other than your hand. So Rare Candy cannot be used while Archeops' Ancient Power Ability is in effect. (Noble Victories FAQ; Nov 17, 2011 PUI Rules Team; Nov 21, 2011 PUI Rules Team)

starg09: Ah, missed the last condition somehow (and I even reread 2 times just to make sure 😅 )

src/tcgwars/logic/impl/gen8/DarknessAblaze.groovy
flagrama: Right. I figured Deck counts as non-public information, so shuffling into it, even if the end result is drawing all of the cards probably still counts as an effect happening. I will add that assertion though since in the case both are 0 there definitely isn't an actual effect.

src/tcgwars/logic/impl/gen8/DarknessAblaze.groovy
flagrama: Anything to keep lines from reaching the 100 char marker and wrapping in my VSCode :laughing:

PR #714, opened 2020-06-16
Testing with dump() on Alakazam (MT)

starg09:
See if this actually works for detailed debugging, just the object doesn't say much...


flagrama:
The eternal struggle of "I really want just a little bit more detail than I'm given" with this codebase.

luongthomas:
Could you provide what info dump() provides after you test? I’m curious

starg09:
> Could you provide what info dump() provides after you test? I’m curious

initially looked like nothing, but @flagrama noticed the info was being printed, but mistaken for (broken) html tags:

```
<tcgwars.logic.groovy.tcgbuilders$abilitybuilder$4@205b0b69 prefix="groovy.lang.Reference@2e9cf112" this$0="tcgwars.logic.groovy.TcgBuilders$AbilityBuilder@2aa36019"></tcgwars.logic.groovy.tcgbuilders$abilitybuilder$4@205b0b69>
```

PR #719, opened 2020-06-16
Fix Umbreon ex, remove ef debug from Alakazam MT

starg09:
Breaks the engine and doesn't permit the other testing to be done. Returns null anyways so rip this.

### DIFF
--- a/src/tcgwars/logic/impl/gen3/UnseenForces.groovy
+++ b/src/tcgwars/logic/impl/gen3/UnseenForces.groovy
@@ -2829,8 +2829,8 @@ public enum UnseenForces implements LogicCardInfo {
                     if (h.effect.target == defending && h.effect.ability instanceof PokePower) {
                       h.object = true
                     }
-                    new CheckAbilities().run(bg)
                   }
+                  new CheckAbilities().run(bg)
                 }
                 unregister{
                   eff.unregister()
--- a/src/tcgwars/logic/impl/gen3/UnseenForces.groovy
+++ b/src/tcgwars/logic/impl/gen3/UnseenForces.groovy
@@ -2823,10 +2823,18 @@ public enum UnseenForces implements LogicCardInfo {
             cantRetreat(defending)
             targeted (defending) {
               delayed {
-                getter (IS_ABILITY_BLOCKED) { Holder h->
-                  if (h.effect.target == defending && h.effect.ability instanceof PokePower) {
-                    h.object = true
+                def eff
+                register {
+                  eff = getter (IS_ABILITY_BLOCKED) { Holder h->
+                    if (h.effect.target == defending && h.effect.ability instanceof PokePower) {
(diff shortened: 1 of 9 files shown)


src/tcgwars/logic/impl/gen3/UnseenForces.groovy
```diff
@@ -795,18 +795,27 @@ public enum UnseenForces implements LogicCardInfo {
...
+                  wcu "Intimidating Ring prevents attack"
+                  prevent()
+                }
               }
             }
-          }
-          getterA (IS_ABILITY_BLOCKED) { Holder h->
-            if (self.active && h.effect.target.owner != self.owner && !h.effect.target.evolution && h.effect.ability instanceof PokePower) {
-              h.object=true
+            eff2 = getter (IS_ABILITY_BLOCKED) { Holder h->
```
flagrama, line 808: You changed this from a `getterA` to `getter`. Not sure that was intentional.

flagrama: Actually, think you did this with many pokepowers. Is this the way it's supposed to be done on abilities?

starg09: Yeah, had to register it manually so I can re-check abilities on deactivation of the Poké-Body.

starg09: > Actually, think you did this with many pokepowers. Is this the way it's supposed to be done on abilities?

Luckily, modern abilities don't get deactivated (most of the time) but entirely removed from the game while the ability is on. However, on that topic GALARIAN_WEEZING from RCL is an example of getters being used this way.

src/tcgwars/logic/impl/gen3/UnseenForces.groovy
```diff
@@ -2499,12 +2508,20 @@ public enum UnseenForces implements LogicCardInfo {
...
+                if (h.effect.ability instanceof PokePower || h.effect.ability instanceof PokeBody) {
+                  h.object=true
+                }
               }
             }
+            new CheckAbilities().run(bg)
+          }
+          onDeactivate {
+            eff.unregister()
+            new CheckAbilities().run(bg)
```
axpendix, line 2524: actually, `getterA`,  `onActivate` and `onDeactivate` can all coexist in a single ability definition. @starg09

starg09: Oh? I stand corrected then. Will revert this later, sorry!

axpendix: The order of run is: 
- actionA registered [only a single one allowed]
- delayedA runs [only a single one allowed]
- getterA's run [multiples allowed]
- onActivate called

PR #722, opened 2020-06-16
Darkness Ablaze [W] fixes

flagrama:


flagrama:
The Dracovish "fix" is more of a test.

PR #730, opened 2020-06-17
Reduce Damage refactor

ufodynasty:
reduceDamage isn't being ignored by DamageManager.DamageFlag.NO_DEFENDING_EFFECT and I think this will fix that.


src/tcgwars/logic/groovy/TcgStatics.groovy
```diff
@@ -412,9 +412,27 @@ class TcgStatics {
...
+  //Leaving this in since legacy code probably depends on it.
   static reduceDamageNextTurn (HP reduce, Move thisMove){
     new ReduceDamageNextTurn(reduce, thisMove.name).run(bg())
   }
+  static reduceDamageNextTurn (HP reduce, Move thisMove, PokemonCardSet self){
+    afterDamage {
+      delayed {
+        before APPLY_ATTACK_DAMAGES, {
+          bg.dm().each {
+            if(it.to == self && it.dmg.value){
```
starg09, line 424: Wouldn't `&& it.notNoEffect` be needed over here? Shred/Swift attacks should not decrease the damage inflicted when attacking.

ufodynasty: You are correct but for some reason in my testing switf/shred damage went through just fine without it.

starg09: Huh, weird.

axpendix:
This is a very good refactor, thank you. What is missing?

src/tcgwars/logic/groovy/TcgStatics.groovy
```diff
@@ -412,9 +412,27 @@ class TcgStatics {
...
+          bg.dm().each {
+            if(it.to == self && it.dmg.value && it.notNoEffect){
+              bc "${thisMove.name} reduces damage"
+              it.dmg-=reduce
+            }
+          }
+        }
+        unregisterAfter 2
+        after SWITCH, self, {unregister()}
+        after EVOLVE, self, {unregister()}
```
axpendix, line 432: can we add `DEVOLVE` here as well?

src/tcgwars/logic/groovy/TcgStatics.groovy
```diff
@@ -412,9 +412,27 @@ class TcgStatics {
   static healAll(PokemonCardSet target, Source source=Source.ATTACK){
     bg().em().run(new RemoveDamageCounter(target,source,target.damage))
   }
+  //Leaving this in since legacy code probably depends on it.
   static reduceDamageNextTurn (HP reduce, Move thisMove){
```
axpendix, line 416: I think it's better to change the method name IMO to `reduceDamageToSelfNextTurn` ?

src/tcgwars/logic/impl/gen1/WizardsBlackStarPromosNG.groovy
```diff
@@ -152,8 +152,7 @@ public enum WizardsBlackStarPromosNG implements LogicCardInfo {
           energyCost C
           attackRequirement {}
           onAttack {
-            reduceDamageNextTurn(hp(10), thisMove)
-            // I don't understand the benching condition
```
axpendix, line 156: The Benching condition here is different than other effects;

- If there is `benching either pokemon ends this effect` is present, then the delayed effect is cleared for either switch. 
- If not (default case) then the effect only ends when self was switched. 

I think this should also be added to the method definition. 

`if(either)` -> add `after SWITCH, defending, {unregister()}` as well

src/tcgwars/logic/groovy/TcgStatics.groovy
ufodynasty: The goal of this PR was to change ReduceDamageNextTurn so that it would be ignored by no effect damage. Ideally I would have removed this method entirely but I assumed some legacy code would depend on it. What would be the benefit of renaming this over deleting it?

ufodynasty:
This was fixed in the engine so no refactor :D

PR #736, opened 2020-06-17
Add Implementation for POP Sets 1-5

luongthomas:
This PR adds the implementations for POP series 1-5, which is needed for the 2006 and 2007 formats.

### DIFF
--- a/src/tcgwars/logic/impl/gen3/PopSeries1.groovy
+++ b/src/tcgwars/logic/impl/gen3/PopSeries1.groovy
@@ -213,8 +213,9 @@ public enum PopSeries1 implements LogicCardInfo {
           onAttack {
             damage 50
             if (opp.bench) {
+              def pcs = opp.bench.select("Which Pokémon to deal 20 damage to if flipping heads?")
               flip {
-                damage 20, opp.bench.select("Which Pokemon to deal 20 damage to?")
+                damage 20, pcs
               }
             }
           }
@@ -273,13 +274,15 @@ public enum PopSeries1 implements LogicCardInfo {
         pokeBody "Insomnia", {
           text "Murkrow can't be Asleep."
           delayedA {
-            after APPLY_SPECIAL_CONDITION, {
-              if(self.active){
-                if(self.isSPC(ASLEEP)){
+            before APPLY_SPECIAL_CONDITION, self, {
+                if (ef.type == ASLEEP) {
                   bc "Insomnia prevents $self from being Asleep."
-                  clearSpecialCondition(self, SRC_ABILITY, [ASLEEP])
+                  prevent()
                 }
               }
+            }
+            onActivate {
+              clearSpecialCondition(self, SRC_ABILITY, [ASLEEP])
             }
           }
         }
--- a/src/tcgwars/logic/impl/gen3/PopSeries1.groovy
+++ b/src/tcgwars/logic/impl/gen3/PopSeries1.groovy
@@ -1,4 +1,4 @@
-package tcgwars.logic.impl.gen;
+package tcgwars.logic.impl.gen3;
 
 import static tcgwars.logic.card.HP.*;
 import static tcgwars.logic.card.Type.*;
@@ -32,7 +32,7 @@ import tcgwars.logic.effect.special.*;
 import tcgwars.logic.util.*;
 
 /**
- * @author axpendix@hotmail.com
+ * @author lithogenn@gmail.com
  */
 public enum PopSeries1 implements LogicCardInfo {
 
@@ -110,7 +110,7 @@ public enum PopSeries1 implements LogicCardInfo {
       return evolution (this, from:"Combusken", hp:HP110, type:R, retreatCost:2) {
         weakness W
         move "Fire Punch", {
-          text "40 damage. "
+          text "40 damage."
           energyCost R, C
           attackRequirement {}
           onAttack {
@@ -122,7 +122,7 @@ public enum PopSeries1 implements LogicCardInfo {
           energyCost C, C, C
           attackRequirement {}
           onAttack {
-            damage 50
+            flip 2, { damage 50 }
           }
         }
       };
@@ -131,7 +131,7 @@ public enum PopSeries1 implements LogicCardInfo {
         weakness R
         resistance G, MINUS30
         move "Metal Claw", {
-          text "30 damage. "
+          text "30 damage."
           energyCost M, C
           attackRequirement {}
           onAttack {
@@ -144,6 +144,7 @@ public enum PopSeries1 implements LogicCardInfo {
           attackRequirement {}
           onAttack {
             damage 50
+            flip { discardDefendingEnergy() }
           }
         }
       };
@@ -155,11 +156,14 @@ public enum PopSeries1 implements LogicCardInfo {
           energyCost C
           attackRequirement {}
           onAttack {
-            damage 10
+            flip {
+              damage 10
+              preventAllEffectsNextTurn()
+            }
           }
         }
         move "Dragon Claw", {
-          text "30 damage. "
+          text "30 damage."
           energyCost R, L
           attackRequirement {}
           onAttack {
@@ -177,6 +181,7 @@ public enum PopSeries1 implements LogicCardInfo {
           attackRequirement {}
           onAttack {
             damage 20
+            removeDamageCounterEqualToDamageDone()
           }
         }
         move "Leaf Blade", {
@@ -185,6 +190,7 @@ public enum PopSeries1 implements LogicCardInfo {
           attackRequirement {}
           onAttack {
             damage 40
+            flip { damage 30 }
           }
         }
       };
@@ -197,6 +203,7 @@ public enum PopSeries1 implements LogicCardInfo {
           attackRequirement {}
           onAttack {
             damage 20
+            flip { applyAfterDamage(PARALYZED) }
           }
         }
         move "Mud Splash", {
@@ -205,6 +212,11 @@ public enum PopSeries1 implements LogicCardInfo {
           attackRequirement {}
           onAttack {
             damage 50
+            if (opp.bench) {
+              flip {
+                damage 20, opp.bench.select("Which Pokemon to deal 20 damage to?")
+              }
+            }
           }
         }
       };
@@ -217,6 +229,7 @@ public enum PopSeries1 implements LogicCardInfo {
           attackRequirement {}
           onAttack {
             damage 30
+            heal 10, self
           }
         }
         move "Whirlwind", {
@@ -225,6 +238,9 @@ public enum PopSeries1 implements LogicCardInfo {
           attackRequirement {}
           onAttack {
             damage 40
+            afterDamage {
+              whirlwind()
+            }
           }
         }
       };
@@ -238,10 +254,11 @@ public enum PopSeries1 implements LogicCardInfo {
           attackRequirement {}
           onAttack {
             damage 10
+            flip { damage 20 }
           }
         }
         move "Gust", {
-          text "50 damage. "
+          text "50 damage."
           energyCost G, C, C
           attackRequirement {}
           onAttack {
@@ -256,6 +273,14 @@ public enum PopSeries1 implements LogicCardInfo {
         pokeBody "Insomnia", {
           text "Murkrow can't be Asleep."
           delayedA {
+            after APPLY_SPECIAL_CONDITION, {
+              if(self.active){
+                if(self.isSPC(ASLEEP)){
+                  bc "Insomnia prevents $self from being Asleep."
+                  clearSpecialCondition(self, SRC_ABILITY, [ASLEEP])
+                }
+              }
+            }
           }
         }
         move "Feint Attack", {
@@ -263,7 +288,7 @@ public enum PopSeries1 implements LogicCardInfo {
           energyCost C, C
           attackRequirement {}
           onAttack {
-
+            swiftDamage(20, opp.all.select("Which Pokémon to deal 20 damage to?"))
           }
         }
       };
@@ -276,6 +301,7 @@ public enum PopSeries1 implements LogicCardInfo {
           attackRequirement {}
           onAttack {
             damage 20
+            flip { damage 10 }
           }
         }
       };
@@ -287,7 +313,7 @@ public enum PopSeries1 implements LogicCardInfo {
           energyCost F
           attackRequirement {}
           onAttack {
-
+            amnesia delegate
           }
         }
         move "Body Slam", {
@@ -296,6 +322,7 @@ public enum PopSeries1 implements LogicCardInfo {
           attackRequirement {}
           onAttack {
             damage 20
+            flip { applyAfterDamage(PARALYZED) }
           }
         }
       };
@@ -308,6 +335,7 @@ public enum PopSeries1 implements LogicCardInfo {
           attackRequirement {}
           onAttack {
             damage 10
+            sandAttack(thisMove)
           }
         }
       };
@@ -321,6 +349,7 @@ public enum PopSeries1 implements LogicCardInfo {
           attackRequirement {}
           onAttack {
             damage 10
+            flip { applyAfterDamage(PARALYZED) }
           }
         }
         move "Spark", {
@@ -329,6 +358,13 @@ public enum PopSeries1 implements LogicCardInfo {
           attackRequirement {}
           onAttack {
             damage 10
+            if (opp.bench){
+              multiSelect(opp.bench, 2).each {
+                targeted(it) {
+                  damage 10, it
+                }
+              }
+            }
           }
         }
       };
@@ -342,6 +378,7 @@ public enum PopSeries1 implements LogicCardInfo {
           attackRequirement {}
           onAttack {
             damage 10
+            flip { damage 10 }
           }
         }
         move "Agility", {
@@ -350,6 +387,7 @@ public enum PopSeries1 implements LogicCardInfo {
           attackRequirement {}
           onAttack {
             damage 20
+            flip { preventAllEffectsNextTurn() }
           }
         }
       };
@@ -362,6 +400,7 @@ public enum PopSeries1 implements LogicCardInfo {
           attackRequirement {}
           onAttack {
             damage 10
+            flip { applyAfterDamage(PARALYZED) }
           }
         }
       };
@@ -374,7 +413,7 @@ public enum PopSeries1 implements LogicCardInfo {
           energyCost C
           attackRequirement {}
           onAttack {
-
+            increasedBaseDamageNextTurn("Agility", hp(40))
           }
         }
         move "Agility", {
@@ -383,6 +422,7 @@ public enum PopSeries1 implements LogicCardInfo {
           attackRequirement {}
           onAttack {
             damage 30
+            flip { preventAllEffectsNextTurn() }
           }
         }
       };
@@ -395,7 +435,7 @@ public enum PopSeries1 implements LogicCardInfo {
           energyCost F, C
           attackRequirement {}
           onAttack {
-
+            damage 30
           }
         }
         move "Supersonic Claws", {
@@ -404,6 +444,7 @@ public enum PopSeries1 implements LogicCardInfo {
           attackRequirement {}
           onAttack {
             damage 80
+            dontApplyResistance()
           }
         }
       };
@@ -413,7 +454,7 @@ public enum PopSeries1 implements LogicCardInfo {
         weakness F
         resistance P, MINUS30
         move "Scratch", {
-          text "20 damage. "
+          text "20 damage."
           energyCost C
           attackRequirement {}
           onAttack {
@@ -423,9 +464,12 @@ public enum PopSeries1 implements LogicCardInfo {
         move "Critical Crush", {
           text "80 damage. Discard 2 Basic Energy cards attached to Tyranitar ex or this attack does nothing."
           energyCost F, F, C, C
-          attackRequirement {}
+          attackRequirement {
+            assert self.cards.filterByType(BASIC_ENERGY).size() >= 2 : "$self needs 2 or more Basic Energies attached."
+          }
           onAttack {
             damage 80
+            self.cards.filterByType(BASIC_ENERGY).select(count:2, "Discard 2 Basic Energies to discard.").discard()
           }
         }
       };
--- a/src/tcgwars/logic/impl/gen3/PopSeries1.groovy
+++ b/src/tcgwars/logic/impl/gen3/PopSeries1.groovy
@@ -468,7 +468,7 @@ public enum PopSeries1 implements LogicCardInfo {
           text "80 damage. Discard 2 Basic Energy cards attached to Tyranitar ex or this attack does nothing."
           energyCost F, F, C, C
           attackRequirement {
-            assert self.cards.filterByType(BASIC_ENERGY).size() >= 2 : "$self needs 2 or more Basic Energies attached."
+            assert self.cards.filterByType(BASIC_ENERGY).size() >= 2 : "$self needs 2 or more Basic Energies attached"
           }
           onAttack {
             damage 80
--- a/src/tcgwars/logic/impl/gen3/PopSeries1.groovy
+++ b/src/tcgwars/logic/impl/gen3/PopSeries1.groovy
@@ -469,7 +469,7 @@ public enum PopSeries1 implements LogicCardInfo {
           }
           onAttack {
             damage 80
-            self.cards.filterByType(BASIC_ENERGY).select(count:2, "Discard 2 Basic Energies to discard.").discard()
+            self.cards.filterByType(BASIC_ENERGY).select(count:2, "Select 2 Basic Energies to discard.").discard()
           }
         }
       };
--- a/src/tcgwars/logic/impl/gen3/PopSeries2.groovy
+++ b/src/tcgwars/logic/impl/gen3/PopSeries2.groovy
@@ -1,5 +1,7 @@
 package tcgwars.logic.impl.gen3;
 
+import tcgwars.logic.impl.gen1.BaseSetNG;
+import tcgwars.logic.impl.gen3.FireRedLeafGreen;
 import tcgwars.logic.impl.gen7.CelestialStorm;
 
 import static tcgwars.logic.card.HP.*;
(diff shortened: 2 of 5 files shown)


src/tcgwars/logic/impl/gen3/PopSeries2.groovy
```diff
@@ -245,34 +279,61 @@ public enum PopSeries2 implements LogicCardInfo {
...
       return stadium (this) {
         text "Once during each player's turn, when that player attaches an Energy card from his or her hand to 1 of his or her Benched Pokémon, he or she removes 1 damage counter from that Pokémon."
+        def eff
         onPlay {
+          eff = delayed {
+            after ATTACH_ENERGY, {
+              if (ef.reason == PLAY_FROM_HAND) {
+                heal 10, ef.resolvedTarget
+              }
+            }
```
starg09, line 328:
```suggestion
            after ATTACH_ENERGY, {
              if (ef.reason == PLAY_FROM_HAND && bg.em().retrieveObject("Pokemon_Park_" + thisCard.hashCode()) != bg.turnCount) {
                bg.em().storeObject("Pokemon_Park_" + thisCard.hashCode(), bg.turnCount)
                heal 10, ef.resolvedTarget
              }
            }
```

> Q. Do you have to use Pokémon Park on your first energy attachment for the turn, or can you choose to wait until the second or third attachment (i.e. with Blastoise-EX)?
>
> A. It's not optional - there isn't a "may" clause. That player *must* remove 1 damage counter from that Pokémon, provided it is the first time that turn that he or she is attaching an Energy card from the hand to one of his or her BENCHED Pokémon. (Sep 29, 2005 PUI Rules Team)

src/tcgwars/logic/impl/gen3/PopSeries2.groovy
```diff
@@ -245,34 +279,61 @@ public enum PopSeries2 implements LogicCardInfo {
...
         }
       };
       case POKEMON_PARK_10:
       return stadium (this) {
         text "Once during each player's turn, when that player attaches an Energy card from his or her hand to 1 of his or her Benched Pokémon, he or she removes 1 damage counter from that Pokémon."
+        def eff
         onPlay {
+          eff = delayed {
+            after ATTACH_ENERGY, {
+              if (ef.reason == PLAY_FROM_HAND) {
```
flagrama, line 325: This seems okay, but to throw a wrench in here ...

When the card says "when that player attaches an Energy card from his or her hand" does it mean so long as the card was in the hand before being attached, like Rare Candy does with evolution, and evolution preventing abilities check for to prevent Rare Candy?

If so you probably need a flag that resets at the end of the turn, and check explicitly if the card was in hand before being attached instead of relying on the `reason`.

src/tcgwars/logic/impl/gen3/PopSeries2.groovy
```diff
@@ -362,6 +427,7 @@ public enum PopSeries2 implements LogicCardInfo {
           attackRequirement {}
           onAttack {
             damage 30
+            flip 1, {}, { directDamage 10, self}
```
starg09: This Pikachu is a reprint of the Base Set one

`return copy(BaseSetNG.PIKACHU, this);`

src/tcgwars/logic/impl/gen3/PopSeries2.groovy
```diff
@@ -313,7 +375,7 @@ public enum PopSeries2 implements LogicCardInfo {
           energyCost C
           attackRequirement {}
           onAttack {
-
+            // This also does nothing
```
flagrama, line 378: This needs an attackRequirement. Probably `assert false : "This attack does nothing"`

src/tcgwars/logic/impl/gen3/PopSeries2.groovy
flagrama, line 430: Again, don't think this should be placing a damage counter directly.
```suggestion
            flip 1, {}, { damage 10, self}
```

src/tcgwars/logic/impl/gen3/PopSeries2.groovy
```diff
@@ -370,7 +436,12 @@ public enum PopSeries2 implements LogicCardInfo {
         weakness P
         pokePower "Time Reversal", {
           text "Once during your turn, when you put Celebi ex from your hand onto your Bench, you may search your discard pile for a card, show it to your opponent, and put it on top of your deck."
-          actionA {
+          onActivate {r->
+            if (r==PLAY_FROM_HAND && bg.em().retrieveObject("Time_Reversal")!=bg.turnCount && my.discard.filterByType(TRAINER) && confirm('Use Time Reversal to search your discard pile for a card to put on top of your deck?')) {
```
flagrama, line 440: I don't think it's proper to use a storedObject here. Once a card goes to your hand or deck it is private information and no longer has a record of used effects. This should only be done for "You can only use 1 $name per turn" abilities.

src/tcgwars/logic/impl/gen3/PopSeries2.groovy
```diff
@@ -379,6 +450,7 @@ public enum PopSeries2 implements LogicCardInfo {
           attackRequirement {}
           onAttack {
             damage 30
+            preventAllEffectsFromCustomPokemonNextTurn(thisMove, self, {it.EX})
```
flagrama, line 453: Does this prevent ex Pokémon only from applying effects to Celebi ex? It looks like it should be all Pokémon that are prevented from applying effects to it.

src/tcgwars/logic/impl/gen3/PopSeries2.groovy
```diff
@@ -370,7 +436,12 @@ public enum PopSeries2 implements LogicCardInfo {
         weakness P
         pokePower "Time Reversal", {
           text "Once during your turn, when you put Celebi ex from your hand onto your Bench, you may search your discard pile for a card, show it to your opponent, and put it on top of your deck."
-          actionA {
+          onActivate {r->
+            if (r==PLAY_FROM_HAND && bg.em().retrieveObject("Time_Reversal")!=bg.turnCount && my.discard.filterByType(TRAINER) && confirm('Use Time Reversal to search your discard pile for a card to put on top of your deck?')) {
+              bg.em().storeObject("Time_Reversal", bg.turnCount)
+              powerUsed()
+              my.discard.select("Select a card to put on top of your deck").showToOpponent("Selected card to move to the top of their deck").moveTo(addToTop:true, my.deck)
+            }
```
starg09: Why is this being limited to 1 Time Reversal per turn? Should be reusable with super scoop up, as far as I understand at least. Also, it puts any card, not just trainers.

```suggestion
          onActivate {r->
            if (r==PLAY_FROM_HAND && my.discard && confirm('Use Time Reversal to search your discard pile for a card to put on top of your deck?')) {
              powerUsed()
              my.discard.select("Select a card to put on top of your deck").showToOpponent("Selected card to move to the top of their deck").moveTo(addToTop:true, my.deck)
            }
```

src/tcgwars/logic/impl/gen3/PopSeries1.groovy
```diff
@@ -256,14 +273,22 @@ public enum PopSeries1 implements LogicCardInfo {
         pokeBody "Insomnia", {
           text "Murkrow can't be Asleep."
           delayedA {
+            after APPLY_SPECIAL_CONDITION, {
```
ufodynasty, line 276: This doesn't cover the case where an ability blocker is removed from play while murkrow is asleep. I think something like
`delayedA {
            before APPLY_SPECIAL_CONDITION,self, {
             if(ef.type == ASLEEP){
                prevent()
              }
            }
          }
          onActivate {
            clearSpecialCondition(self, SRC_ABILITY, [ASLEEP])
          }`

src/tcgwars/logic/impl/gen3/PopSeries3.groovy
```diff
@@ -123,6 +128,8 @@ public enum PopSeries3 implements LogicCardInfo {
           attackRequirement {}
           onAttack {
             damage 60
+            directDamage 10, self
```
starg09, line 131:
```suggestion
            damage 10, self
```

src/tcgwars/logic/impl/gen3/PopSeries1.groovy
ufodynasty:
```suggestion
            delayedA {
            before APPLY_SPECIAL_CONDITION,self, {
              if(ef.type == ASLEEP){
                prevent()
              }
            }
          }
          onActivate {
            clearSpecialCondition(self, SRC_ABILITY, [ASLEEP])
          }
```

src/tcgwars/logic/impl/gen3/PopSeries1.groovy
```diff
@@ -256,14 +273,22 @@ public enum PopSeries1 implements LogicCardInfo {
         pokeBody "Insomnia", {
           text "Murkrow can't be Asleep."
           delayedA {
```
ufodynasty, line 275:
```suggestion
          delayedA {
          before APPLY_SPECIAL_CONDITION,self, {
              if(ef.type == ASLEEP){
                prevent()
              }
            }
          }
          onActivate {
            clearSpecialCondition(self, SRC_ABILITY, [ASLEEP])
          }
```

src/tcgwars/logic/impl/gen3/PopSeries3.groovy
```diff
@@ -135,14 +142,18 @@ public enum PopSeries3 implements LogicCardInfo {
           attackRequirement {}
           onAttack {
             damage 10
+            flip { damage 20 }
           }
         }
         move "Fire Spin", {
           text "70 damage. Discard 2 Basic energy cards attached to Flareon or this attack does nothing."
           energyCost F, F, C
           attackRequirement {}
```
starg09, line 151: I'm in favour of this approach, tho some cards do have the energy available added here I guess?

src/tcgwars/logic/impl/gen3/PopSeries1.groovy
ufodynasty: This will cover the case where an ability blocker is removed from play while murkrow is asleep

src/tcgwars/logic/impl/gen3/PopSeries3.groovy
```diff
@@ -135,14 +142,18 @@ public enum PopSeries3 implements LogicCardInfo {
...
           onAttack {
-            damage 70
+            if (self.cards.filterByType(BASIC_ENERGY).size() >= 2) {
```
flagrama, line 153: Same questions as the other two similar attacks.

src/tcgwars/logic/impl/gen3/PopSeries3.groovy
```diff
@@ -176,7 +188,12 @@ public enum PopSeries3 implements LogicCardInfo {
           energyCost C
           attackRequirement {}
           onAttack {
-
+            if (all.findAll.findAll {it.name == "Plusle"}) {
```
flagrama, line 191: Is `findAll.findAll` a mistake, or something I'm not familiar with that you have to do it this way?

src/tcgwars/logic/impl/gen3/PopSeries3.groovy
```diff
@@ -185,6 +202,7 @@ public enum PopSeries3 implements LogicCardInfo {
           attackRequirement {}
           onAttack {
             damage 20
+            reduceDamageNextTurn(hp(10), thisMove)
```
flagrama, line 205: This only applies to your current active, doesn't it? I think you need to use this method instead
```suggestion
            reduceDamageFromDefendingNextTurn(hp(10), thisMove, defending)
```

src/tcgwars/logic/impl/gen3/PopSeries3.groovy
```diff
@@ -193,6 +211,16 @@ public enum PopSeries3 implements LogicCardInfo {
...
+            def validTargets = opp.all.findAll {
+              it.abilities.keySet().find { ability ->
+                ability instanceof PokePower
+              }
+            }
+
+            if (validTargets) {
+              validTargets.select("Choose 1 of your opponent's Pokémon that has any Poké-Powers to deal 30 damage to.")
+            }
           }
```
starg09, line 224: Should apply the damage to the target, not the active.

```suggestion
          attackRequirement {
            assert opp.all.any {
              it.abilities.keySet().find { ability -> ability instanceof PokePower }
            } : "Your opponent has no Pokémon with Poké-Powers"
          }
          onAttack {
            def validTargets = opp.all.findAll {
              it.abilities.keySet().find { ability ->
                ability instanceof PokePower
              }
            }
            damage 30, validTargets.select("Choose 1 of your opponent's Pokémon that has any Poké-Powers to deal 30 damage to it.")
          }
```

src/tcgwars/logic/impl/gen3/PopSeries3.groovy
```diff
@@ -222,14 +256,24 @@ public enum PopSeries3 implements LogicCardInfo {
...
+            def validTargets = opp.all.findAll {
+              it.abilities.keySet().find { ability ->
+                ability instanceof PokeBody
+              }
+            }
+
+            if (validTargets) {
+              validTargets.select("Choose 1 of your opponent's Pokémon that has any Poké-bodies to deal 30 damage to.")
+            }
           }
```
starg09, line 269: Same fix as Minun.

```suggestion
          attackRequirement {
            assert opp.all.any {
              it.abilities.keySet().find { ability -> ability instanceof PokeBody }
            } : "Your opponent has no Pokémon with Poké-Bodies"
          }
          onAttack {
            def validTargets = opp.all.findAll {
              it.abilities.keySet().find { ability ->
                ability instanceof PokeBody
              }
            }
            damage 30, validTargets.select("Choose 1 of your opponent's Pokémon that has any Poké-Bodies to deal 30 damage to it.")
          }
```

src/tcgwars/logic/impl/gen3/PopSeries2.groovy
ufodynasty: Another way to so it would be something like 
```suggestion
              def card
              before ATTACH_ENERGY, {
              if(my.hand.contains(ef.card){
              card = ef.card
              }
              }
              after ATTACH_ENERGY, {
              if(ef.card = card){
              heal 10, ef.resolvedTarget
              }
              }
```

src/tcgwars/logic/impl/gen3/PopSeries3.groovy
```diff
@@ -302,23 +348,41 @@ public enum PopSeries3 implements LogicCardInfo {
...
+            if (h.effect.target.types.contains(G) || h.effect.target.types.contains(L)) {
+              h.object += hp(10)
+            }
+          }
         }
-        playRequirement{
+        onRemoveFromPlay {
+          eff.unregister()
         }
       };
```
starg09, line 387: Reprints:
* High Pressure System (EX Dragon 85)
* Low Pressure System (EX Dragon 86)

src/tcgwars/logic/impl/gen3/PopSeries3.groovy
```diff
@@ -302,23 +348,41 @@ public enum PopSeries3 implements LogicCardInfo {
...
       case DITTO_12:
```
starg09, line 388:
```suggestion
      case DITTO_12:
      // TODO: Implement Duplicate
```

src/tcgwars/logic/impl/gen3/PopSeries3.groovy
```diff
@@ -370,14 +437,15 @@ public enum PopSeries3 implements LogicCardInfo {
           attackRequirement {}
           onAttack {
             damage 50
```
starg09, line 439: Text card is mistranscribed, should do 40 damage instead (text also needs to be updated).

```suggestion
            damage 40
```

image

src/tcgwars/logic/impl/gen3/PopSeries3.groovy
```diff
@@ -401,29 +469,18 @@ public enum PopSeries3 implements LogicCardInfo {
           energyCost L, L
           attackRequirement {}
           onAttack {
-            damage 20
+            flip "Bustle", 2, {}, {},
```
flagrama, line 472: Couldn't this be simplified to just
```suggestion
            flip "Bustle", 2, {damage 20 applyAfterDamage CONFUSED}, {}
```
Or is there an instance before damage completes that getting confused a second time would be an issue? If so, setting a flag and using that to apply the confused status would also work. Using the map just doesn't seem entirely necessary to me.

src/tcgwars/logic/impl/gen3/PopSeries3.groovy
```diff
@@ -370,14 +437,15 @@ public enum PopSeries3 implements LogicCardInfo {
...
+            applyAfterDamage(POISONED)
           }
         }
       };
       case MARSHTOMP_15:
       return evolution (this, from:"Mudkip", hp:HP070, type:F, retreatCost:1) {
         weakness G
         move "Mud Slap", {
-          text "20 damage. "
+          text "20 damage."
```
starg09, line 448: Not this attack, the one below (Double Edge) should do `damage 10, self`

src/tcgwars/logic/impl/gen3/PopSeries4.groovy
```diff
@@ -106,53 +110,41 @@ public enum PopSeries4 implements LogicCardInfo {
...
+      case CHIMECHO_DELTA_1:
       return basic (this, hp:HP060, type:M, retreatCost:1) {
         weakness P
         move "Wrap", {
           text "Flip a coin. If heads, the Defending Pokémon is now Paralyzed."
           energyCost C
           attackRequirement {}
           onAttack {
-
+            flip { applyAfterDamage(PARALYZED) }
```
starg09, line 121: Minor but since no damage
```suggestion
            flip { apply PARALYZED }
```

src/tcgwars/logic/impl/gen3/PopSeries2.groovy
luongthomas: Looks like the text got chopped off when it was generating the file.  I'll add the rest of the card text.  Here is an image of the card for reference. https://images-na.ssl-images-amazon.com/images/I/61mIljo8BCL._AC_.jpg

src/tcgwars/logic/impl/gen3/PopSeries4.groovy
```diff
@@ -192,14 +192,25 @@ public enum PopSeries4 implements LogicCardInfo {
         pokePower "Energy Trans", {
           text "As often as you like during your turn (before your attack), move a [G] Energy card attached to 1 of your Pokémon to another of your Pokémon. This power can't be used if Sceptile is affected by a Special Condition."
           actionA {
+            checkNoSPC()
+            assert my.all.findAll {it.cards.energyCount(G)>0}
+            assert my.all.size()>=2
```
flagrama, line 197: Some warning text for the user would be nice.

src/tcgwars/logic/impl/gen3/PopSeries3.groovy
luongthomas: Yeah you are correct, this could be simplified as per your suggestion.  Thanks!

src/tcgwars/logic/impl/gen3/PopSeries4.groovy
```diff
@@ -192,14 +192,25 @@ public enum PopSeries4 implements LogicCardInfo {
         pokePower "Energy Trans", {
```
starg09, line 192: Reprint from Ruby&Sapphire

`return copy(RubySapphire.SCEPTILE_20, this);`

src/tcgwars/logic/impl/gen3/PopSeries3.groovy
luongthomas: It's a mistake 🤣   Thanks for catching it

starg09:
> Regarding 1 though, @starg09 if you have one energy basic energy card, would you be able to declare the attack, discard the 1 energy, but then fail the attack due to not having a second basic energy card to discard?

@flagrama can't reply to the conversation for some reason. No, if you don't have 2 energies for discarding the attack does nothing (so no discard, and no damage)

starg09:
> Or is there an instance before damage completes that getting confused a second time would be an issue? If so, setting a flag and using that to apply the confused status would also work. Using the map just doesn't seem entirely necessary to me.

@flagrama No actual issue, but the confusion message will appear duplicated so I'd personally use a flag :P

src/tcgwars/logic/impl/gen3/PopSeries4.groovy
```diff
@@ -192,14 +192,25 @@ public enum PopSeries4 implements LogicCardInfo {
...
+            energySwitch(src, tar, card)
           }
         }
         move "Tail Rap", {
           text "50x damage. Flip 2 coins. This attack does 50 damage times the number of heads."
           energyCost G, C, C
           attackRequirement {}
           onAttack {
-            damage 50
+            flip { damage 50 }
```
flagrama, line 213: Flip 2 coins.
```suggestion
            flip 2, { damage 50 }
```

src/tcgwars/logic/impl/gen3/PopSeries4.groovy
```diff
@@ -241,41 +255,16 @@ public enum PopSeries4 implements LogicCardInfo {
...
+      return copy(Deoxys.HEAL_ENERGY_94, this);
       case POKEMON_FAN_CLUB_9:
-      return basicTrainer (this) {
-        text "Search your deck for up to 2 Basic Pokémon and put them onto your Bench. Shuffle your deck afterward."
-        onPlay {
-        }
-        playRequirement{
-        }
-      };
+      return copy(Flashfire.POKEMON_FAN_CLUB_94,this);
```
starg09, line 265: This should copy Aquapolis, or be its own card. The Flashfire version places the basic Pokémon in the hand, this version places them on the bench.

src/tcgwars/logic/impl/gen3/PopSeries4.groovy
```diff
@@ -319,23 +313,31 @@ public enum PopSeries4 implements LogicCardInfo {
         pokeBody "Shell Retreat", {
```
starg09, line 313: This Squirtle is a copy from Team Magma vs Team Aqua 46.

src/tcgwars/logic/impl/gen3/PopSeries4.groovy
```diff
@@ -319,23 +313,31 @@ public enum PopSeries4 implements LogicCardInfo {
...
           energyCost C
           attackRequirement {}
           onAttack {
             damage 10
           }
         }
       };
-      case TREECKO_DELTA_SPECIES__15:
+      case TREECKO_DELTA_15:
       return basic (this, hp:HP040, type:P, retreatCost:1) {
```
starg09, line 336: reprint from Crystal Guardians

`return copy (CrystalGuardians.TREECKO_DELTA_68, this)`

src/tcgwars/logic/impl/gen3/PopSeries5.groovy
```diff
@@ -142,125 +148,76 @@ public enum PopSeries5 implements LogicCardInfo {
...
           }
         }
         move "Extra Draw", {
           text "If your opponent has any Pokémon-ex in play, search your deck for up to 2 basic Energy cards and attach them to Mew. Shuffle your deck afterward."
           energyCost R
-          attackRequirement {}
+          attackRequirement { assert my.deck : "Deck is empty." }
           onAttack {
-
+            if (opp.all.findAll { it.EX }) {
```
flagrama, line 181: I think this should be an attackRequirement. Also forgot to shuffle the deck.

src/tcgwars/logic/impl/gen3/PopSeries5.groovy
```diff
@@ -293,16 +251,27 @@ public enum PopSeries5 implements LogicCardInfo {
           attackRequirement {}
           onAttack {
             damage 10
+            flip  {
+              self.cards.energyCount(C), {damage 20}
```
flagrama, line 255: the count should be outside the closure so that flip executes that many times. Not sure how this will react as written.

src/tcgwars/logic/impl/gen3/PopSeries5.groovy
```diff
@@ -340,6 +313,11 @@ public enum PopSeries5 implements LogicCardInfo {
         pokePower "Purple Ray", {
           text "Once during your turn, when you put Espeon * from your hand onto your Bench, you may use this power. Each Active Pokémon (both yours and your opponent's) is now Confused."
           actionA {
+            if (it==PLAY_FROM_HAND && opp.hand && confirm("Use Purple Ray?")) {
```
flagrama, line 316: Shouldn't this be in an onActivate? I think that's how UF Umbreon does it's PokePower.

Edit: and how the Umbreon here does too it seems. :laughing:

src/tcgwars/logic/impl/gen3/PopSeries5.groovy
```diff
@@ -348,6 +326,7 @@ public enum PopSeries5 implements LogicCardInfo {
           attackRequirement {}
           onAttack {
             damage 30
+            damage 10*defending.cards.energyCount(C)
```
flagrama, line 329: Need to count the cards, not the amount attached.

src/tcgwars/logic/impl/gen3/PopSeries2.groovy
```diff
@@ -141,6 +157,7 @@ public enum PopSeries2 implements LogicCardInfo {
           attackRequirement {}
           onAttack {
             damage 20
+            flip { damage 30 }
           }
         }
       };
```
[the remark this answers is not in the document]
luongthomas: I think it's good to have a version with a groovy implementation whenever we can, that way we can fix the card if we need.  But that's free for debate

src/tcgwars/logic/impl/gen3/PopSeries2.groovy
```diff
@@ -245,34 +279,61 @@ public enum PopSeries2 implements LogicCardInfo {
       return supporter (this) {
```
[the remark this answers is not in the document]
luongthomas: I think it's good to have a version with a groovy implementation here, just in case we need to fix a bug or interaction

axpendix:
amazing stuff

PR #737, opened 2020-06-17
Fix Aipom's id

starg09:
(Build was already fixed apparently)


axpendix:
Was this a new card?

starg09:
> Was this a new card?

Yup, from the recent Pikachu&Zekrom Battle Deck.

PR #742, opened 2020-06-17
Darkness ablaze [L]

flagrama:


src/tcgwars/logic/impl/gen8/DarknessAblaze.groovy
```diff
@@ -1453,6 +1454,7 @@ public enum DarknessAblaze implements LogicCardInfo {
           attackRequirement {}
           onAttack {
             damage 30
+            switchYourActive may:true
```
luongthomas, line 1457: Does this need parenthesis?

src/tcgwars/logic/impl/gen8/DarknessAblaze.groovy
```diff
@@ -1530,9 +1551,11 @@ public enum DarknessAblaze implements LogicCardInfo {
         move "Risk Taker", {
           text "Flip a coin, if heads discard 5 cards from the top of your opponent’s deck. If tails, discard 5 cards from the top of your deck."
           energyCost C, C
-          attackRequirement {}
+          attackRequirement {
+            assert my.deck || opp.deck : "Both decks are empty"
+          }
           onAttack {
-
+            flip {if (opp.deck) opp.deck[0..4].discard()}, {if (my.deck) my.deck[0..4].discard()}
```
luongthomas, line 1558: I think you need to use `flip 1, {...}, {...}`

src/tcgwars/logic/impl/gen8/DarknessAblaze.groovy
flagrama: Since I am providing arguments and it isn't nested in another method, I _should not_ need parentheses. If it causes issues it'll be a quick fix at least.

src/tcgwars/logic/impl/gen8/DarknessAblaze.groovy
flagrama: `static flip (Closure eachHead, Closure eachTail={}, multi=[:])`
This overload should be the one this code ends up using, which is an implied 1 flip.

PR #753, opened 2020-06-18
Cleanup CARRACOSTA_GX_239's Stone Age GX attack

scoot404:


src/tcgwars/logic/impl/gen7/SunMoonPromos.groovy
```diff
@@ -3291,16 +3291,15 @@ public enum SunMoonPromos implements LogicCardInfo {
           move "Stone Age GX", {
             text "Put any number of Pokémon that evolve from Unidentified Fossil from your discard pile onto your Bench. (You can’t use more than 1 GX attack in a game.)"
             energyCost C
+            def babyFossils = { my.discard.findAll { it.cardTypes.is(EVOLUTION) && it.predecessor == "Unidentified Fossil" } }
             attackRequirement {
               gxCheck()
-              assert my.discard.findAll { it.cardTypes.is(EVOLUTION) && it.predecessor == "Unidentified Fossil" } : "No Pokémon that evolve from Unidentified Fossil in your discard pile."
+              assert babyFossils : "No Pokémon that evolve from Unidentified Fossil in your discard pile."
```
flagrama, line 3297: This needs to be a call to the closure as well. 
```suggestion
              assert babyFossils() : "No Pokémon that evolve from Unidentified Fossil in your discard pile."
```

PR #754, opened 2020-06-18
Took out testing code for Dark Tyranitar

Grayahm:


flagrama:
Perhaps this could be replaced with text saying that the ability activated instead?

Grayahm:
Good thinking

PR #756, opened 2020-06-18
Revert "Fix energy icons not updating early enough"

flagrama:
Reverts axpendix/tcgone-engine-contrib#755

This seems to just break special energy instead. I guess the delayed update is the best option at the moment? Will need to ask @axpendix if there is a way to refresh more frequently. Seems to only refresh the images when the energy count are read due to using an attack or similar.

After this revert, when two counter energy are attached and you are behind on prizes, both will show as a single colorless if attached before you were behind on prizes. IIRC one attached after will use the correct image, but I'll have to verify that after the revert. Trying to use an attack that requires a certain type of energy will evaluate the Energy which prompts the override, but it doesn't seem to reevaluate all of the energy? It seems like it evaluates enough to know "this attack should work" so in some cases the second Counter Energy might not get reevaluated and will continue to show as C even though it is Rainbow,Rainbow.

### DIFF
--- a/src/tcgwars/logic/impl/gen3/Deoxys.groovy
+++ b/src/tcgwars/logic/impl/gen3/Deoxys.groovy
@@ -2662,8 +2662,6 @@ public enum Deoxys implements LogicCardInfo {
           def check = {
             if(!it.evolution || it.EX){discard thisCard}
           }
-          def provides3Rainbow = {self && self.owner.pbg.prizeCardSet.size() > self.owner.opposite.pbg.prizeCardSet.size()}
-          typeImagesOverride = provides3Rainbow() ? [RAINBOW, RAINBOW, RAINBOW] : [C]
           onPlay {reason->
             eff = delayed {
               after EVOLVE, self, {check(self)}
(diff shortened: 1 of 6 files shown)


axpendix:
Please do not set image overrides for non-rainbow energy. It is not designed to be updated that way.

flagrama:
Does that mean you want me to undo most of this as well? https://github.com/axpendix/tcgone-engine-contrib/pull/667 and just reapply the double rainbow/rainbow copy/capture energy stuff from it?

axpendix:
Appreaciate the concern about this situation, but I am sorry this is not the solution and needs to be reverted.

Please remove all `owner.typeImagesOverride = ...` statements in these cards, that field is not meant to be used from that location. Only those cards will show rainbow icon with an unpredictable delay and many other cards won't have this feature. We want to avoid inconsistent behavior as they get bug reported all the time [happened in past lots of times, players get confused and irritated a lot with these behaviours] We don't want confusion. Displaying Colorless energy is fine for now. 

On the other hand in the future the engine can figure out if an energy provides multiple types and adjust the icon automatically.

PR #760, opened 2020-06-19
Fixed Bench-Reducing effect cards (Giant Stump, Road Block, Bellelba & Brycen-Man) not restoring abilities, plus some simplifications

starg09:
~~def didn't just copy that last card's name from the groovy file~~

### DIFF
--- a/src/tcgwars/logic/impl/gen3/LegendMaker.groovy
+++ b/src/tcgwars/logic/impl/gen3/LegendMaker.groovy
@@ -1074,12 +1074,8 @@ public enum LegendMaker implements LogicCardInfo {
         pokeBody "Luna Shade", {
           text "As long as you have Lunatone in play, each player's [C] Pokémon (excluding Pokémon-ex) can't use any Poké-Powers."
           getterA (IS_ABILITY_BLOCKED) { Holder h ->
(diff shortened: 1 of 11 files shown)


src/tcgwars/logic/impl/gen3/LegendMaker.groovy
```diff
@@ -1074,12 +1074,8 @@ public enum LegendMaker implements LogicCardInfo {
...
           text "As long as you have Lunatone in play, each player's [C] Pokémon (excluding Pokémon-ex) can't use any Poké-Powers."
           getterA (IS_ABILITY_BLOCKED) { Holder h ->
-            if (!h.effect.target.topPokemonCard.cardTypes.is(EX) && h.effect.target.types.contains(C)) {
-              if (h.effect.ability instanceof PokePower) {
-                if (self.owner.pbg.all.find{it.name.contains("Lunatone")}) {
-                  h.object=true
-                }
-              }
+            if (my.all.find{it.name == 'Lunatone'} && h.effect.target.types.contains(C) && !h.effect.target.EX && h.effect.ability instanceof PokePower) {
+              h.object=true
```
axpendix, line 1078: unfortunately this won't work in opponent's turn with `my`. 

`self.owner.pbg.all.any{it.name == 'Lunatone'}`

PR #764, opened 2020-06-19
DAB: Implemented some trainers (170-177, 180)

starg09:


src/tcgwars/logic/impl/gen8/DarknessAblaze.groovy
```diff
@@ -3732,48 +3742,139 @@ public enum DarknessAblaze implements LogicCardInfo {
...
+            assert my.deck.notEmpty
+          }
       };
       case POKEMON_BREEDER_S_NURTURING_174:
       return supporter (this) {
         text "Choose up to 2 of your Pokémon in play. Search your deck for a card that evolves from each of those Pokémon and put it on top of the Pokémon to evolve it. Then, shuffle your deck. (You can’t use this card during your first turn or on a Pokémon that just came into play.)"
         onPlay {
+          def pcsList = my.all.findAll {it.turnCount < bg.turnCount}
+          if (pcsList.size() > 1) {
+            pcsList = pcsList.select(max:2, "Choose up to 2 of your Pokémon in play that can make a regular evolve during this turn.")
```
flagrama, line 3774: If this ends up using the CardList select, I think you need to set the minimum as well since you can't play a Trainer without any effect. If this ends up using a PcsList instead you might end up having to cast it to a CardList since the PcsList doesn't have min/max/count parameters from what I see.

src/tcgwars/logic/impl/gen8/DarknessAblaze.groovy
```diff
@@ -3732,48 +3742,139 @@ public enum DarknessAblaze implements LogicCardInfo {
...
+            })
+            if (sel) { evolve(pcs, sel.first(), OTHER) }
+          }
+
+          shuffleDeck()
         }
         playRequirement{
+          assert my.deck : "Your deck is empty."
+          assert bg.turnCount > 2 : "Cannot use this card during your first turn."
+          assert my.all.findAll {it.turnCount < bg.turnCount}
```
flagrama, line 3789: Would be nice to mention that it isn't usable on Pokémon put in play this turn here I think. Something like
```suggestion
          assert my.all.findAll {it.turnCount < bg.turnCount} : "Can't be used on a Pokémon brought in this turn"
```

src/tcgwars/logic/impl/gen8/DarknessAblaze.groovy
```diff
@@ -3802,9 +3903,16 @@ public enum DarknessAblaze implements LogicCardInfo {
       case SPIKEMUTH_180:
       return stadium (this) {
         text "Whenever a player’s Pokémon is moved from the Active Spot to the Bench during their turn put 2 damage counters on that Pokémon."
+        def eff
         onPlay {
+          eff = delayed{
+            after FALL_BACK, {
+              directDamage 20, ef.fallenBack//, STADIUM_CARD ???
```
flagrama, line 3910: I think TRAINER_CARD is still the correct source currently. Definitely don't want it to be attack at the very least. Maybe OTHER could work?

src/tcgwars/logic/impl/gen8/DarknessAblaze.groovy
flagrama, line 3789: You also need to check that there is at least a possibility of an evolution. I'm not sure how to do that currently, but the relevant ruling I do have

> Q. Can I use Wally or Evosoda on a Snorlax to search my deck, even though there is {currently} no evolution of Snorlax?
A. No, there has to at least be the possibility of a valid evolution in order to use this card. Sorry. (Jul 26, 2018 TPCi Rules Team)

src/tcgwars/logic/impl/gen8/DarknessAblaze.groovy
```diff
@@ -3716,14 +3716,24 @@ public enum DarknessAblaze implements LogicCardInfo {
       return itemCard (this) {
         text "Flip 2 coins. If both are heads, choose 1 card from your discard pile, reveal it, and put it into your hand."
         onPlay {
+          def heads = 2
+          flip 2, {}, {heads -= 1}
+
+          if (heads == 2) {
```
ufodynasty, line 3722:
```suggestion
          flip 2,{},{},[2:{my.discard.select("Which card will you put back into your hand?").move(my.hand)}]
``` is a cleaner way to do this

File: src/tcgwars/logic/impl/gen8/DarknessAblaze.groovy (hunk 2 above)
Comment (starg09) [reply src/tcgwars/logic/impl/gen8/DarknessAblaze.groovy]: Hmm, I agree on this. But it'd probably need a special method on the server, similar to "basicsFromStage2" does for cards like rare candy.

Said method though, should not check for specific formats:

> Q. Can I use Wally or Evosoda to search my deck for a card that is not in the current tournament format? For example, could I use it on Raichu to search for a Raichu BREAK after it rotates out?
>
> A. Yes, you can. Even though a card may not be in the tournament format it still exists and can be searched for even though it is known to fail. (Jul 26, 2018 TPCi Rules Team)

Comment (starg09) [reply src/tcgwars/logic/impl/gen8/DarknessAblaze.groovy]: (cc @axpendix)

File: src/tcgwars/logic/impl/gen8/DarknessAblaze.groovy (hunk 4 above)
Comment (starg09) [reply src/tcgwars/logic/impl/gen8/DarknessAblaze.groovy:3722]: Personally find the if cleaner to read, without added complexity as far as I can see.

File: src/tcgwars/logic/impl/gen8/DarknessAblaze.groovy (hunk 3 above)
Comment (starg09) [reply src/tcgwars/logic/impl/gen8/DarknessAblaze.groovy:3910]: hmm, OTHER might do. Reason I want to avoid TRAINER_CARD here is due to some abilities (I think?) and (this one definitely) ancient traits only being affected by trainers that aren't stadiums or tools.

File: src/tcgwars/logic/impl/gen8/DarknessAblaze.groovy (hunk 4 above)
Comment (flagrama) [reply src/tcgwars/logic/impl/gen8/DarknessAblaze.groovy:3722]: I could see this method being cleaner if you defined the closure outside of the map, and just provided it to the map.
```Groovy
def fromDiscard = {my.discard.select("Which card will you put back into your hand?").move(my.hand)}
flip 2, {}, {}, [2:fromDiscard]
```

src/tcgwars/logic/impl/gen8/DarknessAblaze.groovy
starg09: ...However, this could be solved by adding a filter on the static barrier() method! _(which I legit wasn't aware was there in the first place)_

src/tcgwars/logic/impl/gen8/DarknessAblaze.groovy
axpendix: that is doable but tough, there is currently no way to expose stage1 -> stage2 and basic -> stage1 evolution lines. the system that delivers basic -> stage2 evolution lines is called `carddb-api` and it is a separate module of its own written in spring boot. if you are interested in adding that extra endpoint, I am willing to do the engine integration to your endpoint which will let us implement this card perfectly. maybe that could also be used in some other similar type of cards such as evolution fields and such. @flagrama @starg09

starg09:
> this will fail when there is no selectable pokemon in the deck. please do not use `count:1` here. instead, look at the result of the call.

Good catch, will correct it in a bit

> I also started to think clearing out automatically from unresolveable selections would also be great.

Yeah, though besides that I may take a peek into some of these cards at some point (e.g. Volkner, this one, Rosa). Might be nicer if everything were picked in a single search.

starg09:
Aaaaaand that reminded me of rosa, which does exactly that...

src/tcgwars/logic/impl/gen8/DarknessAblaze.groovy
flagrama: I'm certainly interested in doing anything that can help us make the cards more accurate when played. Might need pointed out where the basic -> stage2 thing is being done, but hopefully can figure it out from there.

src/tcgwars/logic/impl/gen8/DarknessAblaze.groovy
```diff
@@ -3732,48 +3742,139 @@ public enum DarknessAblaze implements LogicCardInfo {
...
-        onPlay {
-        }
-        playRequirement{
-        }
+        text "Search your deck for a [D] Pokémon and an Energy card, reveal them, and put them into your hand. Then, shuffle your deck."
+          onPlay {
+            my.deck.search(count: 1, "Search your deck for a [D] Pokémon",pokemonTypeFilter(D)).moveTo(my.hand)
+            my.deck.search("Search your deck for an Energy card",cardTypeFilter(ENERGY)).moveTo(my.hand)
+            shuffleDeck()
+          }
```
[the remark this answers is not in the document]
starg09: Oh, missed this comment before. Yeah, Volkner may need this implemented as well... Hopefully last commit works well, specifically the specific-type pokemon filter)

starg09:
Ok, all comments addressed I think, specific notes:

* Need to test Breeder's Nurturing working well in some edge-cases, and not treating the PCSList as a CardList.
* The "can this Pokémon evolve" involves a ton of cards, current implementation should follow line with how the rest is currently handled.
* Piers should in theory filter properly now, needs testing tho.

src/tcgwars/logic/impl/gen8/DarknessAblaze.groovy
flagrama: Does this still need addressed? Other stadiums seem to use TRAINER_CARD currently at least, and the default is going to be ATTACK which will be blocked by many things that it shouldn't be.

starg09: Hmm, will commit it as a TRAINER_CARD effect and try solve barrier later on.

src/tcgwars/logic/impl/gen8/DarknessAblaze.groovy
axpendix: sent you the invite @flagrama 
contact me on discord

src/tcgwars/logic/impl/gen8/DarknessAblaze.groovy
axpendix: @flagrama  is correct. `pcsList.select` returns a single `pcs` not a list.

src/tcgwars/logic/impl/gen8/DarknessAblaze.groovy
```diff
@@ -3725,48 +3734,145 @@ public enum DarknessAblaze implements LogicCardInfo {
...
+          if (pcsList.size() > 1) {
+            pcsList = pcsList.select(max:2, "Choose up to 2 of your Pokémon in play that can make a regular evolve during this turn.")
+          }
+
+          pcsList.each{ pcs =>
+            def sel = deck.search ("Select a Pokémon that evolves from $pcs.name.", {
+              it.cardTypes.is(EVOLUTION) && it.predecessor == pcs.name
+            })
+            if (sel) { evolve(pcs, sel.first(), OTHER) }
+          }
```
axpendix, line 3780:
```suggestion
          multiSelect (my.all.findAll {it.turnCount < bg.turnCount}, 2, "Choose up to 2 of your Pokémon in play that can make a regular evolve during this turn.").each { pcs =>
            def sel = deck.search ("Select a Pokémon that evolves from $pcs.name.", {
              it.cardTypes.is(EVOLUTION) && it.predecessor == pcs.name
            })
            if (sel) { evolve(pcs, sel.first(), OTHER) }
          }
```

src/tcgwars/logic/impl/gen8/DarknessAblaze.groovy
axpendix: Resolved with commit

starg09: I see. Will correct this so it chooses one by one then.

axpendix:
Leaving Wally/Evosoda fix to another PR, it'll be too late for this one.

PR #768, opened 2020-06-20
Fix Delta Rainbow Energy being misnamed in POP5

starg09:


src/tcgwars/logic/impl/gen3/PopSeries5.groovy
```diff
@@ -49,9 +49,9 @@ public enum PopSeries5 implements LogicCardInfo {
   BILL_S_MAINTENANCE_6 ("Bill's Maintenance", 6, Rarity.UNCOMMON, [TRAINER]),
   RARE_CANDY_7 ("Rare Candy", 7, Rarity.UNCOMMON, [TRAINER]),
   BOOST_ENERGY_8 ("Boost Energy", 8, Rarity.UNCOMMON, [TRAINER]),
-  DELTA_RAINBOW_ENERGY_9 ("Delta Rainbow Energy", 9, Rarity.UNCOMMON, [ENERGY, SPECIAL_ENERGY]),
   CHARMANDER_DELTA_10 ("Charmander", 10, Rarity.COMMON, [POKEMON, BASIC, DELTA, _LIGHTNING_]),
   MEOWTH_DELTA_11 ("Meowth", 11, Rarity.COMMON, [POKEMON, BASIC, DELTA, _DARKNESS_]),
+  DELTA_RAINBOW_ENERGY_9 ("δ Rainbow Energy", 9, Rarity.UNCOMMON, [ENERGY, SPECIAL_ENERGY]),
```
luongthomas, line 54: Why did the entry get moved to after meowth_delta_11?

starg09: Not sure, but will correct it one sec

PR #778, opened 2020-06-20
Re #580: Fixed cards method calling, added some all examples, made the final part of the failMessage conditional

starg09:


starg09:
Gonna add some more test cases for `all`

PR #779, opened 2020-06-20
Electrode ex fix

Grayahm:
Attaching energy cards with Extra Energy Bomb is no longer optional

### DIFF
--- a/src/tcgwars/logic/impl/gen3/FireRedLeafGreen.groovy
+++ b/src/tcgwars/logic/impl/gen3/FireRedLeafGreen.groovy
@@ -2533,7 +2533,7 @@ public enum FireRedLeafGreen implements LogicCardInfo {
               checkLastTurn()
               powerUsed()
               new Knockout(self).run(bg)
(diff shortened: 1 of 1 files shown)


src/tcgwars/logic/impl/gen3/FireRedLeafGreen.groovy
```diff
@@ -2533,7 +2533,7 @@ public enum FireRedLeafGreen implements LogicCardInfo {
               checkLastTurn()
               powerUsed()
               new Knockout(self).run(bg)
-              5.times{ attachEnergyFrom(may : true,my.discard, my.all.findAll{!it.EX}) }
+              5.times{ attachEnergyFrom(may : false, my.discard, my.all.findAll{!it.EX}) }
```
starg09, line 2536: Just a little nitpick, this might as well be entirely removed :P

```suggestion
              5.times{ attachEnergyFrom(my.discard, my.all.findAll{!it.EX}) }
```

PR #785, opened 2020-06-21
assertAnyPIP(): Migrated TEU-RCL, improved and expanded the variant filter, added stage-based filter

starg09:
* New assert methods are now used by all cards from Team Up to Rebel Clash.
* Reworked this whole set of params into a single `hasVariants` filter.
  - Receives either a single CardType or a list of them.
  - Added support for Tag Team Pokémon, Pokémon-ex and Delta Pokémon, in addition to previous ones.
  - There's now a `negateVariants` param. If set to true, only PokemonCardSets **not** matching any of the given variants will be counted as valid.
* Added a `isStage` filter, based on evolution stages. Works in a similar fashion to hasVariants but filtered PokemonCardSets must follow all of the given filters here.
  - Possible values: 
    + `UNEVOLVED` (`it.notEvolution`, pre-DP "Basic Pokémon")
    + `EVOLVED` (`it.evolution`)
    + `BASIC` (`it.basic`, DP-onwards "Basic Pokémon")
    + `STAGE1`
    + `STAGE2`
    + `EVOLUTION` (`it.realEvolution`)
* Reordered the print order of some elements, so they're more consistent with the corresponding cards text.


axpendix:
The parameterization has become too complex to keep in mind IMO. You wouldn’t want to check these params everytime u call this function. The concept should be more generic somehow. Essentially what this method does it it injects auto generated message to assert statements. That’s all we need from this function. Maybe there is a better way to do it via some groovy magic.

axpendix:
Maybe something with [decorator pattern](http://docs.groovy-lang.org/docs/next/html/documentation/design-pattern-in-groovy.html#_decorating_with_an_interceptor)?

starg09:
> The parameterization has become too complex to keep in mind IMO. You wouldn’t want to check these params everytime u call this function. The concept should be more generic somehow. Essentially what this method does it it injects auto generated message to assert statements. That’s all we need from this function. Maybe there is a better way to do it via some groovy magic.

Yeah, got a feeling it might get this reply. Def needs a scaling down, tho I still consider the amount of cards checking worth it being passable in the params map somehow.

axpendix:
IMO `assertMyBench {it.pokemonGX && it.basic}` is way more readable than `assertMyBench (hasPokemonGX: true, basic: true}` because with the latter you have to remember the individual parameters and how they work. 

So that if one uses `assertMyBench {it.pokemonGX && it.basic}` -> can we decorate some methods of `PokemonCardSet` and generate the message automatically? Possibly, yes. 

Except it won't be possible to understand `&&` or `||` in middle. We can use a generic clause such as `and/or` or simply `/`, so that above example can print something like: "No GX / Basic Pokemon on your bench"; a great improvement from `Unsatisfied condition`.

starg09:
@axpendix not using decorators (yet), but last commit simplifies the filter for Pokemon Variants into a single, cleaner to read param that's in line with some other statics (thought of naming it hasCardType, but that'd clash with stages, which aren't filtered in the same way, so used a different term). 

Sadly can't do this as easily for the other filter (Basic/Stage 1/Stage 2/Evolution/Evolved/Unevolved), as the last 3 don't have a CARD_TYPE I can use as possible values. Thought of maybe creating an EvolutionStage enum, but maybe this can be solved in some other way I'm not thinking of...

src/tcgwars/logic/impl/gen7/TeamUp.groovy
```diff
@@ -589,14 +589,15 @@ public enum TeamUp implements LogicCardInfo {
             text "Once during your turn (before your attack), you may discard 2 Fire Energy cards from your hand. If you do, switch 1 of your opponent's Benched Pokémon with their Active Pokémon."
             actionA{
               checkLastTurn()
-              assert opp.bench.notEmpty() : "Opponent bench empty"
```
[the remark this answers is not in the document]
starg09: Oh, not sure why I edited here should've moved it to a general fixes PR.

But yeah, a chunk of cards with this kind of "cost" (actually two consecutive effects) were ruled a couple months ago, after the rules team contacting the JP side.

> Q. Can I use Ninetales' "Nine Temptations" Ability if my opponent has no benched Pokemon?
>
> A. Yes, you can discard 2 {R} Energy cards from your hand, and that's it. (Team Up FAQ; Jan 31, 2019 TPCi Rules Team)

I'll revert this change from the PR though, and go through all the cards involved (I remember a list being available somewhere in Pokegym, from when the rulings were added.)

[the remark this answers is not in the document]
starg09: Aha, here it is 😄 

https://pokegym.net/index.php/2019/04/01/trash-talk/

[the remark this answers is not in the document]
axpendix: Okay, sorry for the trash talk :)

starg09:
Ok, rebased and fixed merge conflicts.

axpendix:
Will review asap

src/tcgwars/logic/impl/gen7/CosmicEclipse.groovy
```diff
@@ -4862,7 +4860,11 @@ public enum CosmicEclipse implements LogicCardInfo {
           playRequirement{
             assert my.deck : "Your deck is empty."
             assert bg.turnCount > 2 : "Cannot use this card during your first turn."
-            assert my.all.any{it.turnCount < bg.turnCount && it.lastEvolved < bg.turnCount} : "Cannot use this on Pokémon put into play this turn"
+            assertMyAll(info: "that weren't put into play this turn.", {
+              (it.turnCount < bg.turnCount && it.lastEvolved < bg.turnCount) &&
+              //TODO: Change this. Still can't learn if a Pokémon has a GX evolution, but this will at least filter obvious non-allowed cases for standard and expanded.
+              !(it.pokemonEX || it.pokemonBreak || it.pokemonGX || it.pokemonV || it.topPokemonCard.cardTypes.is(PRISM_STAR))
+              })
```
axpendix, line 4867: Not a fan of this change, looks too complex.

src/tcgwars/logic/impl/gen7/UnifiedMinds.groovy
```diff
@@ -1179,10 +1179,10 @@ public enum UnifiedMinds implements LogicCardInfo {
             text "This attack does 20 damage for each Basculin you have in play to 1 of your opponent's Benched Pokémon. (Don't apply Weakness and Resistance for Benched Pokémon.)"
             energyCost C
             attackRequirement {
-              assert my.bench.findAll({ it.name == "Basculin" }) : "You have no Basculin on your bench."
+              assertMyAll(info: "that are Basculin", { it.name == "Basculin" })
```
axpendix, line 1182: bug detected + opponent needs to have a bench

src/tcgwars/logic/impl/gen7/UnifiedMinds.groovy
```diff
@@ -4437,8 +4437,8 @@ public enum UnifiedMinds implements LogicCardInfo {
             }
           }
           playRequirement{
-            assert opp.all.size() >= 2 : "Opponent only has one Pokémon in play."
-            assert opp.all.findAll {it.numberOfDamageCounters} : "There are no damage counters to move."
+            assertOppBench(repText: true, info: "Your opponent only has 1 Pokémon in play")
```
axpendix, line 4440: Not a fan of this TBH, the previous one is much more clear. maybe even change it to `.size() > 1` instead of `.size() >= 2`.

src/tcgwars/logic/impl/gen7/UnbrokenBonds.groovy
```diff
@@ -1252,8 +1252,8 @@ public enum UnbrokenBonds implements LogicCardInfo {
             text "Move an Energy from 1 of your opponent's Pokémon to another of their Pokémon. If you do, put 3 damage counters on the Pokémon you moved the Energy to."
             energyCost C
             attackRequirement {
-              assert opp.all.findAll{it.cards.filterByType(ENERGY)}
-              assert opp.bench
+              assertOppAll(info: "with Energy attached to them", {it.cards.filterByType(ENERGY)})
+              assertOppBench(repText: true, info: "Your opponent only has 1 Pokémon in play")
```
axpendix, line 1256: Not a fan of this TBH, the previous one is much more clear.

src/tcgwars/logic/impl/gen7/UnifiedMinds.groovy
```diff
@@ -4605,8 +4605,8 @@ public enum UnifiedMinds implements LogicCardInfo {
             pcs.cards.filterByType(ENERGY).select(max:2, "Select up to 2 Energy cards to move to the target.").each{energySwitch(pcs,tar,it)}
           }
           playRequirement{
-            assert my.all.findAll {it.tagTeam && it.cards.filterByType(ENERGY)} : "No valid targets."
-            assert my.all.size() >= 2 : "You only have one Pokémon in play."
+            assertMyBench(repText: true, info: "You only have 1 Pokémon in play")
```
axpendix, line 4608: Same one

src/tcgwars/logic/impl/gen7/CosmicEclipse.groovy
```diff
@@ -4748,7 +4747,11 @@ public enum CosmicEclipse implements LogicCardInfo {
             }
           }
           playRequirement {
+<<<<<<< HEAD
             assertMyBenched()
+=======
+            assertMyBench()
+>>>>>>> f2455e5e... Benched -> Bench
```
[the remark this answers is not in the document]
starg09: Think this was solved already?

src/tcgwars/logic/groovy/TcgStatics.groovy
```diff
@@ -1158,7 +1159,9 @@ class TcgStatics {
...
+   *   + hasVariants: A list of specific CardType values (currently: POKEMON_V | VMAX | TAG_TEAM | POKEMON_GX | POKEMON_EX | DELTA | EX). If set, the area filter will only accept PCS that have at least one of these CardTypes on its top card; otherwise, it'll take any Pokémon.
+   *   + negateVariants: If set to true, hasVariants will be inverted: only PCS that are __not__ any of the variants provided will be accepted.
+   *   + isStage: A list of specific CardType values (currently: EVOLVED | UNEVOLVED | BASIC | STAGE1 | STAGE2 | EVOLUTION). If set, the area filter will only accept PCS that return true for every single one of the included values; otherwise, it'll take any Pokémon regardless of stage.
    *   + info: If set, it'll replace the end of the failed assert warning with a custom text, instead of the default "follow the stated condition(s)".
    *   + repText: If true, params.info will override the entirety of the failed assert warning.
```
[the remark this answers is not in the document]
starg09: That does sound clearer to me 👍

src/tcgwars/logic/impl/gen7/UnifiedMinds.groovy
starg09: Which would be the bug here? Wouldn't `checkedArea.any(areaFilter)` return false if checkedArea is empty?
Will push the opp.bench check on the repo directly.

src/tcgwars/logic/impl/gen7/UnifiedMinds.groovy
starg09: Yeah, I see what you mean. Not everything needs to use the static, this would be a good example.
Will revert these in a bit

src/tcgwars/logic/impl/gen7/CosmicEclipse.groovy
starg09: Left a TODO, but yeah it's not that big a deal. Will be fixed once that's implemented, for now it can remain a general check as before.

src/tcgwars/logic/impl/gen7/UnifiedMinds.groovy
axpendix: When this attack is used by another pokemon (copied), let's say "Mew", does the attack look for "Basculin" or "Mew"? 

If "Basculin", then `assertMyAll(info: "that are Basculin", { it.name == "Basculin" })` should stay. Else, removed. 

What I meant by "bug" was adding a `assertOppBench()` call

starg09: Yeah, in this case it searches specifically for "Basculin". No replicating mon, no Basculin-EX/GX/V. Similar to how Durant and now Falinks work (tho Falinks only looks for Falinks in the name, counting Falinks V)

PR #786, opened 2020-06-21
Removed some NOT_IMPLEMENTED tags, added a TODO on nurturing and debug for assertAnyPIP, fixed Sand Damage

starg09:


src/tcgwars/logic/impl/gen3/DragonFrontiers.groovy
```diff
@@ -1903,10 +1903,12 @@ public enum DragonFrontiers implements LogicCardInfo {
           text "As long as Flygon ex is your Active Pokémon, put 1 damage counter on each of your opponent's Benched Basic Pokémon between turns. You can't use more than 1 Sand Damage Poké-Body between turns."
           delayedA {
             before BEGIN_TURN, {
+              boolean flag = 1
```
flagrama, line 1906: Defines boolean specifically, but then proceeds to use an integer value instead of a boolean value to assign to it :laughing:

starg09: I blame whoever did weezing, and my brain post handling a father's day lunch videoconference video 😛

src/tcgwars/logic/impl/gen3/DragonFrontiers.groovy
```diff
@@ -1903,10 +1903,12 @@ public enum DragonFrontiers implements LogicCardInfo {
           text "As long as Flygon ex is your Active Pokémon, put 1 damage counter on each of your opponent's Benched Basic Pokémon between turns. You can't use more than 1 Sand Damage Poké-Body between turns."
           delayedA {
             before BEGIN_TURN, {
+              boolean flag = true
               if (self.active) {
                 self.owner.opposite.pbg.bench.each {
                   if (!it.evolution) {
-                    directDamage 10, it
+                    if(flag) {bc "Sand Damage activates"; flag = false}
+                    directDamage 10, it, SRC_ABILITY
```
flagrama, line 1911: Are you sure this works right? To me it looks like it activates always. The boolean does not accomplish anything since you explicitly set it to true in the ability itself, so all copies will see it as true. I think you should define the flag outside, assign the flag to false in `before BEGIN_TURN` then reassign it to true in an `after BEGIN_TURN`

You may have to storeObject as well to ensure all copies see the same value instead of creating their own copy of the flag.

starg09: Seems to work well on weezing? Feel like I tested something with it before, should trigger only once but let me confirm.

starg09: Beyond the boolean set to 1 ofc :P

flagrama: Double checking it, I don't think this flag matters at all since it relies on the Pokémon being the active. The flag seems to just be extraneous data that isn't required. I think the addition of the effect text to the log is the only thing needed.

starg09: Thought it worked well as it was already when porting it over from weezing since it only prints Sand Damage/Detention Gas usage if any Pokémon is actually damaged by it.

Having it print every turn even tho it did nothing, or checking for any Pokémon from the given area following it before going through the same area all over again felt like a waste.

flagrama: Ah, okay, there we go. I guess the flag does have a purpose, I just wasn't seeing it.

starg09: Oh, didn't pass the screenshot. Just tested with two weezing as actives, and placing basics across the bench. One message per weezing before its damage, right before starting

PR #792, opened 2020-06-21
Fix Dual Ball, Remove debug code, fix old Drag Off attacks

starg09:
No longer needed for now, but it improves #785 slightly.


starg09:
(Left on draft for a min while checking Garde ex, that works fine tho)

PR #793, opened 2020-06-22
Add suppressLog to Swoop! Teleporter

Grayahm:


src/tcgwars/logic/impl/gen3/TeamRocketReturns.groovy
```diff
@@ -2403,7 +2403,7 @@ public enum TeamRocketReturns implements LogicCardInfo {
             def spc = pcs.specialConditions
             def selected = my.deck.search(max:1,"Select a Basic Pokémon (excluding Pokémon-ex)",{it.cardTypes.is(BASIC) && it.cardTypes.isNot(EX)})
             if (selected) {
-              selected.moveTo(pcs.cards)
+              selected.moveTo(suppressLog: true, pcs.cards)
```
ufodynasty, line 2406: I'm curious if this will work, however generally the syntax is hidden:true

flagrama: Doesn't hidden still say "put n cards to <thislocation>"? I think the idea here is to suppress that message too, since in this case it is confusing to a player. I could see creating our own log message here saying something like "$selected retained $pcs's attached cards" to still keep that information in a less confusing way.

PR #798, opened 2020-06-23
Delta Species fixes

starg09:


src/tcgwars/logic/impl/gen3/DeltaSpecies.groovy
```diff
@@ -371,7 +374,7 @@ public enum DeltaSpecies implements LogicCardInfo {
             powerUsed()
 
             def source = my.all.findAll { it.cards.filterByType(ENERGY) }.select("Choose a Pokemon to move an Energy from")
-            def target = my.all.select("Select a Pokemon to move the Energy to")
+            def target = my.all.findAll{it != source}.select("Select a Pokemon to move the Energy to")
```
ufodynasty, line 377: Personally I like ordering these From -> EnergyCard -> To while this is ordered From -> To -> EnergyCard. This can be done like 
```suggestion
def source = my.all.findAll { it.cards.filterByType(ENERGY) }.select("Choose a Pokemon to move an Energy from")
def energyCard = source.cards.filterByType(ENERGY).select("Choose the energy to move")
            def target = my.all.findAll{it != source}.select("Select a Pokemon to move the Energy to")
            energySwitch(source,target,energyCard)
```
Its up to you tho

src/tcgwars/logic/impl/gen3/DeltaSpecies.groovy
```diff
@@ -3154,7 +3185,7 @@ public enum DeltaSpecies implements LogicCardInfo {
           text "Any damage done to Azumarill by attacks from [R] Pokémon and [W] Pokémon is reduced by 30 (after applying Weakness and Resistance)."
           delayedA {
             before APPLY_ATTACK_DAMAGES, {
-              if(ef.attacker.owner != self.owner && (ef.attacker.types.contains(R) || ef.attacker.types.contains(W))) {
+              if(ef.attacker.types.contains(R) || ef.attacker.types.contains(W)) {
```
luongthomas, line 3188: I'm curious how this would work with confusion damage onto itself

src/tcgwars/logic/impl/gen3/DeltaSpecies.groovy
starg09: Yeah, that sounds good. Will do a separate commit just as to remove the line below.

src/tcgwars/logic/impl/gen3/DeltaSpecies.groovy
```diff
@@ -3119,21 +3151,20 @@ public enum DeltaSpecies implements LogicCardInfo {
...
-
-            def takenPrizes = opp.prizeCardSet.takenCount
-            def energies = my.discard.filterByType(ENERGY).size()
-            if (takenPrizes && energies) {
-              def count = Math.min(takenPrizes, energies)
-              (1..count).each {
-                attachEnergyFrom(type:M, my.discard, self)
+            afterDamage{
+              def takenPrizes = opp.prizeCardSet.takenCount
+              def energies = my.discard.filterByType(ENERGY).size()
```
luongthomas, line 3164: How about `filterByType(ENERGY).filterByEnergyType(M).size()`?

src/tcgwars/logic/impl/gen3/DeltaSpecies.groovy
starg09: Hmm, can't remember when confusion damage was changed to 3 counters. Think it was before R&S, so it would just place the counters and call it a day :P

src/tcgwars/logic/impl/gen3/DeltaSpecies.groovy
starg09: Good catch, will commit in a bit 👍

src/tcgwars/logic/impl/gen3/DeltaSpecies.groovy
starg09: > Confused Pokémon that fail to attack have 3 damage counters placed on them (compared to dealing 20 damage in the Wizards era).

Yup, luckily no weird interactions with that, haha.

src/tcgwars/logic/impl/gen3/DeltaSpecies.groovy
```diff
@@ -2669,15 +2716,15 @@ public enum DeltaSpecies implements LogicCardInfo {
         text "You can play only one Supporter card each turn. When you play this card, put it next to your Active Pokémon. When your turn ends, discard this card." +
           "Discard a card from your hand. If you can't discard a card from your hand, you can't play this card. Search your discard pile for 3 basic Energy cards and any combination of 3 Basic Pokémon or Evolution cards, show them to your opponent, and put them on top of your deck. Shuffle your deck afterward."
         onPlay {
-          def toDiscard = my.hand.getExcludedList(thisCard).select(count:1, "Discard a card to discard.")
+          def toDiscard = my.hand.getExcludedList(thisCard).select(count:1, "Discard a card from your hand in order to play ${thisCard}.")
           toDiscard.discard()
 
           if (my.discard.filterByType(BASIC_ENERGY)) {
-            my.discard.filterByType(BASIC_ENERGY).select(max: 3, "Select 3 Basic Energies to shuffle into your deck").moveTo(my.deck)
+            my.discard.filterByType(BASIC_ENERGY).select(count: 3, "Select 3 Basic Energies to shuffle into your deck").moveTo(my.deck)
```
luongthomas, line 2723: Does this work if you only have 2 Basic Energies in discard?

starg09: Yeah, it's used by some other cards already. Just tested on Energy Recycler (Guardians Rising) and it took less than 5 basic energies fine :)

src/tcgwars/logic/impl/gen3/DeltaSpecies.groovy
```diff
@@ -3040,7 +3075,7 @@ public enum DeltaSpecies implements LogicCardInfo {
         pokePower "Evolutionary Swirl", {
           text "Once during your turn, when you play Vaporeon ex from your hand to evolve 1 of your Pokémon, you may have your opponent shuffle his or her hand into his or her deck. Then, your opponent draws up to 4 cards."
           onActivate {
```
flagrama, line 3077: "Your opponent draws up to 4 cards." Can the opponent choose the number of cards they draw then?

src/tcgwars/logic/impl/gen3/DeltaSpecies.groovy
```diff
@@ -555,16 +559,23 @@ public enum DeltaSpecies implements LogicCardInfo {
           energyCost L, M
           onAttack {
             def count=0
```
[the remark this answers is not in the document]
starg09: Yeah, that's fair. Not much difference re: where it's used now.

src/tcgwars/logic/impl/gen3/DeltaSpecies.groovy
starg09: Good catch, tend to miss those texts since modern cards don't really offer that choice :P

PR #821, opened 2020-06-25
Applied some Erratas (TRR, FRLG), plus some notes for a couple already implemented

starg09:
Applied some erratas:
  - Added some comments and edited card text for more clarity (so they're not unerrata'd by mistake)
  - Now Mantine (TRR 45) can only use Ripples from the Active Position:
    > * Mantine's Poké-POWER only works when it is the player's Active Pokémon. Official Wording: Once during your turn (before your attack), if Mantine is your Active Pokémon, you may remove 1 damage counter from 1 of your Pokémon (excluding Mantine). This power can't be used if Mantine is affected by a Special Condition. (Dec 16, 2004 PUI Announcements)
  - Mt. Moon (FRLG 94) now affects Pokémon with <= 70 HP, instead of just < 70:
    > * Mt. Moon should read: "Any Pokémon (both yours and your opponent's) with maximum HP of 70 or less can't use any Poké-Powers." This errata replaces text of "less than 70" with "70 or less". If a Pokémon has a maximum HP of 70, that Pokémon cannot use a Poké-POWER while this Stadium card is in play. (Oct 1, 2004 PUI Announcements)
  - Electrode-ex (FRLG 107) now counts Energy cards discarded instead of total Energy:
    > * Electrode-EX's "Crush and Burn" attack does extra damage based on the number of Energy cards discarded, not the amount of Energy provided. Official Wording: You may discard as many Energy cards as you like attached to your Pokémon in play. If you do, this attack does 30 damage plus 20 more damage for each Energy [B]card[/B] you discarded. (Dec 16, 2004 PUI Announcements)

### DIFF
--- a/src/tcgwars/logic/impl/gen3/Deoxys.groovy
+++ b/src/tcgwars/logic/impl/gen3/Deoxys.groovy
@@ -2461,6 +2461,7 @@ public enum Deoxys implements LogicCardInfo {
       case STRENGTH_CHARM_92:
         return pokemonTool (this) {
           text "Attach a Pokémon Tool to 1 of your Pokémon that doesn’t already have a Pokémon Tool attached to it.\nWhenever an attack from the Pokémon that Strength Charm is attached to does damage to the Active Pokémon, the attack does 10 more damage (before applying Weakness and Resistance). Discard Strength Charm at the end of the turn in which this Pokémon attacks."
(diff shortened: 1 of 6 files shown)


starg09:
Gonna apply some of the fixes manually, one by one as to test them working. Not closing this completely as I may leave the erratas for review.

PR #823, opened 2020-06-26
Added game long text for Girafarig's Foresight

Grayahm:


Grayahm:
The reason is that I changed the wording halfway and forgot to change the other half lol. I’ll fix it when I get home :)

Grayahm:
Fixed

PR #839, opened 2020-06-26
multi energy 2/n is only selectable after multi energy 1/n has been selected

ufodynasty:
The goal of this commit is to increase clarity of which energy you are discarding when you have multiple of the same multi energy attached. Can't request review of starg directly but will ask via discord

### DIFF
--- a/src/tcgwars/logic/impl/gen3/UnseenForces.groovy
+++ b/src/tcgwars/logic/impl/gen3/UnseenForces.groovy
@@ -2602,13 +2602,13 @@ public enum UnseenForces implements LogicCardInfo {
               for (enCard in self.cards.filterByType(ENERGY)){
                 def enTypes = enCard.getEnergyTypes()
                 if (enTypes.size() == 1){
-                  potentialEnergy.add([enTypes.first(), enCard, "${enCard}"])
+                  potentialEnergy.add([enTypes.first(), enCard, "${enCard}"],0)
                 } else {
                   def i = 1
                   def total = enTypes.size()
                   //TODO: Cover multiple of the same energy, add some identifier for e.g. 2 Holon's Castform
                   for (enTypeSet in enTypes) {
-                    potentialEnergy.add([enTypeSet, enCard, "${enCard} - Energy #${i}/${total}"])
+                    potentialEnergy.add([enTypeSet, enCard, "${enCard} - Energy #${i}/${total}"],total-i)
                     i++
                   }
                 }
@@ -2648,7 +2648,7 @@ public enum UnseenForces implements LogicCardInfo {
               def cardsToBeDiscarded = []
 
               for (enReq in energyRequired){
-                def options = potentialEnergy.findAll{it[0].contains(enReq[0])}
+                def options = potentialEnergy.findAll{it[0].contains(enReq[0]) && it[3]==0}
                 def optionsNum = (0..options.size()-1).toList()
                 def optionsLabels = options.collect{it[2]}.toList()
                 def cnt = 0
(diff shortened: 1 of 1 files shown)


src/tcgwars/logic/impl/gen3/UnseenForces.groovy
```diff
@@ -2602,13 +2602,13 @@ public enum UnseenForces implements LogicCardInfo {
...
                 if (enTypes.size() == 1){
-                  potentialEnergy.add([enTypes.first(), enCard, "${enCard}"])
+                  potentialEnergy.add([enTypes.first(), enCard, "${enCard}"],0)
                 } else {
                   def i = 1
                   def total = enTypes.size()
                   //TODO: Cover multiple of the same energy, add some identifier for e.g. 2 Holon's Castform
                   for (enTypeSet in enTypes) {
-                    potentialEnergy.add([enTypeSet, enCard, "${enCard} - Energy #${i}/${total}"])
+                    potentialEnergy.add([enTypeSet, enCard, "${enCard} - Energy #${i}/${total}"],total-i)
```
starg09:
```suggestion
                    potentialEnergy.add([enTypeSet, enCard, "${enCard} - Energy #${i}/${total}",total-i-1])
```

Initially just moved the square bracket, but prev value would do 1-2 instead of 0-1 (for a double energy)

src/tcgwars/logic/impl/gen3/UnseenForces.groovy
```diff
@@ -2602,13 +2602,13 @@ public enum UnseenForces implements LogicCardInfo {
...
                 if (enTypes.size() == 1){
-                  potentialEnergy.add([enTypes.first(), enCard, "${enCard}"])
+                  potentialEnergy.add([enTypes.first(), enCard, "${enCard}"],0)
                 } else {
                   def i = 1
                   def total = enTypes.size()
                   //TODO: Cover multiple of the same energy, add some identifier for e.g. 2 Holon's Castform
                   for (enTypeSet in enTypes) {
-                    potentialEnergy.add([enTypeSet, enCard, "${enCard} - Energy #${i}/${total}"])
+                    potentialEnergy.add([enTypeSet, enCard, "${enCard} - Energy #${i}/${total}"],i-1)
```
starg09, line 2611:
```suggestion
                    potentialEnergy.add([enTypeSet, enCard, "${enCard} - Energy #${i}/${total}",i-1])
```

src/tcgwars/logic/impl/gen3/UnseenForces.groovy
```diff
@@ -2671,6 +2671,14 @@ public enum UnseenForces implements LogicCardInfo {
                     energyToBeDiscarded.add(options[chosenEnergy])
                     potentialEnergy.remove(options[chosenEnergy])
                     options.remove(options[chosenEnergy])
+                    potentialEnergy.each{
```
starg09, line 2674: Might be a lil nitpicky but I'd personally try to avoid doing this loop every single time... Maybe single energies could have 0 or -1 in it[3], and have multiple energies use a value higher? Then check here if chosenOption (added a variable on last push, for when you rebase) has a value higher than 0/-1.

starg09: Oh, maybe I didn't push that huh. Guess it'd be `options[chosenEnergy][3]`

ufodynasty: pushing this for now, don't want to deal with merge conflicts of adding multiple new indicies

starg09: kk, should work fine as it is 👍

PR #854, opened 2020-06-28
Delta Switch only activates when player has Basic Energy in play

Grayahm:
Will this be compatible with the implementation of Holon's Pokemon?

### DIFF
--- a/src/tcgwars/logic/impl/gen3/DeltaSpecies.groovy
+++ b/src/tcgwars/logic/impl/gen3/DeltaSpecies.groovy
@@ -583,7 +583,7 @@ public enum DeltaSpecies implements LogicCardInfo {
         pokePower "Delta Switch", {
           text "Once during your turn, when you put Mewtwo from your hand onto your Bench, you may move any number of basic Energy cards attached to your Pokémon to your other Pokémon (excluding Mewtwo) in any way you like."
           onActivate {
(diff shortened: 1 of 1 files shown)


src/tcgwars/logic/impl/gen3/DeltaSpecies.groovy
```diff
@@ -583,7 +583,7 @@ public enum DeltaSpecies implements LogicCardInfo {
         pokePower "Delta Switch", {
           text "Once during your turn, when you put Mewtwo from your hand onto your Bench, you may move any number of basic Energy cards attached to your Pokémon to your other Pokémon (excluding Mewtwo) in any way you like."
           onActivate {
-            if (it == PLAY_FROM_HAND && confirm("Use Delta Switch?")) {
+            if (it == PLAY_FROM_HAND && my.all.findAll{it.cardTypes.is(BASIC_ENERGY) && confirm("Use Delta Switch?")) {
```
starg09, line 586: Holon's Pokémon are turned into special energy, so this should avoid them without issue. 

I'd use an any instead of findAll for this check though. As a side-note, the check added would be for individual cards, my.all/opponent.all hace PokemonCardSets so something like the filter below would be needed:

```suggestion
            if (it == PLAY_FROM_HAND && my.all.any{it.cards.filterByType(BASIC_ENERGY)} && confirm("Use Delta Switch?")) {
```

PR #855, opened 2020-06-28
AssertionString Enum

flagrama:
This is an attempt to standardize the most commonly used assertion strings, as it is a little awkward that the same situation can give vastly different text, which makes it hard to tell what the issue is at a quick glance.

The current Enum has some of the more common assertions I've seen, and maybe a few from just DAA that seemed like they are likely in other sets too. I also added them to DAA for testing/examples.

There are a couple different ways to use the Enum, and I try to demonstrate several of them throughout DAA, though I think the two preferred methods should probably be `AssertString.ENUM_VALUE.text` for using the supplied string (though I could see an argument for using `AssertString.getAssertString "my.string"`)
Ex: `AssertString.MY_DECK.text`
Or `AssertString.getAssertString "my.deck"`
> "Your deck is empty"

and `AssertString.getAssertString "string.id", "Additional information to append"`  for providing additional information to the supplied string
Ex: `AssertString.getAssertString "my.discard.filter" "Basic Energy"`
> "Not in discard pile: Basic Energy"

If someone has an idea for a better implementation though, I'm all ears.

### DIFF
--- a/src/tcgwars/logic/impl/gen8/DarknessAblaze.groovy
+++ b/src/tcgwars/logic/impl/gen8/DarknessAblaze.groovy
@@ -282,6 +282,7 @@ public enum DarknessAblaze implements LogicCardInfo {
     return name();
   }
 
+//TODO: Remove empty closures
   @Override
   public Card getImplementation() {
     switch (this) {
@@ -535,7 +536,7 @@ public enum DarknessAblaze implements LogicCardInfo {
           text "Search your deck for a [G] Energy card and attach it to 1 of your Pokémon."
           energyCost C
           attackRequirement {
-            assert my.deck : "Your deck is empty"
+            assert my.deck : AssertString.MY_DECK.text
           }
           onAttack {
             attachEnergyFrom type:G, my.deck, my.all
@@ -581,7 +582,7 @@ public enum DarknessAblaze implements LogicCardInfo {
           text "Choose 1 of your opponent’s Pokémon. This attack does 20 damage to that Pokémon for each Energy attached to this Pokémon. (Don't apply Weakness and Resistance for Benched Pokémon.)"
           energyCost C
           attackRequirement {
-            assert self.cards.energyCount(C) : "$self.name has no Energy attached"
+            assert self.cards.energyCount(C) : AssertString.getAssertString 'energy.energyCount', "[C]"
           }
           onAttack {
             def damageAmount = 20 * self.cards.energyCount(C)
@@ -821,7 +822,7 @@ public enum DarknessAblaze implements LogicCardInfo {
           text "Draw a card."
           energyCost C
           attackRequirement {
-            assert my.deck : "Your deck is empty"
+            assert my.deck : AssertString.MY_DECK.text
           }
           onAttack {
             draw 1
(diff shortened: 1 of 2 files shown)


flagrama:
Was testing in the Groovy Compiler in the same file, but since they are in different files here I'll probably need it in a class to be able to import it?

flagrama:
I think a combination of the two could work very well too. This lays the groundwork for that being possible. You'll notice I didn't do anything for the bench related stuff in DAA because your methods do exist, however I do have one major issue with them at present, which would likely be solved by using this Enum. You do manipulations to the string that very likely only work for the English language, which makes localization somewhere between painful, to impossible, and this is very much something players would likely appreciate in the far future.

starg09:
> You do manipulations to the string that very likely only work for the English language, which makes localization somewhere between painful, to impossible, and this is very much something players would likely appreciate in the far future.

Yup, originally handled it more locally but rolled it back for simplicity. An enum handling it (and translating) could certainly be more ideal 👍

axpendix:
I believe this is definitely a good start for the problem, but its scope is bigger than this. 

I think the new method is way too verbose and can be improved. 

How about adding new assert methods to `CardList`, `BenchSet` and so on, like this:

`my.deck.assertNotEmpty()` 
`my.discard.assertHasType(ENERGY)`

flagrama:
I do agree something less verbose for normal usage sounds good. However, I still like the idea of centralizing common strings. What if something like this Enum still existed, but was used by said assertion methods. Something like 
```groovy
// DeckCardList.groovy?
def assertNotEmpty() {
  // Should probably have some way to display whose deck is empty, but for the purposes of example...
  assert this.notEmpty() : AssertString.getAssertString "my.deck"
}
```
which, in requirement closures, would still result in us writing
```groovy
attackRequirements {
  my.deck.assertNotEmpty()
}
```

Being able to reference all the possible assertion strings in a single file may help finding the relevant assertion methods too since you could just do a search on the message's key to find the assertion method that uses it, which you can then use in the card you're working on.

axpendix:
That could be useful if we ever want to support other languages than English. 

I couldn't see how those message keys can be useful for searching the codebase, wouldn't you search for `assertNotEmpty()` calls instead?

axpendix:
I fully appreciate the effort and concern to have unified assert statements of course and I don't want to be a blocker for improvements, but IMO care should be given to improve the readability and implementation efficiency. These method calls achieve neither at current stage. 

I am fine with adding an enum to resolve the messages, but then we will only be able to apply that enum to **some** of the messages, as there are a ton of messages out there which won't be affected. If I am not mistaken, assuming all calls are unified to individual `assertX()` methods then all messages will be written inside those methods anyway, which also kind of makes the enum irrelevant (abstractions which have a single use case are unnecessary), unless we need to support different languages (it's a much bigger task than assert statements).

flagrama:
Yeah, I see your point. This is probably over-engineering for a simple problem that has a simpler solution. I do think that care should be taken to not over-localize the messages within the methods for English if additional languages is something that is intended/desired in the future, even if it is the far future. It's not fun to correct that mistake later.

axpendix:
Closing for now

PR #857, opened 2020-06-29
Extra Energy Bomb knocks out Electrode before effect

Grayahm:


starg09:
While this is probably the better solution for now. I think prize taking should be delayed until after the effect is finished. Prize taking should never happen until all what's stated in an ability/power/body is finished. 

There are rulings regarding the electrode discard, saying you can attach the energy discarded with it, so current implementation is ideal in most cases. But in a situation where the opponent knows what its prizes are ("Here Comes Team Rocket!" from TRR), picking the prizes before the ability effect is finished could have a negative impact on said player.

@axpendix is there some method to stop and queue (not prevent) prize-taking events someone receives, and execute them only when another method re-enables them? Discard should still happen. (Alternatively, since it's a KO by its own ability and there's no modifier for prizes for something like that, that I'm aware of; it could prevent the prizes and just give 2 prizes at the end)

Grayahm:
Here Comes Team Rocket! is the only card I know of in the EX format that looks at prizes, and it's unplayable. I agree that the effect should take place before the knockout but I am 99% sure it will never be relevant. I think an easy fix could be discarding all energy from Electrode ex before checking the discard pile for energy to attach.

starg09:
Yup, it's a niche fix at best; not really a priority :P

That could be an option, and assuming what I'm working on for another card functions correctly, it could also allow to select a new active as if electrode was knocked out already. Might test with it more, but once the update is pushed (current behavior here is more than enough!)

PR #860, opened 2020-06-29
Fixed conditional switches from bench to active

starg09:
Many (but not all) attacks, abilities and poké-powers/bodies trigger a switch effect that's applied to the benched Pokémon. `sw` on itself is triggered for both switched pokemon as an ATTACK/TRAINER_CARD/SRC_ABILITY effect **(Update: This may not be accurate, tho a change is still needed.)**. ~~Using [swFromBench](https://github.com/axpendix/tcgone-engine-contrib/blob/0b6c994535cdb5fa4d44ea5d42ddfa0ba6ab21d1/src/tcgwars/logic/groovy/TcgStatics.groovy#L442) handles the targeting part on itself, and also provides a boolean return for when a following effect depends on the switch being successful or not (Drag Off and Guzma being prime examples).~~

**Update: Switched to a `sw2` method after talking with @axpendix, as sw already returns a boolean. `sw2` will return false if the switch is prevented, and also make the proper `Switch` call so the correct pcs is targeted (while making some switch lines shorter and clearer).**

This change makes trainers work properly against Omega Barrier or Unnerve Pokémon. Plus, attacks that depend on the switch in order to damage the new Pokémon properly do so (not affecting the active when not switched) in a cleaner way.

### DIFF
--- a/src/tcgwars/logic/impl/gen3/Deoxys.groovy
+++ b/src/tcgwars/logic/impl/gen3/Deoxys.groovy
@@ -2880,11 +2880,11 @@ public enum Deoxys implements LogicCardInfo {
             energyCost C, C
             onAttack {
               def pcs = defending
-              if(opp.bench && confirm("Switch 1 of your opponent's Benched Pokémon with the Defending Pokémon?")){
+              if (opp.bench && confirm("Switch 1 of your opponent's Benched Pokémon with the Defending Pokémon?")){
                 def target = opp.bench.select("Select the new Active Pokémon.")
                 if ( sw2(target) ) { pcs = target }
               }
-              damage 20, pcs
+              damage 20
             }
           }
           move "Darkness Blast", {
--- a/src/tcgwars/logic/impl/gen3/Deoxys.groovy
+++ b/src/tcgwars/logic/impl/gen3/Deoxys.groovy
@@ -1389,7 +1389,7 @@ public enum Deoxys implements LogicCardInfo {
               assert opp.bench : "There is no Pokémon to switch"
               powerUsed()
               flip {
-                sw2(opp.bench.select("Select your opponent's new Active Pokémon."), SRC_ABILITY)
+                switchYourOpponentsBenchedWithActive(SRC_ABILITY)
               }
             }
           }
--- a/src/tcgwars/logic/impl/gen3/Deoxys.groovy
+++ b/src/tcgwars/logic/impl/gen3/Deoxys.groovy
@@ -1389,7 +1389,7 @@ public enum Deoxys implements LogicCardInfo {
               assert opp.bench : "There is no Pokémon to switch"
               powerUsed()
               flip {
-                sw opp.active, opp.bench.select("Select the new active Pokémon.")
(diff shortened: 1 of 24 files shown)


axpendix:
Thanks for submittal. I will add my review soon.

starg09:
Edited OP with info, but this method should do for a better solution.

It depends on a backend update though, so please do not merge this PR until that's done!

src/tcgwars/logic/impl/gen3/Deoxys.groovy
```diff
@@ -2880,13 +2880,10 @@ public enum Deoxys implements LogicCardInfo {
...
-                if(confirm("Switch 1 of your opponent’s Benched Pokémon with the Defending Pokémon before doing damage?")){
-                  pcs = opp.bench.select()
-                  sw defending, pcs
-                }
+              if(opp.bench && confirm("Switch 1 of your opponent's Benched Pokémon with the Defending Pokémon?")){
+                def target = opp.bench.select("Select the new Active Pokémon.")
+                if ( sw2(target) ) { pcs = target }
               }
-
               damage 20, pcs
```
axpendix, line 2887: actually this assignment logic is not necessary `damage 20` will always register the damage to the current active (defending) pokemon

starg09: That sounds good; didn't even think about it after the initial tests with roselia (since first bug was it applying to the unswitched active). Will correct this in a bit.

starg09: (Never mind that, mixed this with another card when reading the email; still fixing it tho, just different reason :P)

src/tcgwars/logic/impl/gen3/UnseenForces.groovy
```diff
@@ -3555,13 +3555,14 @@ public enum UnseenForces implements LogicCardInfo {
...
+            assert opp.bench : "Your opponent has no Benched Pokémon"
+          }
           onAttack {
-            def target = defending
-            if (opp.bench) {
-              target = opp.bench.select("Select the new active")
-              sw defending, target
-              apply BURNED
-              apply CONFUSED
+            def target = opp.bench.select("Select the new Active Pokémon.")
```
axpendix, line 3562: "Your opponent chooses the Defending Pokémon to switch."

src/tcgwars/logic/impl/gen7/BurningShadows.groovy
```diff
@@ -2655,12 +2656,9 @@ public enum BurningShadows implements LogicCardInfo {
...
           onPlay {
-            def pcs = opp.bench.select("New active")
-            targeted (pcs, TRAINER_CARD) {
-              sw opp.active, pcs, TRAINER_CARD
-              if(my.bench) {
-                sw my.active, my.bench.select("New active"), TRAINER_CARD
-              }
+            def target = opp.bench.select("Select the new Active Pokémon.")
+            if ( sw2(target, TRAINER_CARD) && my.bench) {
+              sw my.active, my.bench.select("Select your new Active Pokémon."), TRAINER_CARD
```
axpendix, line 2661: 👍

PR #861, opened 2020-06-29
Remove period in ROCKET_S_ADMIN__86

luongthomas:
Screen Shot 2020-06-29 at 1 09 45 PM

This will remove the extra period in the name that shows up when playing the card


starg09:
Might send a revert for this. The name does include the dot, hadn't noticed at first.

Bit annoying on log though, but this causes deck import to fail if the list uses the dot (opposite case would still show the card)

luongthomas:
> Might send a revert for this. The name does include the dot, hadn't noticed at first.
> 
> Bit annoying on log though, but this causes deck import to fail if the list uses the dot (opposite case would still show the card)

I believe you’re right, we’ll need to re-add the period to the name.  I didn’t realize the dot was part of the actual name.

PR #864, opened 2020-07-01
"EX Legend Maker" fixes

starg09:


src/tcgwars/logic/impl/gen3/LegendMaker.groovy
```diff
@@ -393,6 +401,7 @@ public enum LegendMaker implements LogicCardInfo {
           }
           onAttack {
             attachEnergyFrom(type : F, my.discard, self)
+            //TODO: Confirm the energy being attached maybe?
```
flagrama, line 404: Yeah, for attacks/abilities that can copy this attack, should probably put the `attackRequirement` in an if as well and contain these in it. At least from my understanding from what @ufodynasty has said.

src/tcgwars/logic/impl/gen3/LegendMaker.groovy
```diff
@@ -775,12 +784,12 @@ public enum LegendMaker implements LogicCardInfo {
             checkLastTurn()
             checkNoSPC()
             assert my.deck : "Deck is empty"
-            assert self.cards.findAll { it.name.contains("React Energy") }.size() == 0 : "Huntail has React Energy cards attached"
+            assert !(self.cards.any{ it.name == "React Energy" }) : "Huntail has React Energy cards attached"
             powerUsed()
 
-            def energy = my.deck.search(max: 1, "Select a React Energy card.", {it.name.contains("React Energy")}).first()
+            def energy = my.deck.search(max: 1, "Select a React Energy card.", { it.name == "React Energy" })
```
flagrama, line 790: Shuffle the deck

src/tcgwars/logic/impl/gen3/LegendMaker.groovy
```diff
@@ -1145,17 +1156,21 @@ public enum LegendMaker implements LogicCardInfo {
           text "Put 7 damage counters on the Defending Pokémon at the end of your opponent's next turn."
           energyCost P, P, C
           onAttack {
-            delayed {
-              before BETWEEN_TURNS, {
-                if (bg.currentTurn == self.owner.opposite) {
-                  directDamage 70, self.owner.opposite.pbg.active
-                  bc "Shadow Tag activates"
+            targeted (defending){
+              bc "7 damage counters will be put on ${self.owner.opposite}'s Defending ${defending} at the end of their next turn. (This effect can be removed by evolving or benching ${defending}.)"
```
flagrama, line 1160: `${opp.owner.getPlayerUsername(bg)}` may be more user-friendly

src/tcgwars/logic/impl/gen3/LegendMaker.groovy
```diff
@@ -1512,10 +1528,11 @@ public enum LegendMaker implements LogicCardInfo {
...
           text "30 damage. If the Defending Pokémon is not Asleep, this attack does nothing."
           energyCost P
+          attackRequirement{
+            assert defending.isSPC(ASLEEP) : "The Defending Pokémon is not asleep"
+          }
           onAttack {
-            if (defending.isSPC(ASLEEP)) {
-              damage 30
-            }
+            damage 30
```
flagrama, line 1535: The if statement is still required for copying this attack AFAIK.

src/tcgwars/logic/impl/gen3/LegendMaker.groovy
```diff
@@ -1534,6 +1551,7 @@ public enum LegendMaker implements LogicCardInfo {
           text "30 damage. Before doing damage, discard all Trainer cards attached to the Defending Pokémon."
           energyCost D, C
           onAttack {
+            //TODO: Check this not discarding Fossil placed as Pokémon.
```
flagrama, line 1554: Fossils should be changed implementation to Pokémon and should not be affected by this.

src/tcgwars/logic/impl/gen3/LegendMaker.groovy
```diff
@@ -1547,10 +1565,10 @@ public enum LegendMaker implements LogicCardInfo {
         pokeBody "Reactive Aroma", {
           text "As long as Roselia has any React Energy cards attached to it, remove 1 damage counter from each of your Pokémon (excluding Pokémon-ex) that has any React Energy cards attached to it between turns. You can't use more than 1 Reactive Aroma Poké-Body each turn."
           delayedA {
-            before BETWEEN_TURNS, {
-              if (self.cards.findAll{it.name.contains("React Energy")}) {
+            before BEGIN_TURN, {
+              if (self.cards.any{it.name == "React Energy"}) {
```
flagrama, line 1569: I think this may need to be checked that it is only applying one of this ability each turn using a storedObject.

src/tcgwars/logic/impl/gen3/LegendMaker.groovy
```diff
@@ -1604,9 +1620,9 @@ public enum LegendMaker implements LogicCardInfo {
           text "Whenever you attach a React Energy card from your hand to Tangela, remove all damage counters from Tangela."
           delayedA {
             after ATTACH_ENERGY, self, {
-              if (ef.reason==PLAY_FROM_HAND && ef.card.name.contains("React Energy")) {
+              if (ef.reason==PLAY_FROM_HAND && ef.card.name == "React Energy") {
```
flagrama, line 1623: I don't think this will work right. You may need to check that the card is in hand in a `before PLAY_ENERGY` to ensure it works properly in all cases. (See [Crazy Code Porygon-Z](https://tcgone.net/cards/porygon-z-unb-157) ability which should still activate this effect when used)

src/tcgwars/logic/impl/gen3/LegendMaker.groovy
```diff
@@ -1683,7 +1702,7 @@ public enum LegendMaker implements LogicCardInfo {
           energyCost C, C
           onAttack {
             damage 20
-            if (self.cards.findAll { it.name.contains("React Energy") }) {
+            if (self.cards.any{ it.name == "React Energy" }) {
```
flagrama, line 1705: I think this is the proper way to check, and may need applied to some of the other React Energy checks.

src/tcgwars/logic/impl/gen3/LegendMaker.groovy
```diff
@@ -2534,13 +2545,12 @@ public enum LegendMaker implements LogicCardInfo {
           text "Whenever you attach a [R] Energy from your hand to Arcanine ex, remove 1 damage counter and all Special Conditions from Arcanine ex."
           delayedA {
             after ATTACH_ENERGY, self, {
-              if (ef.reason==PLAY_FROM_HAND && ef.card.asEnergyCard().containsType(R) && confirm("Use Fire Remedy?")) {
-                bc "Fire Remedy heals 10 and removes all Special Conditions from Arcanine ex"
+              if (ef.reason==PLAY_FROM_HAND && ef.card.asEnergyCard().containsType(R)) {
```
flagrama, line 2548: See my comment on Tangela

src/tcgwars/logic/impl/gen3/LegendMaker.groovy
```diff
@@ -262,11 +262,12 @@ public enum LegendMaker implements LogicCardInfo {
           text "Count the number of React Energy cards attached to Cradily and choose up to that number of your opponent's Evolved Pokémon. Remove the highest Stage Evolution card from each of those Pokémon, then have your opponent shuffle those cards into his or her deck."
           energyCost C, C
           attackRequirement {
-            assert self.cards.findAll {it.name.contains("React Energy")} : "No React Energies attached to this Pokemon"
-            assert opp.all.findAll{ it.evolution } : "Your opponent has no evolved pokemon"
+            assert self.cards.any{it.name.contains("React Energy")} : "No React Energies attached to this Pokemon"
+            assert opp.all.any{ it.evolution } : "Your opponent has no evolved pokemon"
           }
           onAttack {
```
flagrama, line 268: Looks like this is moving the cards to the hand? Should be shuffled into deck according to attack text.

src/tcgwars/logic/impl/gen3/LegendMaker.groovy
starg09: You mean as in duplicating it? Having both the assert and do an if? Just to confirm.

src/tcgwars/logic/impl/gen3/LegendMaker.groovy
starg09: Ok, I may have been removing a couple of these if thinking of them as unneeded. Will avoid doing so in the future, thanks for the correction 👌

src/tcgwars/logic/impl/gen3/LegendMaker.groovy
starg09: Yup, good catch!

src/tcgwars/logic/impl/gen3/LegendMaker.groovy
starg09: Crazy code seems to attach as a PLAY_FROM_HAND, doesn't it? Tested it with Combusken from Dragon Majesty and it removed SPC no problem. Some cards do need improved checks tho, as in not asking for basic energies but that's for another commit/PR.

src/tcgwars/logic/impl/gen3/LegendMaker.groovy
starg09: Thanks for the reminder, just corrected all missing instances of it 👍

src/tcgwars/logic/impl/gen3/LegendMaker.groovy
starg09: Yup, left this one as a TODO before leaving for bed a couple days ago. Will look at it in the morning, should do a full rewrite of it.

src/tcgwars/logic/impl/gen3/LegendMaker.groovy
starg09: Gonna leave the TODO in here, as I still plan on working out a check for "energy got attached" check at some point. Added the double conditional back though.

src/tcgwars/logic/impl/gen3/LegendMaker.groovy
flagrama: Hmm, my understanding was that wasn't working right on Blastoise ex, but it looks like that wasn't using `attachEnergy` but was instead using `attachEnergyFrom`.

src/tcgwars/logic/impl/gen3/LegendMaker.groovy
starg09: This one I wanted to be able to test it, so is already commited in master: https://github.com/axpendix/tcgone-engine-contrib/compare/d087369a3bddc17ec62b2189feac1eff4339f2f7...2177bdbc0453d9cf1c972ba125fd4816738e49bf

Now shuffles to deck and is more user-friendly regarding selection.

flagrama: You forgot to shuffle the deck in that commit it seems, but aside from that, all looks good to me now.

PR #865, opened 2020-07-03
Changes to the Limitation/Bother static + Ignore Abilities/Attacks replicating supporters in Trainer/Sup Lock

starg09:
This makes it so it uses the original move's name (so Limitation and Ghastly Sound can make use of it).

It also removes the unregister conditions. From what I see, neither of those attacks nor Bother should be removed if the attacking Pokémon is switched or evolved. Putting this as a PR since it touches a static method, not sure if it may affect some legacy card I'm not aware of.


src/tcgwars/logic/impl/gen4/MysteriousTreasures.groovy
```diff
@@ -2475,8 +2473,8 @@ public enum MysteriousTreasures implements LogicCardInfo {
             attackRequirement {}
             onAttack {
               flip {
-                opponentCantPlaySupporterNextTurn(delegate) //TODO: Que el static use ${thisMove.name}
-                bc "${thisMove.name} - Supporters can't be played from the turn player's hand during the next turn" //TODO: Find way to use "Player A/B" as a $variable
+                opponentCantPlaySupporterNextTurn(delegate)
+                bc "${thisMove.name} - Supporters can't be played from the turn player's hand during the next turn" //TODO: Move this to the static.
```
flagrama, line 2477: Shouldn't this already be done with the changes to the static in this PR?

src/tcgwars/logic/impl/gen3/LegendMaker.groovy
```diff
@@ -2248,15 +2248,23 @@ public enum LegendMaker implements LogicCardInfo {
           }
           eff2 = delayed {
             boolean flag = false
+            def flaggedPokemon
```
flagrama, line 2251: You should be able to use the single flag for both EVOLVE and DEVOLVE effects since they won't happen at the same time.

src/tcgwars/logic/groovy/TcgStatics.groovy
```diff
@@ -1299,13 +1299,11 @@ class TcgStatics {
     _delegate.delayed {
```
flagrama, line 1299: Does this static work correctly if a supporter is not used from hand, such as being used via an attack for example? The Supporter should work if it is played as the effect of an attack, played from the discard (don't know if an effect like this exists) or if you play the effect of a Supporter from the opponent's hand, right?

src/tcgwars/logic/impl/gen4/MysteriousTreasures.groovy
```diff
@@ -2547,16 +2545,15 @@ public enum MysteriousTreasures implements LogicCardInfo {
               damage 10
               afterDamage {
                 flip{
+                  bc "${opp.owner.getPlayerUsername(bg)} can't attach any Energy cards from their hand to their Active Pokémon during their next turn. (This effect is on the player, it isn't removed by benching or evolving the currently Active ${defending}.)"
```
flagrama, line 2548: My comment from #866 applies for this too.

src/tcgwars/logic/impl/gen3/LegendMaker.groovy
starg09: Not the same kind. for evolve you can use `ef.pokemonToBeEvolved`. In devolved a workaround is used, with a variable pointing to the PCSs and checking those not the same. It's too many checks for this one (and flaggedPokemon should be removed once a way to point to "pokemonToBeDevolved" is known/added).

starg09: (PS: Didn't notice I commited this under this PR, was meant to be a direct commit.)

src/tcgwars/logic/impl/gen4/MysteriousTreasures.groovy
starg09: Yeah, could do it already. Will commit it in a bit 👍

src/tcgwars/logic/impl/gen3/LegendMaker.groovy
flagrama: Oh, my bad, didn't realize it was holding a PCS in it.

src/tcgwars/logic/groovy/TcgStatics.groovy
starg09: Didn't check for this, the static already existed as this. Will check it in a bit.

starg09: Thinking about it, it could reduce the "max supporters from hand per turn" limit. Though it'd still cause some weird interactions with Lt. Surge's Strategy.

src/tcgwars/logic/groovy/TcgStatics.groovy
```diff
@@ -1296,16 +1296,15 @@ class TcgStatics {
     }
   }
   static void opponentCantPlaySupporterNextTurn(def _delegate){
+    bc "${_delegate.thisMove} - Supporters can't be played from ${opp.owner.getPlayerUsername(bg)}'s hand during theit next turn"
```
axpendix, line 1299:
```suggestion
    bc "${_delegate.thisMove} - Supporters can't be played from ${opp.owner.getPlayerUsername(bg)}'s hand during their next turn"
```

src/tcgwars/logic/groovy/TcgStatics.groovy
```diff
@@ -1296,16 +1296,15 @@ class TcgStatics {
...
       before PLAY_TRAINER, {
         if(ef.supporter && bg.currentTurn==_delegate.self.owner.opposite) {
-          wcu "Bother prevents playing supporters"
+          wcu "${_delegate.thisMove} prevents playing supporters"
           prevent()
         }
       }
       unregisterAfter 2
-      after SWITCH, _delegate.self, {unregister()}
-      after EVOLVE, _delegate.self, {unregister()}
```
axpendix, line 1308: Why did you remove these?

starg09: From the PR's OP:
> It also removes the unregister conditions. From what I see, neither of those attacks nor Bother should be removed if the attacking Pokémon is switched or evolved. Putting this as a PR since it touches a static method, not sure if it may affect some legacy card I'm not aware of.

Main reason I put it as a PR, besides touching a static. Is there some card that needs this when calling this static? None of the ones in the repo seem to.

flagrama: The way the relevant attacks are worded, these should effects on the Player, not the Pokémon. Switching and evolving should have nothing to do with the effect.

axpendix:
I am fine if you're sure with the rules :) I would also recommend to search the inbox for those cards and delete/reply if appropriate.

starg09:
That sounds good. Will wait for merging this until I have some time to properly test all potentially involved cards.

src/tcgwars/logic/groovy/TcgStatics.groovy
starg09: > Does this static work correctly if a supporter is not used from hand, such as being used via an attack for example?

Finally got to test this a bit. And sadly, I don't see a clean way to solve this as of now. Every single attack preventing the play of trainers (or supporters in specific) currently prevent the effect from being done. But in my opinion a change needs to be done on the side of the attack/ability replicating the supporter.

Currently, they're discarded and played as themselves (being considered a trainer effect, or whatever indicated in the statics of the trainer cards themselves); when in reality only the effect (and needed checks/costs ofc) should be used, after being converted into an attack-sourced effect. A mimikyu impersonating guzma should be able to target a Omega Barrier Pokémon and switch it without trouble, unlike the regular supporter.

Will merge this for now, as the commited fixes are still relevant. And added a flag so the supporters are played at least. But it's still not a perfect fix.

starg09: Cards that I will not add a flag check for in this commit everywhere, but may need it:
* Smeargle (Undaunted, Call of Legends)
* Glalie (Mysterious Treasures)

starg09:
Just pushed a commit here for ignoring impersonation-like use of supporters. [This however won't truly solve it, a bigger fix is needed.](https://github.com/axpendix/tcgone-engine-contrib/pull/865#discussion_r450988936)

Asking for review again, just to confirm this is fine.

luongthomas:
This builds locally fine

axpendix:
Hi @starg09 what are we missing on this one?

starg09:
> Hi @starg09 what are we missing on this one?

Mostly testing; I can get to that today since it's easier to debug now, were anything odd to pop out 🙂 

[There's still this situation](https://github.com/axpendix/tcgone-engine-contrib/pull/865#discussion_r450988936) though, but it's part of a source refactor that needs to be done in engine so not much to add in this PR regarding that.

starg09:
All seems to work fine after testing locally (and applying some minor fixes).

PR #866, opened 2020-07-03
Multiple card fixes, revolving around adding DEVOLVE checks (but more stuff on a couple separate commits)

starg09:
First commit is pretty straightforward, just adds `after DEVOLVE` unregisters. I think they're needed in all modified cards but let me know if you think some shouldn't.

Regarding other commits: One situation I noticed on some cards is the effect remaining on the new active on certain situations. While maybe unnecessary, setting a pcs variable pointing to the specific... well, PCS, should make it work more reliably in theory, at least in my opinion. Also includes some changes in some implementations, beyond the DEVOLVE check above.

### DIFF
--- a/src/tcgwars/logic/groovy/TcgStatics.groovy
+++ b/src/tcgwars/logic/groovy/TcgStatics.groovy
@@ -429,6 +429,7 @@ class TcgStatics {
         unregisterAfter 2
         after SWITCH, defending, {unregister()}
         after EVOLVE, defending, {unregister()}
+        after DEVOLVE, defending, {unregister()}
       }
     } }
   }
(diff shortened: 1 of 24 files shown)


flagrama:
All of these "delayed effect has been set" texts seem quite verbose. Since the card itself has most of the text, maybe it'd be better to just mention that the effect has been set? While I think mentioning such could be useful, I don't think you need to have almost all of the text from the card itself included.

starg09:
@flagrama While for most effects I'd agree, I think this kind of delayed effects should be as clear as possible. They're not as easy to follow, and a couple vary between cards between affecting the opponent, the opponent's defending Pokémon, or another variation. Ideally, at some point a list of effects can be stored and this verbose can be located there instead of at the log. But for now, I think it should be printed for these special cases.

PR #873, opened 2020-07-09
"EX Crystal Guardians" Fixes

starg09:


src/tcgwars/logic/impl/gen3/CrystalGuardians.groovy
```diff
@@ -1904,7 +1936,7 @@ public enum CrystalGuardians implements LogicCardInfo {
           }
         }
         playRequirement{
-          assert all.findAll {it.cards.hasType(POKEMON_TOOL)} || (bg.stadiumInfoStruct && stadiumCanBeAffectedByItemAndSupporter())
+          assert all.any{it.cards.hasType(POKEMON_TOOL)} || (bg.stadiumInfoStruct && stadiumCanBeAffectedByItemAndSupporter())
```
flagrama, line 1939: Is this trainer at all different than field blower? If not, one should probably copy the other.

src/tcgwars/logic/impl/gen3/CrystalGuardians.groovy
```diff
@@ -779,6 +786,14 @@ public enum CrystalGuardians implements LogicCardInfo {
...
         weakness F
+        pokeBody "Hover Lift", {
+          text "You pay [C] less to retreat your Jigglypuff, Wigglytuff, Wigglytuff ex, and Igglybuff."
+          getterA GET_RETREAT_COST ,{ h->
+            if (h.effect.target.owner == self.owner && ["Jigglypuff", "Wigglytuff", "Wigglytuff ex", "Igglybuff"].contains(h.effect.target.name)) {
+              h.object -= 1
+            }
+          }
+        }
         pokePower "Baby Evolution", {
```
flagrama, line 797: Just making a note here that this Poké-Power isn't on the carddb and should probably be added at some point.

src/tcgwars/logic/impl/gen3/CrystalGuardians.groovy
```diff
@@ -1916,21 +1948,22 @@ public enum CrystalGuardians implements LogicCardInfo {
       case AGGRON_EX_89:
       return evolution (this, from:"Lairon", hp:HP150, type:M, retreatCost:4) {
         weakness R
+        weakness F
```
flagrama, line 1951: Also missing from carddb

src/tcgwars/logic/impl/gen3/CrystalGuardians.groovy
```diff
@@ -2174,14 +2211,17 @@ public enum CrystalGuardians implements LogicCardInfo {
...
+            afterDamage{
+              discardSelfEnergy C,C
+            }
           }
         }
       };
       case SCEPTILE_EX_DELTA_96:
       return evolution (this, from:"Grovyle", hp:HP140, type:P, retreatCost:1) {
         weakness G
+        weakness R
```
flagrama, line 2224: also missing from carddb

src/tcgwars/logic/impl/gen3/CrystalGuardians.groovy
starg09: Looking at them, they should be functionally identical. Will use copy on this 👍

PR #874, opened 2020-07-10
Make opponent's hand to appear randomized to the player when needed

starg09:
Fixes potential exploits where a player may use a hand-seeing power, then one that is meant to be random or hidden from their view. Without randomization they can predict which card to discard specifically, breaking many decks in the process.

Work on this must be done for legacy sets ASAP in my opinion. But this should solve a good amount of those interaction, as long as at least one groovy-coded card is involved.

### DIFF
--- a/src/tcgwars/logic/impl/gen7/ForbiddenLight.groovy
+++ b/src/tcgwars/logic/impl/gen7/ForbiddenLight.groovy
@@ -2159,17 +2159,18 @@ public enum ForbiddenLight implements LogicCardInfo {
             }
             onAttack {
               if (opp.hand) {
-              def randomOppHand = opp.hand.shuffledCopy()
-              if(randomOppHand.hasType(SUPPORTER)){
-                def support = randomOppHand.select(max: 0, "Your opponent's hand. You may discard a Supporter card you find there and use the effect of that card as the effect of this attack.", cardTypeFilter(SUPPORTER))
-                if (support){
-                  discard support.first()
-                  bg.deterministicCurrentThreadPlayerType=self.owner
-                  bg.em().run(new PlayTrainer(support.first()))
-                  bg.clearDeterministicCurrentThreadPlayerType()
+                def randomOppHand = opp.hand.shuffledCopy()
+                if(randomOppHand.hasType(SUPPORTER)){
+                  def support = randomOppHand.select(max: 0, "Your opponent's hand. You may discard a Supporter card you find there and use the effect of that card as the effect of this attack.", cardTypeFilter(SUPPORTER))
+                  if (support){
+                    discard support.first()
+                    bg.deterministicCurrentThreadPlayerType=self.owner
+                    bg.em().run(new PlayTrainer(support.first()))
+                    bg.clearDeterministicCurrentThreadPlayerType()
+                  }
+                } else {
+                  randomOppHand.showToMe("Your opponent's hand. No supporter in there.")
                 }
-              } else {
-                randomOppHand.showToMe("Your opponent's hand. No supporter in there.")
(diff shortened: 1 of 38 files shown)


src/tcgwars/logic/impl/gen7/TeamUp.groovy
```diff
@@ -1423,10 +1423,13 @@ public enum TeamUp implements LogicCardInfo {
...
             energyCost P,P
+            attackRequirement {
+              assert opp.hand : "Your opponent has no cards in their hand"
+            }
             onAttack{
-              if(opp.hand){
-                opp.hand.showToMe("Your opponent's hand")
-                damage 50*opp.hand.filterByType(TRAINER).size()
+              if (opp.hand){
+                trainersInHand = opp.hand.shuffledCopy().showToMe("Your opponent's hand").filterByType(TRAINER)
```
luongthomas, line 1431: `trainersInHand` isn't defined here yet

src/tcgwars/logic/impl/gen7/TeamUp.groovy
```diff
@@ -3076,11 +3082,11 @@ public enum TeamUp implements LogicCardInfo {
...
               afterDamage{
                 if(opp.hand.size() >= 4){
-                  if(opp.hand.size() == 4){
-                    opp.hand.showToMe("Your opponent's hand")
-                  }
-                  else{
-                    opp.hand.select(count:opp.hand.size() - 4,"Choose the cards to discard").discard()
+                  def randomOppHand = opp.hand.shuffledCopy()
+                  if (opp.hand.size() == 4){
+                    randomOppHand.shuffledCopy().showToMe("Your opponent's hand.")
```
luongthomas, line 3087: is the second `shuffledCopy()` operation needed here?

src/tcgwars/logic/util/CardList.groovy
```diff
@@ -61,6 +61,11 @@ public class CardList extends ArrayList<Card> {
   public CardList copy(){
     return new CardList(this);
   }
+  public CardList shuffledCopy(){
```
luongthomas, line 64: This is an extra shuffledCopy(), there is another function already defined in line 70

src/tcgwars/logic/impl/gen7/TeamUp.groovy
starg09: Nope, leftover from before starting to use a variable. Nice catch 👍

src/tcgwars/logic/util/CardList.groovy
starg09: Welp, guess that's one thing to lookout when rebase-merging these PRs (tested a card in master with the method)

src/tcgwars/logic/impl/gen7/ForbiddenLight.groovy
```diff
@@ -2158,14 +2158,18 @@ public enum ForbiddenLight implements LogicCardInfo {
...
             }
             onAttack {
-              if(opp.hand.hasType(SUPPORTER)){
-                def card=opp.hand.select("Opponent's hand. Select a supporter.", cardTypeFilter(SUPPORTER)).first()
-                discard card
-                bg.deterministicCurrentThreadPlayerType=self.owner
-                bg.em().run(new PlayTrainer(card))
-                bg.clearDeterministicCurrentThreadPlayerType()
+              if (opp.hand) {
+              def randomOppHand = opp.hand.shuffledCopy()
```
luongthomas, line 2162: Could you indent this `if` block over by one?  Also it looks like it's missing a closing bracket

PR #875, opened 2020-07-11
"EX Dragon Frontiers" Fixes

starg09:


src/tcgwars/logic/impl/gen3/DragonFrontiers.groovy
```diff
@@ -490,7 +487,7 @@ public enum DragonFrontiers implements LogicCardInfo {
           delayedA {
             before APPLY_ATTACK_DAMAGES, {
               bg.dm().each {
-                if (it.to == self && it.dmg.value && it.notNoEffect && opp.hand.size() >= 5) {
+                if (it.to == self && it.dmg.value && it.notNoEffect && self.owner.opposite.pbg.hand.size() >= 5) {
```
flagrama, line 490: This seems unnecessary unless copying the attack doesn't work properly. You're basically foregoing the static to put the full non-static call.

starg09: This change was so it works for all attacks. Even your own damaging it on bench (since it never mentions attacks coming from the opponent). Keeping `opp` would make it check one's own hand during the opponent's turn, but `my` would have the same issue except during your own turn.

starg09: There is a possibility of a mistranslation though, reading it again. Maybe it's meant to block only opponent's attacks? Looked for a bit but found nothing yet, may look for the JP print of it.

flagrama: Makes sense then.

starg09: Ok yeah checking the text and same-era cards, seems to be translated properly 👍

PR #876, opened 2020-07-12
"EX Power Keepers" Fixes

starg09:
The Missing weakness is already pushed as a PR in carddb, as a note.

### DIFF
--- a/src/tcgwars/logic/impl/gen3/LegendMaker.groovy
+++ b/src/tcgwars/logic/impl/gen3/LegendMaker.groovy
@@ -1191,7 +1191,7 @@ public enum LegendMaker implements LogicCardInfo {
           }
           onAttack {
             def maxSpace = Math.min(my.bench.freeBenchCount, 2)
-            my.deck.search(min:0, max:maxSpace, "Search your deck for up to 2 cards named Omanyte, Kabuto, Aerodactyl, Lileep, or Anorith", {
+            my.deck.search(min:0, max:maxSpace, "Search your deck for ${maxSpace == 2 ? "up to 2 cards" : "a card"} named Omanyte, Kabuto, Aerodactyl, Lileep, or Anorith", {
               ["Omanyte", "Kabuto", "Aerodactyl", "Lileep", "Anorith"].contains(it.name)
             }).each {
               my.deck.remove(it)
--- a/src/tcgwars/logic/impl/gen3/PowerKeepers.groovy
+++ b/src/tcgwars/logic/impl/gen3/PowerKeepers.groovy
@@ -763,7 +763,7 @@ public enum PowerKeepers implements LogicCardInfo {
           text "If your opponent has any Evolved Pokémon in play, remove the highest Stage Evolution card from each of them and put those cards back into his or her hand."
           energyCost C
           attackRequirement {
-            assert opp.all.findAll { it.evolution } : "Opponent does not have any Evolved Pokemon in play."
+            assert opp.all.any{ it.evolution } : "Opponent does not have any Evolved Pokemon in play."
           }
           onAttack {
             opp.all.findAll { it.evolution }.each {
@@ -826,9 +826,9 @@ public enum PowerKeepers implements LogicCardInfo {
           text "If Phoebe's Stadium is in play, prevent all damage done to Sableye by attacks from your opponent's Pokémon-ex."
           delayedA {
             before APPLY_ATTACK_DAMAGES, {
-              bg.dm().each {
-                if (it.to == self && it.from.EX) {
-                  if (bg.stadiumInfoStruct && bg.stadiumInfoStruct.stadiumCard.name == "Phoebe's Stadium") {
+              if (ef.attacker.EX && bg.stadiumInfoStruct && bg.stadiumInfoStruct.stadiumCard.name == "Phoebe's Stadium") {
+                bg.dm().each {
+                  if (it.to == self && it.dmg.value && it.notNoEffect) {
                     bc "Synergy Effect prevents all damage"
                     it.dmg=hp(0)
                   }
@@ -844,10 +844,11 @@ public enum PowerKeepers implements LogicCardInfo {
             assert my.deck : "No cards in deck"
           }
           onAttack {
-            if (my.deck.size() < 2) {
-              draw 2
-            } else {
-              my.deck.subList(my.deck.size() - 2, my.deck.size()).moveTo(hidden:true, my.hand)
+            if (my.deck) {
+              def drawStart = Math.max(0, my.deck.size() - 2)
+              def cardsDrawn = my.deck.size() - drawStart
+              my.deck.subList(drawStart, my.deck.size()).moveTo(hidden:true, my.hand)
+              bc "${my.owner.getPlayerUsername(bg)} drew ${cardsDrawn} cards from the bottom of their deck."
             }
           }
         }
--- a/src/tcgwars/logic/impl/gen3/PowerKeepers.groovy
+++ b/src/tcgwars/logic/impl/gen3/PowerKeepers.groovy
@@ -1556,7 +1556,7 @@ public enum PowerKeepers implements LogicCardInfo {
           energyCost P
           onAttack {
             damage 10
-            flip { apply PARALYZED }
+            flip { applyAfterDamage PARALYZED }
           }
         }
         move "Cross Chop", {
@@ -1579,10 +1579,12 @@ public enum PowerKeepers implements LogicCardInfo {
             assert my.deck : "Deck is empty"
           }
           onAttack {
-            my.deck.search(min:0, max:2, "Search your deck for a card named Omanyte, Kabuto, Aerodactyl, Lileep, or Anorith", {it.name == "Omanyte" || it.name == "Kabuto" || it.name == "Aerodactyl" || it.name == "Lileep" || it.name == "Anorith"}).each {
+            def maxSpace = Math.min(my.bench.freeBenchCount, 2)
+            my.deck.search(min:0, max:maxSpace, "Search your deck for ${maxSpace == 2 ? "up to 2 cards" : "a card"} named Omanyte, Kabuto, Aerodactyl, Lileep, or Anorith", {
+              ["Omanyte", "Kabuto", "Aerodactyl", "Lileep", "Anorith"].contains(it.name)
+            }).each {
               my.deck.remove(it)
               benchPCS(it)
-              // TODO: Mark as basic
             }
             shuffleDeck()
           }
--- a/src/tcgwars/logic/impl/gen3/PowerKeepers.groovy
+++ b/src/tcgwars/logic/impl/gen3/PowerKeepers.groovy
@@ -1915,11 +1915,9 @@ public enum PowerKeepers implements LogicCardInfo {
           eff = delayed {
             before APPLY_SPECIAL_CONDITION, {
               def pcs=e.getTarget(bg)
-              if (pcs.types.contains(D)) {
-                if (ef.type == POISONED || ef.type == CONFUSED || ef.type == PARALYZED) {
-                  bc "Sidney's Stadium - [D] Pokemon can't be Asleep, Confused or Paralyzed."
-                  prevent()
-                }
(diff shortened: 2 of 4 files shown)


src/tcgwars/logic/impl/gen3/PowerKeepers.groovy
```diff
@@ -1858,11 +1915,9 @@ public enum PowerKeepers implements LogicCardInfo {
           eff = delayed {
             before APPLY_SPECIAL_CONDITION, {
               def pcs=e.getTarget(bg)
-              if (pcs.types.contains(D)) {
-                if (ef.type == POISONED || ef.type == CONFUSED || ef.type == PARALYZED) {
-                  bc "Sidney's Stadium - [D] Pokemon can't be Asleep, Confused or Paralyzed."
-                  prevent()
-                }
+              if ( pcs.types.contains(D) and [ASLEEP, CONFUSED, PARALYZED].contains(ef.type) ) {
```
luongthomas, line 1918: I think `and` should be changed to `&&`

luongthomas: Also the IDE says `ef.type` doesn't exist, however `ef.effectType` does exist if you want to use that

starg09: >and
_whoops_

will correct the ef.type in a bit

starg09: and weird on ef.type, a ton of cards seem to use it? But I'll apply the change

src/tcgwars/logic/impl/gen3/LegendMaker.groovy
```diff
@@ -1191,7 +1191,7 @@ public enum LegendMaker implements LogicCardInfo {
           }
           onAttack {
             def maxSpace = Math.min(my.bench.freeBenchCount, 2)
-            my.deck.search(min:0, max:maxSpace, "Search your deck for up to 2 cards named Omanyte, Kabuto, Aerodactyl, Lileep, or Anorith", {
+            my.deck.search(min:0, max:maxSpace, "Search your deck for ${maxSpace == 2 ? "up to 2 cards" : "a card"} named Omanyte, Kabuto, Aerodactyl, Lileep, or Anorith", {
```
flagrama, line 1194: I still don't like doing this kind of stuff as it really isn't conducive to adding more languages in the future.

src/tcgwars/logic/impl/gen3/PowerKeepers.groovy
```diff
@@ -397,9 +406,14 @@ public enum PowerKeepers implements LogicCardInfo {
             checkNoSPC()
             checkLastTurn()
             assert my.hand.filterByType(ENERGY) : "No Energy in hand"
-            my.hand.filterByType(ENERGY).select("Discard").discard()
-            draw choose([0,1,2,3], ["0","1","2","3"], "Draw how many cards?", 3)
+            //TODO: Confirm whether to use EX or LV.X ruling regarding being able to use it without cards in deck (EX ruling: You can't use the power at all; LV.X ruling: you can, you discard the energy and nothing else happens)
```
flagrama, line 409: I'd say newest ruling trumps an older ruling especially if they are about the same exact situation and card.

src/tcgwars/logic/impl/gen3/PowerKeepers.groovy
```diff
@@ -421,13 +435,11 @@ public enum PowerKeepers implements LogicCardInfo {
...
             powerUsed()
 
-            my.deck.search(max: 1, "Search for a [P] Energy card to attach to one of your Pokemon.", {
-              Card card->card.cardTypes.contains(BASIC_ENERGY) && card.asEnergyCard().containsTypePlain(P)
-              }).each {
-                def tar = my.all.select("Attach $it to? That Pokemon will receive 2 damage counters.")
-                attachEnergy(tar, it)
-                directDamage 20, tar
-              }
+            my.deck.search(max: 1, "Search for a [P] Energy card to attach to one of your Pokemon.", {Card card -> card.asEnergyCard().containsTypePlain(L)}).each {
```
flagrama, line 438: Why is this looking for Lightning type instead of Psychic type now?

src/tcgwars/logic/impl/gen3/PowerKeepers.groovy
```diff
@@ -445,8 +457,9 @@ public enum PowerKeepers implements LogicCardInfo {
         pokeBody "Primal Stare", {
           text "As long as Kabutops is your Active Pokémon, your opponent can't play any Basic Pokémon or Evolution cards from his or her hand to evolve his or her Active Pokémon."
           delayedA {
-            before EVOLVE_STANDARD, self.owner.opposite.pbg.active,{
-              if (bg.currentTurn==self.owner.opposite) {
+            before EVOLVE_STANDARD, {
+              if (ef.pokemonToBeEvolved.owner != self.owner && ef.pokemonToBeEvolved.active && self.active) {
+                wcu "Primal Stare prevents evolving your Active Pokémon"
                 prevent()
```
flagrama, line 463: This doesn't seem to handle the case of playing Basic Pokémon.

src/tcgwars/logic/impl/gen3/PowerKeepers.groovy
```diff
@@ -595,7 +615,7 @@ public enum PowerKeepers implements LogicCardInfo {
           text "Once during your opponent's turn, when any of your Pokémon is Knocked Out by your opponent's attacks, you may use this power. Choose a basic Energy card discarded from the Knocked Out Pokémon and attach it to Lanturn. You can't use more than 1 Energy Grounding Poké-Power each turn."
           delayedA{
             before KNOCKOUT, {
-              if ((ef as Knockout).byDamageFromAttack && bg.currentTurn==self.owner.opposite && ef.pokemonToBeKnockedOut!=self  && ef.pokemonToBeKnockedOut.cards.energyCount(C)) {
+              if ((ef as Knockout).byDamageFromAttack && bg.currentTurn==self.owner.opposite && ef.pokemonToBeKnockedOut != self  && ef.pokemonToBeKnockedOut.cards.filterByType(BASIC_ENERGY).energyCount(C)) {
                 if (confirm("Move an energy from ${ef.pokemonToBeKnockedOut} to $self?")) {
                   moveEnergy(basic:true, ef.pokemonToBeKnockedOut,self)
                 }
```
flagrama, line 621: Does this handle the "Once during your opponent's turn" part of the ability properly?

src/tcgwars/logic/impl/gen3/PowerKeepers.groovy
```diff
@@ -2006,31 +2067,30 @@ public enum PowerKeepers implements LogicCardInfo {
...
         weakness R
+        weakness F
         resistance G, MINUS30
         pokePower "Magnetic Redraw", {
           text "Once during your turn (before your attack), if Metagross ex is your Active Pokémon, you may use this power. Each player shuffles his or her hand into his or her deck. Then, each player draws 4 cards. This power can't be used if Metagross ex is affected by a Special Condition."
           actionA {
             checkLastTurn()
             checkNoSPC()
             assert self.active : "$self is not an Active Pokémon"
+            assert ( [my, opp].any{curPlayer -> curPlayer.hand || curPlayer.deck} ) : "No player has cards in their deck or in their hand (...how?)"
```
flagrama, line 2085: That statement in parenthesis isn't necessary. It may be a rare case, but it could happen. Why didn't the player using the PokéPower just pass in that case? Who knows. Why do PTCGO players play solitaire when they can just attack.

src/tcgwars/logic/impl/gen3/LegendMaker.groovy
starg09: Whenever that happens, the entire gstring needs to be translated regardless. As long as the logic is properly converted, it shouldn't be a problem right? It can always be turned down into "up to 1 cards" if needed.

src/tcgwars/logic/impl/gen3/PowerKeepers.groovy
starg09: Yup, will remove this TODO in #877 after checking around and talking with players at the time. Specially with the card itself being the specific one the ruling was altered on shortly after.

src/tcgwars/logic/impl/gen3/PowerKeepers.groovy
starg09: tfw thought I'd tested this card earlier. Will fix in #877, thanks!

src/tcgwars/logic/impl/gen3/PowerKeepers.groovy
starg09: That's... A really weird wording. Is it meant to block Baby Evolution maybe? Will check in a bit.

src/tcgwars/logic/impl/gen3/PowerKeepers.groovy
starg09: > bg.currentTurn==self.owner.opposite

That should cover for it, and there should be no way for a KO to be caused by damaged during between turns.

src/tcgwars/logic/impl/gen3/PowerKeepers.groovy
starg09: Eh, just a little joke comment 😛 

There's def 99% chances of this never happening. It'd require all 60 cards (or 54 if not counting initial prizes) to be either in some PCS in play, as a stadium placed, or in the lost zone.

src/tcgwars/logic/impl/gen3/PowerKeepers.groovy
flagrama: An attack that knocks out the active and a benched Pokémon or multiple benched Pokémon, such as Tag Bolt-GX or a spread attack would activate this multiple times from my understanding.

src/tcgwars/logic/impl/gen3/PowerKeepers.groovy
flagrama: Or in the Discard Pile

src/tcgwars/logic/impl/gen3/PowerKeepers.groovy
starg09: Testing with Dracovish (DAA), that also seems to fail to stop Baby Evolution. Will need to think it a bit more on how to prevent it, seems to be the main case that this is meant to cover re: Basic Pokémon.

src/tcgwars/logic/impl/gen3/PowerKeepers.groovy
starg09: Hmm, that's a potential case. Hadn't noticed it had that difference with the """reprint""" from Celestial Storm.

Compendium doesn't seem to cover that case, the Muk-ex one is potentially a timing nightmare though oof.

src/tcgwars/logic/impl/gen3/PowerKeepers.groovy
starg09: Not sure why I specifically ignored that being a possibility. Still, it's an unlikely situation and a minor note, but I'll remove it 🤷

src/tcgwars/logic/impl/gen3/LegendMaker.groovy
flagrama: Perhaps. Though, let me tell you, plurals are ... "fun" https://developer.mozilla.org/en-US/docs/Mozilla/Localization/Localization_and_Plurals

src/tcgwars/logic/impl/gen3/PowerKeepers.groovy
starg09: Ok Muk-ex was only in a forum post, must've been removed after confirming.

Back to this, the "once per turn" probably cover most potential cases. Will add the restriction in #877

src/tcgwars/logic/impl/gen3/PowerKeepers.groovy
flagrama: Probably something best left for using a debugger on to figure out the proper fix honestly.

starg09: Yup. Will leave a TODO in it.

PR #878, opened 2020-07-16
Add some Trainer card's requirements

starg09:
Initially was meant to fix Ultra Recon Squad, but a couple other cards could also use some requirements.

### DIFF
--- a/src/tcgwars/logic/impl/gen7/UltraPrism.groovy
+++ b/src/tcgwars/logic/impl/gen7/UltraPrism.groovy
@@ -3033,7 +3033,7 @@ public enum UltraPrism implements LogicCardInfo {
           text "You may play 4 Missing Clover cards at once. \n-If you played 1 card, look at the top card of your deck \n if you played 4 cards, take a prize card \nYou may play as many Item cards as you like during your turn (before your attack)."
           onPlay {
             if(my.hand.findAll({it.name=="Missing Clover"}).size()>=4) {
-              if(confirm("Use your 4 Missing Clover and take a prize card?")){
+              if(!my.deck || confirm("Use your 4 Missing Clover and take a prize card?")){
                 bg.em().run(new TakePrize(thisCard.player, null))
                 my.hand.findAll({it.name=="Missing Clover"}).subList(0,4).discard()
               }
@@ -3042,6 +3042,7 @@ public enum UltraPrism implements LogicCardInfo {
             }
           }
           playRequirement{
+            assert (my.hand.count{it.name=="Missing Clover"} >= 4 || my.deck) : "You have no cards in your deck"
           }
         };
       case MT__CORONET_130:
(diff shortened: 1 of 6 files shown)


src/tcgwars/logic/impl/gen7/UltraPrism.groovy
```diff
@@ -3033,7 +3033,7 @@ public enum UltraPrism implements LogicCardInfo {
           text "You may play 4 Missing Clover cards at once. \n-If you played 1 card, look at the top card of your deck \n if you played 4 cards, take a prize card \nYou may play as many Item cards as you like during your turn (before your attack)."
           onPlay {
             if(my.hand.findAll({it.name=="Missing Clover"}).size()>=4) {
-              if(confirm("Use your 4 Missing Clover and take a prize card?")){
+              if(!my.deck || confirm("Use your 4 Missing Clover and take a prize card?")){
```
luongthomas, line 3036: Why is it important that your deck is not empty when you play 4 missing clovers?

starg09: It should (in theory) avoid asking if you wanna play the 4 clovers together, since it should be the only option (you can't look at the top of an empty deck, so playing a single clover wouldn't be allowed then)

src/tcgwars/logic/impl/gen7/UnbrokenBonds.groovy
```diff
@@ -4256,6 +4256,13 @@ public enum UnbrokenBonds implements LogicCardInfo {
             bg.em().storeObject("KOGA_S_TRAP_TURN", bg.turnCount)
           }
           playRequirement{
+            def defendingConfused = opp.active.isSPC(CONFUSED)
+            def defendingPoisoned = opp.active.isSPC(POISONED)
+            if (defendingPoisoned){
+              //Check for multi-counter poison; this only detects active increasers.
+              def extraPoisonCount = bg.em().retrieveObject("extra_poison_counter_"+opp.active.hashCode()) ?: 0
+            }
+            assert !defendingConfused || !defendingPoisoned || extraPoisonCount : "Your opponent's Active Pokémon is already Confused and Poisoned"
```
luongthomas, line 4265: Wouldn't this block the situation where the you want to use Koga's Trap when the opponent's active is either only confused or only poisoned?

starg09: As long as one of the three conditions is true, it should allow to use the card. I'm on doubt about extraPoisonCount being reset when poison is removed though, will check in a bit.

But being only poisoned/confused should pass this assert.

starg09: Ok, there seems to be an EE on Vileplume Delta. Pushed a fix for it and Koga, should work fine now (will push a fix if it keeps erroring after merge).

PR #879, opened 2020-07-16
Some DAA Fixes As They Come Up

flagrama:


flagrama:
[TOXAPEX_54](https://trello.com/c/tLeyVDcN/488-toxapex54)

axpendix:
Hey, what is the status of this one?

flagrama:
A couple more cards had issues or things I need to check. Shouldn't actually interfere with #883 and if it does, I can rebase.

src/tcgwars/logic/impl/gen8/DarknessAblaze.groovy
```diff
@@ -1,4 +1,5 @@
-package tcgwars.logic.impl.gen8;
+package tcgwars.logic.impl.gen8
```
flagrama, line 1: I guess IntelliJ didn't like all these imports. I can undo this if needed.

src/tcgwars/logic/impl/gen8/DarknessAblaze.groovy
```diff
@@ -14,21 +15,11 @@ import static tcgwars.logic.effect.EffectPriority.*
...
-import tcgwars.logic.effect.ability.*;
-import tcgwars.logic.effect.advanced.*;
-import tcgwars.logic.effect.basic.*;
-import tcgwars.logic.effect.blocking.*;
-import tcgwars.logic.effect.event.*;
-import tcgwars.logic.effect.getter.*;
+import tcgwars.logic.effect.ability.*
+import tcgwars.logic.effect.basic.*
 import tcgwars.logic.effect.special.*;
 import tcgwars.logic.util.*;
```
flagrama, line 25: I guess IntelliJ didn't like all these imports. I can undo this if needed.

src/tcgwars/logic/impl/gen8/DarknessAblaze.groovy
```diff
@@ -1390,9 +1387,9 @@ public enum DarknessAblaze implements LogicCardInfo {
         bwAbility "Primal Law", {
           text "If this Pokémon is your Active Pokémon, your opponent can’t play any Pokémon from their hand to evolve their Pokémon."
           delayedA {
-            before PLAY_EVOLUTION, {
-              if (self.active && bg.currentTurn == self.owner.opposite && ef.cardToPlay.player.pbg.hand.contains(ef.cardToPlay)) {
-                wcu "$self.name's Primal Law prevents playing Pokémon from your hand to evolve your Pokémon"
+            before EVOLVE_STANDARD, {
+              if (self.active && bg.currentTurn == self.owner.opposite && (ef as EvolveStandard).evolutionCard.player.pbg.hand.contains(ef.evolutionCard)) {
+                wcu "$self's Primal Law prevents playing Pokémon from your hand to evolve your Pokémon"
```
starg09, line 1392: Really minor detail, but it seems two warning conflict with each other when directly playing an evolution from hand.

Old Rare Candy seems to skip this prevention still, but that's not really important (and shouldn't even be a problem for Unlimited, whenever common code is implemented for it like with Pokemon Catcher)

Ditto Prism Star seems to bypass this effect. Same with Ascension DNA in Eevee-GX, and Quick-Ripening Herb in Meganium (Lost Thunder).

starg09: > Really minor detail, but it seems two warning conflict with each other when directly playing an evolution from hand

Oh, forgot to pass a screenshot I took of that:

image

flagrama: The stacking issue seems to be something Vaadin 14 fixes https://github.com/vaadin/framework/issues/676#issuecomment-544166861

Playing an evolution directly from hand shouldn't cause stacked notifications now, however Rare Candy does. A lot. 4 of them. 2 of which are empty currently due to a bug in the main repo.

src/tcgwars/logic/impl/gen8/DarknessAblaze.groovy
```diff
@@ -1390,12 +1387,22 @@ public enum DarknessAblaze implements LogicCardInfo {
         bwAbility "Primal Law", {
           text "If this Pokémon is your Active Pokémon, your opponent can’t play any Pokémon from their hand to evolve their Pokémon."
           delayedA {
-            before PLAY_EVOLUTION, {
-              if (self.active && bg.currentTurn == self.owner.opposite && ef.cardToPlay.player.pbg.hand.contains(ef.cardToPlay)) {
-                wcu "$self.name's Primal Law prevents playing Pokémon from your hand to evolve your Pokémon"
+            def warnAndPrevent = {
+              if (self.active && bg.currentTurn == self.owner.opposite) {
+                wcu "$self's Primal Law prevents playing Pokémon from your hand to evolve your Pokémon"
                 prevent()
```
axpendix, line 1393: Not sure this works here 🤔 `prevent()` is a function of the `before` closure's `delegate`. Have you tested this?

src/tcgwars/logic/impl/gen8/DarknessAblaze.groovy
axpendix: I think it's better to revert these

src/tcgwars/logic/impl/gen8/DarknessAblaze.groovy
starg09: Tested it yesterday and it seemed to work. Just tested with the last changes and debugging, prevent executes without issue 👍

src/tcgwars/logic/impl/gen8/DarknessAblaze.groovy
starg09: > however Rare Candy does
Doesn't seem to pop em up here now, lol.

Hey, if it works it works 😛

src/tcgwars/logic/impl/gen8/DarknessAblaze.groovy
flagrama: I have, and it seems to work fine. Pretty sure this closure's delegate is the before, then the before's delegate has a prevent, so it kind of just bubbles up the tree.

PR #881, opened 2020-07-20
Batch of fixes

starg09:


src/tcgwars/logic/groovy/TcgStatics.groovy
```diff
@@ -1396,6 +1396,19 @@ class TcgStatics {
...
   }
 
+  static putDamageCountersOnOpponentsPokemon(int counters, def selectArea = opp.all){
+    def eff = delayed {
+      before KNOCKOUT, {
+        prevent()
+      }
+    }
+
+    counters.times{ directDamage 10, selectArea.select("Put 1 damage counter on which pokémon? ${it-1}/${counters} counters placed") }
```
luongthomas, line 1406: Using `JYNX_28`'s `Pure Power` move, the first dialog message shows up as 
```
Put 1 damage counter on which pokémon? -1/4 counters placed
```

It still correctly allows 4 damage counters to be put on though. 👍

luongthomas:
This builds locally just fine

src/tcgwars/logic/groovy/TcgStatics.groovy
starg09: Hmm, apparently times does a 0-array for it's value (which makes sense thinking about it; will do a quick check for it maybe causing similar issues somewhere else, before pushing)

Thanks 👍

PR #884, opened 2020-07-21
Fix Exeggutor FRLG

Grayahm:


src/tcgwars/logic/impl/gen3/FireRedLeafGreen.groovy
```diff
@@ -317,10 +317,10 @@ public enum FireRedLeafGreen implements LogicCardInfo {
             text "Shuffle your hand into your deck. Draw up to 8 cards."
             energyCost C
             attackRequirement {
-              assert my.hand.notEmpty && my.deck.notEmpty
+              assert my.deck.notEmpty
```
starg09, line 320: This should've been turned into an `||` instead of outright removed. Should be playable when you have hand but no deck.

PR #885, opened 2020-07-21
Fix Exeggutor #2

Grayahm:


Grayahm:
> Put it in #884 but that's already merged: The assert should check for `my.hand.notEmpty || my.deck.notEmpty` instead of being removed.
> 
> The other change in this PR is 👍 , but could you include this fix here as well?

This is correct, you should be able to use this attack with no cards in hand.

starg09:
> This is correct, you should be able to use this attack with no cards in hand.

That is fine, but with the change applied now you can't play it when you have no cards in deck (even though you have cards in hand, which may be more than 8 thus avoiding a deckout through the attack.)

Grayahm:
Oh I understand now, I'll change that.

PR #886, opened 2020-07-29
Electrode ex Crush and Burn fix

Grayahm:
Crush and Burn no longer allows you to select the same energy card indefinitely.


starg09:
Already patched in #881. Just waiting on a review before merging and testing it 👍

PR #890, opened 2020-08-02
General Fixes

starg09:
Moved some stuff from #888, and started working on some other more immediate fixes (the switch changes are a bit larger, I'd rather leave those on its own PR without these mixed in-between.)

Fixes to confirm working once merged: 

- [x] Frogadier (Unbroken Bonds) should no longer crash when "Afterimage Strike" flips heads.
- [x] Dragonite δ (Delta Species) should no longer be able to use "Delta Charge" when no [L] Energy is found in the discard pile.
- [x] Attacks that use the "put damage counters in the opponent's Pokémon" static should now check for Pokémon being in the area (mainly for when counters are placed in bench only). This should fix a timeout bug due to no valid target being selectable, specifically in Dragapult VMAX (Rebel Clash).
- [x] Tapu Lele (SM Promo SM45) should stop crashing after using Magical Swap, after adding a missing `def`.
    + Made it so both this one and the Ultra Prism print no longer allow moving a counter from a Pokémon to the same one (which didn't make a lot of sense)
- [x] Staraptor (Crimson Invasion) should now switch due to KO only once when using "Sky Hunting".
- [x] Added a bc for Porygon (Team Rocket). Not really a major change, and it's a groovy set not currently being used but might as well add it 😛 
- [x] Added a missing comma in Roserade[C] (Supreme Victors).
- [x] Fixed Rainbow Energy, Double Rainbow Energy, R Energy and Dark Metal Energy counting as their energy types while in hand when they shouldn't.
- [x] Updated Magnezone (Ultra Prism, Forbidden Light) and Frosmoth (Sword & Shield - Base Set) to work similarly to Porygon-Z (Unbroken Bonds) and Blastoise-ex (FireRed & LeafGreen).
- [x] Added a missing assert fail message on Feraligatr (Dragon Majesty).
- [x] Added some missing asserts for Quagsire (Dragon Majesty), Weavile (Unified Minds) and Gengar (Sword & Shield - Base Set).
- [x] Fixed Torkoal V (Sword & Shield - Base Set) so it discards Energies after damage is applied.
- [x] Fixed multiple "You can evolve first turn" Abilities. All now make a message in log, but also are restricted to its owner's turn (currently can be played on opponent's turn)
- [x] Fixed Full Flame not preserving the Burned condition on evolve/devolve.

### DIFF
--- a/src/tcgwars/logic/groovy/TcgStatics.groovy
+++ b/src/tcgwars/logic/groovy/TcgStatics.groovy
@@ -1397,16 +1397,18 @@ class TcgStatics {
   }
 
   static putDamageCountersOnOpponentsPokemon(int counters, def selectArea = opp.all){
-    def eff = delayed {
-      before KNOCKOUT, {
-        prevent()
+    if (selectArea.notEmpty) {
+      def eff = delayed {
+        before KNOCKOUT, {
+          prevent()
+        }
       }
+
+      counters.times{directDamage 10, selectArea.select("Put 1 damage counter on which pokémon? ${it}/${counters} counters placed") }
+
+      eff.unregister()
+      checkFaint()
     }
-
-    counters.times{directDamage 10, selectArea.select("Put 1 damage counter on which pokémon? ${it}/${counters} counters placed") }
-
-    eff.unregister()
-    checkFaint()
   }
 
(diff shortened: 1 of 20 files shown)


starg09:
All fixes seem to work correctly after testing.

PR #891, opened 2020-08-02
Fix some SPC prevention logic

starg09:
Should now properly detect the targeted PCS, and prevent only when meant to. Fixes some issues related to "gust of wind" effects.

### DIFF
--- a/src/tcgwars/logic/impl/gen7/LostThunder.groovy
+++ b/src/tcgwars/logic/impl/gen7/LostThunder.groovy
@@ -3522,10 +3522,12 @@ public enum LostThunder implements LogicCardInfo {
                 power=false
               }
               before null, null, Source.TRAINER_CARD, {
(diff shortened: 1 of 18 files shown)


src/tcgwars/logic/impl/gen7/LostThunder.groovy
```diff
@@ -3522,7 +3522,7 @@ public enum LostThunder implements LogicCardInfo {
                 power=false
               }
               before null, null, Source.TRAINER_CARD, {
-                def target=e.getTarget(bg)
+                def target = ef.getResolvedTarget(bg, e)
```
axpendix, line 3525: not all `ef` instances have this method, so it may throw exception. better add a condition with `ef instanceof TargetedEffect`

starg09:
Oh, missed the "you may merge" part. Sorry for the ping for review 😛

PR #893, opened 2020-08-05
Specify Coin Flip Player If Needed

flagrama:


axpendix:
One small concern, I'd probably put this as third parameter before the closures.

src/tcgwars/logic/impl/gen3/TeamRocketReturns.groovy
```diff
@@ -485,7 +485,7 @@ public enum TeamRocketReturns implements LogicCardInfo {
                     self.damage = self.fullHP - hp(10)
                     bc "Buffer saved $self!"
                     prevent()
-                  }
+                  }, self.owner
```
axpendix, line 488: Does it work like this?

src/tcgwars/logic/groovy/TcgStatics.groovy
```diff
@@ -90,9 +96,13 @@ class TcgStatics {
...
     flip(info, 1, eachHead, eachTail, multi)
   }
-  static flip (String info, int count, Closure eachHead, Closure eachTail={}, multi=[:]){
+  static flip (String info, int count, Closure eachHead, Closure eachTail={}, multi=[:], playerType=null){
+    if(playerType == null) {
+      playerType = bg.getCurrentTurn()
+    }
     CoinFlip cf=new CoinFlip(count, toEffect (eachHead), toEffect (eachTail))
     cf.setInfo(info)
+    cf.setPlayer(playerType)
```
axpendix, line 105:
```suggestion
    CoinFlip cf=new CoinFlip(count, toEffect (eachHead), toEffect (eachTail))
    cf.setInfo(info)
    if(playerType) {
      cf.setPlayer(playerType)
    }
```

axpendix: The resolution decision should be left to `CoinFlip` instead of here.

axpendix:
I think it'll be better to put the `player` parameter as third for all `coinFlip` functions.

src/tcgwars/logic/impl/gen3/TeamRocketReturns.groovy
flagrama: Yes, it does.
```
● Glimwood Tangle is now in play.
● Inteleon VMAX used Hydro Snipe
● Hoppip resists WATER
● Inteleon VMAX did 30 damage to Hoppip
● Buffer coin flip result: HEADS
● Buffer saved Hoppip!
● Between turns
● Turn 4 begins
● Player B drew 1 card
● Between turns
● Turn 5 begins
● Player A drew 1 card
● Inteleon VMAX used Hydro Snipe
● Hoppip resists WATER
● Inteleon VMAX did 30 damage to Hoppip
● Buffer coin flip result: TAILS
● Hoppip was knocked out
● Player A took one prize card.
● Hoppip is now active
● Between turns
● Turn 6 begins
● Player B drew 1 card
● [DEV] A: added Paras (DAB 3) to HAND
● Between turns
● Turn 7 begins
● Player A drew 1 card
● Put Paras onto bench
● Between turns
● Turn 8 begins
● Player B drew 1 card
● Between turns
● Turn 9 begins
● Player A drew 1 card
● [DEV] A: added Parasect (DAB 4) to HAND
● Paras evolved into Parasect
● Parasect is now active
● [DEV] A: switched
● Parasect used Mushroom Tackle
● Coin flip result: TAILS
● Used Glimwood Tangle and discarded those flips
● Coin flip result: TAILS
● Parasect did 30 damage to Hoppip
● Buffer coin flip result: HEADS
● Buffer saved Hoppip!
● Between turns
● Turn 10 begins
● Player B drew 1 card
```
The coin flip works to save the Hoppip if heads, Glimwood Tangle will not activate for that effect. Glimwood Tangle does still work for the attacker though.

flagrama:
Yeah, I suppose that should be fine. Groovy is better about handling optional parameters than I thought.

axpendix:
The problem is I merged https://github.com/axpendix/tcgone/pull/23 so the implementation changes need to change again 😬 sorry

flagrama:
Force-pushed because I committed a dumb while trying to fix another dumb. Sorry.

src/tcgwars/logic/groovy/TcgStatics.groovy
flagrama: Hmm, testing this, the issue I'm having is that player is null `BEFORE` the coin flip, so I can't check the player then, and since AFAICT you can't access the current player without access to the battleground. This works fine otherwise though.

The use case for my issue is Will from CEC. I'm trying to make sure he can only be used by the player playing him instead of the opponent. It works fine how I had it, but gets a null player this way. I can check for this using `if ((ef.player && ef.player == bg.getCurrentTurn()) || !ef.player)` and it seems to work as intended since opponent coin flips explicitly set their player value now, but seems slightly awkward to do. If you don't find it an issue because it still works though, and it probably isn't a very common effect, I can leave it as is.

axpendix: The use case for my issue is Will from CEC. I'm trying to make sure he can only be used by the player playing him instead of the opponent. It works fine how I had it, but gets a null player this way. I can check for this using if ((ef.player && ef.player == bg.getCurrentTurn()) || !ef.player) and it seems to work as intended since opponent coin flips explicitly set their player value now, but seems slightly awkward to do. If you don't find it an issue because it still works though, and it probably isn't a very common effect, I can leave it as is.

Do you mean if Will gets used by another player via an effect like this?

Smeargle (UD 8)
Portrait: Once during your turn (before your attack), if Smeargle is your Active Pokémon, you may look at your opponent's hand. If you do, choose a Supporter card you find there and use the effect of that card as the effect of this power. This power can't be used if Smeargle is affected by a Special Condition.

If yes, Those effects should be wrapped with bg.setDeterministicCurrentThreadPlayerType() call.

axpendix: Not relevant 1-1 but good to think of; 

Another use case to be aware of, is when Portrait is used with Will. Portrait should handle it by wrapping Will's call with `bg.setDeterministicCurrentThreadPlayerType()`. 

> Smeargle (UD 8)
> Portrait: Once during your turn (before your attack), if Smeargle is your Active Pokémon, you may look at your opponent's hand. If you do, choose a Supporter card you find there and use the effect of that card as the effect of this power. This power can't be used if Smeargle is affected by a Special Condition.

axpendix: > Ok I now understand what you mean. That is an interesting use case. There are couple of options;
> 
> 1. resolve player when null in contructor of CoinFlip (but this won't work with legacy sets due to single-instance usage)
> 2. resolve it in `getPlayer()` method (still weird)
> 3. I'd use this condition in Will instead: `(ef.player ?: bg.currentThreadPlayerType) == bg.currentThreadPlayerType`

I've updated this comment multiple times, not sure you've seen my last edit so tagging you @flagrama

flagrama: I'll give that a shot when I get back to this again.

src/tcgwars/logic/impl/gen7/CosmicEclipse.groovy
```diff
@@ -4993,13 +4993,16 @@ public enum CosmicEclipse implements LogicCardInfo {
           onPlay {
             delayed {
               def doit = {
-                boolean res = choose([true,false],["HEADS","TAILS"],"A coin flip is about to occur. Choose HEADS or TAILS as the result.")
-                bc "Will changes the first coinflip to: ${res?'HEADS':'TAILS'}"
-                bg.deterministicCoinFlipQueue.offer(res)
-                unregister()
+                if ((ef.player && ef.player == bg.getCurrentTurn()) || !ef.player) {
```
axpendix, line 4996:
```suggestion
                if ((ef.player ?: bg.currentThreadPlayerType) == bg.currentTurn) {
```

src/tcgwars/logic/groovy/TcgStatics.groovy
```diff
@@ -101,9 +104,10 @@ class TcgStatics {
   static flipTails (Closure eachTail) {
     flip(1, {}, eachTail)
   }
-  static flipUntilTails (Closure eachHead){
+  static flipUntilTails (Closure eachHead, playerType=null){
```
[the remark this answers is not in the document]
axpendix: I'll merge this PR and leave this for another one

PR #897, opened 2020-08-07
Update benchPCS to Allow Checking if Benching Succeeded

luongthomas:
This PR introduces an update to the `benchPCS` method to allow the caller to know if the action of adding the Pokemon to the bench was successful or not.  

This is achieved through the return type of running the `PutOnBench` effect.  

If `benchPCS` successfully placed a Pokemon on the bench, the `PCS` will be returned to the caller, otherwise `null` will be returned.  

As a side-effect, I've also updated `callForFamily` to only remove the selected card from the deck if the benching was successful.

This PR is dependent on https://github.com/axpendix/tcgone/pull/29 (`Update PutOnBench Effect to Allow Checking if Succeeded`)


src/tcgwars/logic/impl/gen3/UnseenForces.groovy
```diff
@@ -3451,12 +3451,15 @@ public enum UnseenForces implements LogicCardInfo {
...
-                opp.hand.remove(it)
-                tar = benchPCS(it, OTHER, TargetPlayer.OPPONENT)
+              def tar = basics.select("Choose a pokemon to bench")
+              if (tar) {
+                def card = tar.first()
+                opp.hand.remove(card)
+                def benched = benchPCS(card, OTHER, TargetPlayer.OPPONENT)
+                if (benched) {
+                  sw2 (benched)
+                }
```
starg09, line 3461: Card removal would need to happen inside the if here, right?

```suggestion
              def tar = basics.select("Choose a pokemon to bench")
              if (tar) {
                def card = tar.first()
                def benched = benchPCS(card, OTHER, TargetPlayer.OPPONENT)
                if (benched) {
                  opp.hand.remove(card)
                  sw2 (benched)
                }
```

src/tcgwars/logic/groovy/TcgStatics.groovy
```diff
@@ -489,16 +489,14 @@ class TcgStatics {
     if(card.getCardTypes().is(BREAK)){
       bg.wcu("BREAK Pokémon cannot be brought to play")
     }
-    PlayerType playerType=targetPlayer.getPlayerType(bg())
     PlayerBattleground pbg=targetPlayer.getPbg(bg())
     assert pbg.bench.isNotFull() : "Bench is full"
```
axpendix, line 493: This check should be done in the new PutOnBench implementation rather than here. It should then throw an `EffectRequirementException`.

luongthomas: I've moved the effect over to PutOnBench

src/tcgwars/logic/groovy/TcgStatics.groovy
```diff
@@ -485,20 +485,16 @@ class TcgStatics {
       opp.bench.remove(pcs)
     }
   }
-  static PokemonCardSet benchPCS (Card card, ActivationReason reason=ActivationReason.OTHER, TargetPlayer targetPlayer=TargetPlayer.SELF){
+  static PokemonCardSet benchPCS (Card card, ActivationReason reason=OTHER, TargetPlayer targetPlayer=TargetPlayer.SELF){
```
axpendix, line 488: targetPlayer parameter should be removed, along with all usages

src/tcgwars/logic/groovy/TcgStatics.groovy
```diff
@@ -485,20 +485,16 @@ class TcgStatics {
...
     if(card.getCardTypes().is(BREAK)){
```
axpendix, line 489: Sorry to bother you again but In my opinion this is another condition check to be moved to the main class.

src/tcgwars/logic/impl/gen3/UnseenForces.groovy
```diff
@@ -3451,12 +3450,15 @@ public enum UnseenForces implements LogicCardInfo {
...
-              def tar
-              basics.select("Choose a pokemon to bench").each {
-                opp.hand.remove(it)
-                tar = benchPCS(it, OTHER, TargetPlayer.OPPONENT)
+              def tar = basics.select("Choose a pokemon to bench")
+              if (tar) {
+                def card = tar.first()
+                def benched = benchPCS(card, OTHER, TargetPlayer.OPPONENT)
+                if (benched) {
+                  opp.hand.remove(card)
```
axpendix, line 3458: 🪓

src/tcgwars/logic/impl/gen7/BurningShadows.groovy
```diff
@@ -1310,7 +1308,6 @@ public enum BurningShadows implements LogicCardInfo {
               def list = opp.hand.shuffledCopy().showToMe("Opponent's hand").filterByType(BASIC)
               if(list){
                 def card = list.select("Put a Basic Pokémon you find there onto your opponent's Bench").first()
-                opp.hand.remove(card)
                 def pcs = benchPCS(card, OTHER, TargetPlayer.OPPONENT)
                 directDamage(30, pcs, SRC_ABILITY)
```
axpendix, line 1312: Add condition pls

src/tcgwars/logic/impl/gen7/CelestialStorm.groovy
```diff
@@ -2769,7 +2766,6 @@ public enum CelestialStorm implements LogicCardInfo {
               def maxSpace = Math.min(my.bench.freeBenchCount,3)
               def hasBenched = false
               deck.search (min: 0,max:maxSpace,cardTypeFilter(BASIC)).each {
-                deck.remove(it)
                 benchPCS(it)
                 hasBenched = true
```
axpendix, line 2770: Wrap with condition pls

src/tcgwars/logic/groovy/TcgStatics.groovy
```diff
@@ -485,20 +485,16 @@ class TcgStatics {
...
-    assert pbg.bench.isNotFull() : "Bench is full"
-    PokemonCardSet bench = new PokemonCardSet(playerType);
-    bench.cards().add(card);
-    pbg.getBench().add(bench);
-    bc "$bench was put on bench"
-    bg().em().run(new CantEvolve(bench, bg().getTurnCount()));
-    bg().em().run(new ActivateAbilities((PokemonCard) card, bench, reason));
-    bench
+
+    Effect effect = new PutOnBench(card, reason);
```
axpendix, line 493: Using def instead of Effect would be better

src/tcgwars/logic/groovy/TcgStatics.groovy
luongthomas: Removed third parameter for TargetPlayer and all usages

src/tcgwars/logic/impl/gen3/UnseenForces.groovy
luongthomas: Axed

src/tcgwars/logic/impl/gen7/BurningShadows.groovy
luongthomas: Added conditions here and in similar usages

src/tcgwars/logic/groovy/TcgStatics.groovy
luongthomas: Moved this check over to the PutOnBench effect

PR #899, opened 2020-08-08
Update TYRANITAR_88 to not crash during Megaton Tail attack

luongthomas:
This fixes the crash when using Megaton Tail on `TYRANITAR_88` from Unleashed.

### DIFF
--- a/src/tcgwars/logic/impl/gen4/Unleashed.groovy
+++ b/src/tcgwars/logic/impl/gen4/Unleashed.groovy
@@ -1896,8 +1896,8 @@ public enum Unleashed implements LogicCardInfo {
             energyCost D, D, C, C
             onAttack {
               damage 120
-              afterDamage{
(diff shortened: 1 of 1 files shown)


src/tcgwars/logic/impl/gen4/Unleashed.groovy
```diff
@@ -1896,8 +1896,8 @@ public enum Unleashed implements LogicCardInfo {
             energyCost D, D, C, C
             onAttack {
               damage 120
-              afterDamage{
-                my.deck.sublist(0,min(3, my.deck.size())).discard()
+              afterDamage {
+                my.deck.subList(0,Math.min(3, my.deck.size())).discard()
```
starg09, line 1900:
```suggestion
                if (my.deck) my.deck.subList(0,Math.min(3, my.deck.size())).discard()
```

Not sure if this is needed (although it's used in a number of similar cards), but just in case.

PR #902, opened 2020-08-09
Fix a crash on the flip static

starg09:


axpendix:
It works if it works!

PR #903, opened 2020-08-09
Implement Eternatus VMAX ability

luongthomas:
This PR introduces the first version of Eternatus VMAX's ability

The changes here are dependent on 2 other PRs

https://github.com/axpendix/tcgone/pull/29
https://github.com/axpendix/tcgone-engine-contrib/pull/897

### DIFF
--- a/src/tcgwars/logic/impl/gen8/DarknessAblaze.groovy
+++ b/src/tcgwars/logic/impl/gen8/DarknessAblaze.groovy
@@ -586,9 +586,9 @@ public enum DarknessAblaze implements LogicCardInfo {
           attackRequirement {}
           onAttack {
             damage 100
-            if (confirm("Return an Energy attached to the Active $opp.active.name to your opponent's hand?")) {
-              def info = "Select the Energy to return to the opponent's hand."
-              def selectedEnergy = defending.cards.filterByType(ENERGY).select info
+            if (defending.cards.filterByType(ENERGY)) {
+              def info = "You may select the Energy to return to the opponent's hand."
+              def selectedEnergy = defending.cards.filterByType(ENERGY).select min:0, info
               if (selectedEnergy) selectedEnergy.moveTo opp.hand
             }
           }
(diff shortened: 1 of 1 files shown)


luongthomas:
[ETERNATUS_VMAX_128](https://trello.com/c/byo4O2Iz/562-eternatusvmax128)

src/tcgwars/logic/impl/gen8/DarknessAblaze.groovy
```diff
@@ -2955,7 +2960,94 @@ public enum DarknessAblaze implements LogicCardInfo {
         weakness F
         bwAbility "Infinity Zone", {
           text "If all of your Pokémon in play are [D], you may now have up to 8 [D] Pokémon on your Bench and can’t play any other type of Pokémon. (When this Ability no longer works, discard your Benched Pokémon until there are 5 left.)"
-          actionA {
+          bwAbility "Infinity Zone", {
```
flagrama, line 2963: `bwAbility "Infinity Zone"` already is there.

luongthomas: Thanks, nice catch

flagrama:
Weird timing issue I think. I use Super Scoop Up to pick up my Eternatus VMAX on the bench, then the bench discard selection lets me choose that same Eternatus VMAX, which causes a game error.
game log trace: https://gist.github.com/flagrama/43dbae5134ad7f21b6eef5afbe242ceb

src/tcgwars/logic/impl/gen8/DarknessAblaze.groovy
```diff
@@ -263,6 +263,11 @@ public enum DarknessAblaze implements LogicCardInfo {
     return tcgwars.logic.card.Collection.DARKNESS_ABLAZE;
   }
 
+  @Override
+  public int getCollectionLineNo() {
+    return collectionLineNo;
+  }
+
```
starg09, line 270: This is causing build to fail. Was removed in axpendix rework of collection numbers iirc

```suggestion
```

luongthomas:
> * [ ]  Removing all non-Dark type Pokemon from the field does not (re)activate Eternal Zone.

Can you provide an example?  I've tested removing all non-darkness type Pokemon and I was able to get the Bench size to increase to 8

starg09:
> [ ] Weird timing issue I think. I use Super Scoop Up to pick up my Eternatus VMAX on the bench, then the bench discard selection lets me choose that same Eternatus VMAX, which causes a game error.

Hmm, weird. In my local instance the Eternatus doesn't show up (a blank space is put in its spot though, inside the select). And I'm forced down to a bench of 4.

flagrama:
> Hmm, weird. In my local instance the Eternatus doesn't show up (a blank space is put in its spot though, inside the select). And I'm forced down to a bench of 4.

Sounds to me like you didn't evolve him and just played him straight from hand?

> Can you provide an example? I've tested removing all non-darkness type Pokemon and I was able to get the Bench size to increase to 8

I removed a non-Dark-type Pokémon from the active spot IIRC.

luongthomas:
Testing reopening

starg09:
> Sounds to me like you didn't evolve him and just played him straight from hand?

@flagrama First time for sure, but I think I also tested it via evolving it and scooping it with SSU specifically.

PR #905, opened 2020-08-10
Some more general fixes

starg09:
* _**Psyduck** (Team Up)_ should now use its attack name when locking trainers.
* _**Typhlosion** (Unseen Forces)_ should now work properly, not damage new actives and turn off by stuff like Cessation Crystal.
* _**Spinda** (Legend Maker)_ should should now only activate slightly later, right before the other Pokémon actually attacks (and no longer activating for attacks called without the required energy).
* _**Rocket's Raikou-ex** (Deoxys)_ should no longer crash when attaching an energy to it.
* _**Marnie** (Sword & Shield - Base Set)_ now properly shuffles the players' hands before putting them at the bottom of each deck. Also has custom, clearer log text now.
* Reimplemented _**Underground Expedition** (Celestial Storm)_ in Groovy, fixing the card not allowing a player to reorder the cards that remain in the deck.
* _**Primeape** (Unleashed)_ should no longer crash when using "Bebop Punch".
* _**Pichu Bros.** (POP Series 3)_ should no longer crash when using "Bustle".
* _**Alolan Ninetales-GX** (Guardians Rising)_ should now actually move its damage counters, instead of moving the damage value (making it cause KOs instantly for stuff like Diantha, and also becoming blockable by certain Pokémon.
* _**Diantha** (Forbidden Light)_ now uses checks for KOs similar to Zinnia, Rosa and Lance Prism. Should make it work consistently.
  - These cards really need a new getter similar to `my.lastKnockoutByOpponentDamageTurn`, but not centered on damage. Will be looked at in the future.
* _**Zygarde** (Unified Minds)_ should no longer softlock the game when the player has no bench.


src/tcgwars/logic/impl/gen3/LegendMaker.groovy
```diff
@@ -1066,7 +1066,7 @@ public enum LegendMaker implements LogicCardInfo {
           text "As long as Spinda is your Active Pokémon, whenever your opponent's Basic Pokémon tries to attack, your opponent flips a coin. If tails, that attack does nothing. You can't use more than 1 Pattern Distraction Poké-Body each turn."
           def endTurn = false
           delayedA {
-            before CHECK_ATTACK_REQUIREMENTS, {
+            before PROCESS_ATTACK_EFFECTS, {
```
axpendix, line 1069: as far as I know PROCESS_ATTACK_EFFECTS  step shouldn't be prevented, does it work properly? 

maybe this is better

```          
delayedA (priority: BEFORE_LAST) {
              before CHECK_ATTACK_REQUIREMENTS, {
```

starg09: Fixed below, thanks for the suggestion 👍 (Also helped solve a problem with Cursed Shovel and Spell Tag, though that needs more work in general (alongside Max Phantom-like attacks)

axpendix: @starg09 Good one, I think it'd makes sense to change priority of energy cost check inside `Attack` to `AFTER_FIRST`  later on.

starg09:
All these changes are tested and ready for a rollout.

src/tcgwars/logic/impl/gen7/CelestialStorm.groovy
```diff
@@ -3318,7 +3318,26 @@ public enum CelestialStorm implements LogicCardInfo {
...
+            if (my.deck) {
+              def drawStart = Math.max(0, my.deck.size() - 4)
+              def cardsInBottom = my.deck.size() - drawStart
+              def deckBottom = my.deck.subList(drawStart, my.deck.size())
+              def sel = deckBottom.select(count: 2, "Select 2 cards to put to your hand.").moveTo(supressLog: true, my.hand)
+              def list = deckBottom.getExcludedList(sel)
+              if (cardsInBottom > 3) {
+                list = rearrange(list, "Rearrange the other 2 cards before putting them back at the bottom of the deck")
+                my.deck.setSubList(Math.max(0, drawStart-2), list)
+              }
```
axpendix, line 3333: Assuming it works 😁

src/tcgwars/logic/impl/gen7/ForbiddenLight.groovy
```diff
@@ -2590,13 +2590,22 @@ public enum ForbiddenLight implements LogicCardInfo {
...
+            }
+          }
           onPlay {
             my.discard.select(count : Math.min(2,my.discard.size())).moveTo(my.hand)
           }
           playRequirement{
+            assert my.discard : "You have no cards in your discard pile"
             assert bg.turnCount
-            assert my.lastKnockoutByOpponentDamageTurn == bg.turnCount - 1: "No Pokémon has been Knocked Out during your opponent’s last turn"
-            assert my.knockoutTypesPerTurn.get(bg.turnCount - 1)?.contains(Y) : "The Pokémon Knocked Out was not Fairy"
```
axpendix, line 2599: So sad that this doesn't work as `knockoutTypesPerTurn` structure was specifically created for this card 😅 Maybe it's better to update `CheckFaint` . Nevertheless, it can be postponed if you think there is not enough time for now.

src/tcgwars/logic/impl/gen3/Deoxys.groovy
```diff
@@ -3012,16 +3012,14 @@ public enum Deoxys implements LogicCardInfo {
...
-                if(!self.specialConditions && ef.reason == PLAY_FROM_HAND && ef.card.asEnergyCard().containsType(D))
-                  def oppPlayer = self.owner.opposite
-                  if(oppPlayer){
-                    if (confirm("Make your opponent switch the Defending Pokémon with 1 of your opponent’s Benched Pokémon? (Your opponent chooses the Benched Pokémon to switch)")) {
-                      sw oppPlayer.pbg.active, oppPlayer.pbg.bench.select("Select the new active Pokémon.", oppPlayer)
-                    }
+                def oppPlayer = self.owner.opposite
+                if(!self.specialConditions && ef.reason == PLAY_FROM_HAND && ef.card.asEnergyCard().containsType(D) && oppPlayer.pbg.bench){
+                  if (confirm("Make your opponent switch the Defending Pokémon with 1 of their Benched Pokémon? (Your opponent chooses the Benched Pokémon to switch)")) {
+                    sw oppPlayer.pbg.active, oppPlayer.pbg.bench.select("Select the new active Pokémon.", oppPlayer)
```
axpendix, line 3018: I'd add a `powerUsed()` here

src/tcgwars/logic/impl/gen8/SwordShield.groovy
```diff
@@ -3417,7 +3417,7 @@ public enum SwordShield implements LogicCardInfo {
         def eff
         onPlay {reason->
           eff = delayed {
-            before (KNOCKOUT, self) {
+            before (KNOCKOUT, self) (priority: BEFORE_LAST) {
```
axpendix, line 3420: priority param should go to `delayed` call

src/tcgwars/logic/impl/gen7/ForbiddenLight.groovy
starg09: Yeah, there are many cards that are heavily played and have this kind of fix. It can definitely be moved over those, but would rather leave it for a separate PR 😅

src/tcgwars/logic/impl/gen7/CelestialStorm.groovy
starg09: Yup, tested and seemed to work well for all cases👍

src/tcgwars/logic/impl/gen8/SwordShield.groovy
starg09: Whoops.

PR #907, opened 2020-08-13
Darkness ablaze fixes

flagrama:
Many small fixes. Almost all very minor. Also made sure you can't play Pokemon Breeder's Nurturing without a Pokemon that can evolve in play.

### DIFF
--- a/src/tcgwars/logic/impl/gen8/DarknessAblaze.groovy
+++ b/src/tcgwars/logic/impl/gen8/DarknessAblaze.groovy
@@ -3226,6 +3226,9 @@ public enum DarknessAblaze implements LogicCardInfo {
               prevent()
             }
           }
+          onActivate {
+            clearSpecialCondition(self, SRC_ABILITY)
+          }
         }
         move "Revenge Stomp", {
           text "120+ damage. If any of your Benched Pokémon have any damage counters on them, this attack does 120 more damage."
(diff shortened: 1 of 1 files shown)


src/tcgwars/logic/impl/gen8/DarknessAblaze.groovy
```diff
@@ -3400,7 +3424,10 @@ public enum DarknessAblaze implements LogicCardInfo {
             assert my.deck : "You have no cards left in your deck"
           }
           onAttack {
-            my.deck.search("Put Energy Card into your hand",cardTypeFilter(ENERGY)).showToOpponent("Opponent's chosen Energy card").moveTo(my.hand)
+            if (my.deck) {
```
luongthomas, line 3427: Won't this always be true because of the `attackRequirement`?

src/tcgwars/logic/impl/gen8/DarknessAblaze.groovy
```diff
@@ -2854,8 +2869,10 @@ public enum DarknessAblaze implements LogicCardInfo {
             assert my.deck : "Your deck is empty"
           }
           onAttack {
-            my.deck.search(max:2,"Select up to 2 cards",{true}).moveTo(hidden:true,my.hand)
-            shuffleDeck()
+            if (my.deck) {
```
luongthomas, line 2872: Won't this also always be true because of the assert in `attackRequirement`?

src/tcgwars/logic/impl/gen8/DarknessAblaze.groovy
luongthomas: Ah nevermind I found it.  You were considering the case where the attack was copied

flagrama: Not if the attack is copied by another attack/ability.

src/tcgwars/logic/impl/gen8/DarknessAblaze.groovy
flagrama: Also not if the attack is copied by another attack/ability.

PR #908, opened 2020-08-13
Implement Eternatus VMAX

luongthomas:
This is similar in spirit to #903 but with some additional changes to address some of the issues @flagrama encountered during testing.

There are now benchTrigger checks that occur `DEVOLVE` and `EVOLVE`.  

<sub><sup>Thank you for the help flagrama with Git... but I'm ashamed to say I've lost the battle against the git merge conflicts</sup></sub>

### DIFF
--- a/src/tcgwars/logic/impl/gen8/DarknessAblaze.groovy
+++ b/src/tcgwars/logic/impl/gen8/DarknessAblaze.groovy
@@ -2980,8 +2980,8 @@ public enum DarknessAblaze implements LogicCardInfo {
           def isOwnerPokemonAllDark = {
             return !self.owner.pbg.all.find {
               it.cards().first() &&
-                it.cards.first().cardTypes.contains(POKEMON) &&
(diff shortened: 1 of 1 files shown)


luongthomas:
> Weird timing issue I think. I use Super Scoop Up to pick up my Eternatus VMAX on the bench, then the bench discard selection lets me choose that same Eternatus VMAX, which causes a game error.
> game log trace: https://gist.github.com/flagrama/43dbae5134ad7f21b6eef5afbe242ceb


I'm able to reproduce this now and I'll keep trying to look for a fix.  The other issues you brought up _should_ be resolved now though

flagrama:
Evosoda and Rare Candy from Sandstorm are able to evolve a D-type (most Sneasels) to a non-D-type (CEC Weavile). I don't think this should be permitted, though I'm not certain.

flagrama:
Should verify somehow that that PR is the correct behaviour before merging.

PR #913, opened 2020-08-14
Fix daa fossil & scizor

flagrama:


src/tcgwars/logic/impl/gen8/DarknessAblaze.groovy
```diff
@@ -3987,7 +3987,7 @@ public enum DarknessAblaze implements LogicCardInfo {
         text "Play this card as if it were a 70-HP Basic [C] Pokémon. At any time during your turn, you may discard this card from play. This card can’t be affected by any Special Conditions, and it can’t retreat."
         onPlay {
           Card pokemonCard, trainerCard = thisCard
-          pokemonCard = basic (new CustomCardInfo(RARE_FOSSIL_175).setCardTypes(BASIC, POKEMON), hp:HP070, type:COLORLESS, retreatCost:0) {
+          pokemonCard = basic (new CustomCardInfo(this[thisCard.enumName] as CardInfo).setCardTypes(BASIC, POKEMON), hp:HP070, type:COLORLESS, retreatCost:0) {
```
axpendix, line 3990: How does this[thisCard.enumName] work? I am curious

flagrama: The square brackets call the getAt() method on the enum, which is a method [every Groovy object has](http://docs.groovy-lang.org/next/html/groovy-jdk/java/lang/Object.html#getAt(java.lang.String)) and Groovy [implicitly coerces a String to an Enum value](https://docs.groovy-lang.org/latest/html/documentation/#_string_to_enum_coercion) where applicable.

This is equivalent to calling `DarknessAblaze.RARE_FOSSIL_167` but with the enum class this code is contained within and the string representation of the card enum value being coerced to the enum value making it less sensitive to changes to the enum value.

PR #919, opened 2020-08-15
Fix Dracovish Primal Law not working for all evolution from hand methods

flagrama:


flagrama:
Play_Evolution and Evolve_Standard were already blocked, but you could still evolve using many abilities that let you evolve a Pokémon from hand (Ditto Prism Star's Almighty DNA, Eevee-GX's DNA Ascension, Meganium (LOT)'s Quick Ripening Herb. This seemed to prevent those fine.

Perhaps the actual solution is modifying the `evolve` method in TcgStatics to call an `EvolveStandard` instead of an `Evolve`?

flagrama:
@axpendix I'm still curious about this. Is blocking Evolve fine since it seems to work as expected? Or should the `evolve` static, and perhaps several other instances where Evolve is used instead of EvolveStandard be changed to use EvolveStandard so that only EvolveStandard gets blocked, and not the Evolve effect?

EvolveStandard's only purpose seems to be to call Evolve, so even having both doesn't make complete sense to me. Is it for instances where a Pokémon should only be affected by something if it successfully evolves, so you use EvolveStandard so you can prevent an evolution without also activating the other before effects that should be activated on successful evolve? However, that seems to work fine when blocking the Evolve effect.

If Evolve is meant for any non-standard evolution however, I don't see how you can avoid blocking it, so I'd really like to know what the issue with blocking it is if there is a known one.

PR #920, opened 2020-08-15
Some Fixes

starg09:
* _**Latias δ**, **Latios δ** and **Pidgeot δ** (all from Holon Phantoms)_ should now properly refresh their locks when a new Pokémon is put into play, or a card removes one from Play.
* _**Poké Ball** (Rebel Clash)_ should now have the correct name displayed in screen, instead of "PokeBall".
* Marked some cards from NG sets (Groovy reimplementations of legacy-code sets), as to know they're being used as reprints somewhere else.
* Redirected _**Professor's Research** (Sword & Shield)_ so it uses XY's Professor Sycamore as a base, instead of the BaseSetNG's Professor Oak.
* Implemented _**Mew** (Wizards Black Star Promo 47)_ in its Groovy form, which the legacy-code print will now point to as to fix some bugs with Neutral Shield.


starg09:
Yup, will look for a better solution at some point (alongside the other couple cards already using this). Merging like this for now though.

PR #921, opened 2020-08-16
Scoop up block

flagrama:
The TCGStatics change may need a second revision, this was just done quickly to reduce issues the ability blocking a RemoveFromPlay in a RemovePCS was running into.

Mr Mime's Scoop-Up Block should work in nearly every case that discarding doesn't happen _and_ need to be prevented by the ability. That remaining case is handled using special cases where the scoop up trainers are concerned, but depending on a future ruling, may or may not need changed to be more general somehow for say Wishiwashi SUM's ability.

Lastly, updates Recycle Energy (UNM copy only) and U-Turn Board to go prevent the MoveCard to discard effect if a MoveCard to hand effect succeeds instead of going to discard, then being moved to hand.

### DIFF
--- a/src/tcgwars/logic/groovy/TcgStatics.groovy
+++ b/src/tcgwars/logic/groovy/TcgStatics.groovy
@@ -912,6 +912,10 @@ class TcgStatics {
     }
   }
 
+  static boolean hasThetaStop(PokemonCardSet pcs) {
+    return pcs.lastAbilities.find {it.key instanceof AncientTrait && it.key.name.contains("Stop")}
+  }
+
   static omega_double(Object delegate){
     delegate.ancientTrait "θ Double", {
       text "This Pokémon may have up to 2 Pokémon Tool cards attached to it."
@@ -1517,7 +1521,7 @@ class TcgStatics {
       if (delegate.thisObject.cardTypes.is(ENERGY)) source = SRC_SPENERGY
     }
     if (params.discard == null) params.discard = true
-    if (bg.em().retrieveObject("ScoopUpBlock_Count$target.owner.opposite") && target.numberOfDamageCounters) {
+    if (bg.em().retrieveObject("ScoopUpBlock_Count$target.owner.opposite") && target.numberOfDamageCounters && !hasThetaStop(target)) {
       bc "Scoop-Up Block prevents $delegate.thisObject.name's effect."
       return
     }
--- a/src/tcgwars/logic/groovy/TcgStatics.groovy
+++ b/src/tcgwars/logic/groovy/TcgStatics.groovy
@@ -467,47 +467,35 @@ class TcgStatics {
   static removePCS(PokemonCardSet pcs){
     if(my.active==pcs){
       def r=new RemoveFromPlay(pcs,null)
-      def failed = r.run(bg)
-      if (!failed) {
-        if (my.bench.isEmpty()){
-          my.active=null
-          bg.game.endGame(opp.owner, WinCondition.NOPOKEMON)
-          return
-        }
-        sw(null, my.bench.select("New active pokemon"))
+      r.run(bg)
+      if (my.bench.isEmpty()){
+        my.active=null
+        bg.game.endGame(opp.owner, WinCondition.NOPOKEMON)
+        return
       }
+      sw ( null, my.bench.select("New active pokemon"))
       //my.bench.remove(pcs)
     }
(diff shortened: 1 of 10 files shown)


src/tcgwars/logic/groovy/TcgStatics.groovy
```diff
@@ -1500,4 +1504,54 @@ class TcgStatics {
     }
   }
 
+  /**
+   * Scoop up PokemonCardSets if not blocked by Scoop-Up Block
+   *
+   * @param params Optional settings map
+   *  + discard: Indicates if any cards should be discarded. Default: true
+   *  + only: A card or list of cards to move to hand instead of discard.
```
axpendix, line 1512: This parameter design looks complicated to me. 

It'd look simpler like this;

- all: boolean - if true, scoops up all cards, else all pokemon cards. discards the rest.
- only: cardlist - only scoops up them, discards the rest.

flagrama: Like this? https://github.com/axpendix/tcgone-engine-contrib/pull/921/commits/c8e17c35fd8d48e16b9975d37abc46a754294edb

axpendix: Don't you think 'all' should be false by default? It would follow default variable conventions.

flagrama: I have conflicting thoughts on this. You're right, it would. It was something that was bothering me when I was calling it discard instead. Particularly because the majority of Pokémon scoop up effects (including ones in the main repo) seem to scoop up all of the cards and scooping up only the Pokémon seems to be a rarer case. I feel that making the most common case the default, and using the params for progressively different behaviour makes it easier to implement any future effects IMO.

If you feel that the default variable convention is more important than that though, I can make that change.

axpendix: Agree with the common case. Then the parameter should better be 'pokemonOnly' ?

src/tcgwars/logic/groovy/TcgStatics.groovy
```diff
@@ -1500,4 +1504,53 @@ class TcgStatics {
...
+        toDiscard = pokemonList.getExcludedList(params.only as CardList)
+        target.owner.pbg.discard.addAll(toDiscard)
+        target.owner.pbg.hand.addAll(pokemonList.getExcludedList(toDiscard as CardList))
+        toDiscard = otherList.getExcludedList(params.only as CardList).discard()
+        otherList.getExcludedList(toDiscard).moveTo(target.owner.pbg.hand)
+      }
+      else {
+        target.owner.pbg.hand.addAll(pokemonList)
+        otherList.moveTo(target.owner.pbg.hand)
+      }
```
axpendix, line 1551: I wasn't too happy with the readability of the internal logic. Do you think this would work? 

```suggestion

CardList toHand
if(params.only) {
  if (params.only instanceof Card) toHand = new CardList(params.only)
  else if (params.only instanceof CardList) toHand = params.only as CardList
  else throw new IllegalArgumentException("scoopUpPokemon() params.only=${params.only} type not supported")
} else if(params.pokemonOnly) {
  toHand = target.cards.filterByType(POKEMON)
} else {
  toHand = new CardList(target.cards)
}
CardList toDiscard = target.cards.getExcludedList(toHand)

bc "Scooped up ${toHand}"

CardList toHand2 = toHand.filterByType(POKEMON)
target.owner.pbg.hand.addAll(toHand2)
toHand.getExcludedList(toHand2).moveTo(target.owner.pbg.hand)

CardList toDiscard2 = toDiscard.filterByType(POKEMON)
target.owner.pbg.discard.addAll(toDiscard2)
toDiscard.getExcludedList(toDiscard2).discard()
```

flagrama: This seems to work fine. I want to test a couple more things with it, but will likely commit that suggestion.

axpendix (approved):
Looks good to me

axpendix:
@flagrama On a seperate note, would a new `EffectType.SCOOP_UP` make more sense for implementation of "Scoop-Up Block"?

flagrama:
It may. At the same time, it seems overkill to me for essentially a single usage of `prevent()` though. If there were several scoop-up block type effects that I knew of, then I'd certainly prefer the new EffectType. If there are just some that I'm not aware of, then even more so.

PR #922, opened 2020-08-16
Fix Arctozolt and Jolteon

flagrama:


src/tcgwars/logic/impl/gen7/HiddenFates.groovy
```diff
@@ -9,6 +9,7 @@ import static tcgwars.logic.card.HP.*;
 import static tcgwars.logic.card.Resistance.ResistanceType.*
 import static tcgwars.logic.card.Type.*
 import static tcgwars.logic.effect.EffectType.*
+import static tcgwars.logic.effect.Source.SRC_ABILITY
```
starg09, line 12: Maybe import the enum here? If needed at all.

PR #923, opened 2020-08-18
New getCardTypes getter

starg09:
Needed for a core update


src/tcgwars/logic/impl/gen3/DragonFrontiers.groovy
```diff
@@ -204,44 +204,16 @@ public enum DragonFrontiers implements LogicCardInfo {
...
-                  list.each { it.cardTypes.remove(DELTA) }
-                  null
-                })
-                null
-              } else {
-                it - 1
-              }
-            })
+          getterA(GET_CARD_TYPES, BEFORE_LAST) {holder ->
+            if (holder.effect.target.player == self.owner)
```
axpendix, line 208: Add basic and evolution condition please

starg09: > Add basic and evolution condition please

Cards prior to Diamond & Pearl that look for "Basic Pokémon or Evolution cards" are meant to search for any Pokémon (including, for example, LV.X which are neither)

> Q. Can I use Celio's Network to search my deck for a Pokémon LV.X?
> 
> A. Yes, you can. Earlier card references of "Basic Pokémon or Evolution card" refers to Pokémon in general, which includes Pokémon LV.X. (May 10, 2007 PUI Rules Team)

Actually need to go through older cards and correct that behaviour. The one card type that could be affected in all of this are Neo/e-Card Baby Pokémon, but even then those count as Basic Pokémon (and cards that don't allow babies explicitely exclude them).

starg09:
Would like to confirm what I replied in the conversation above being ok before merging, as this is required for the core getter PR to be merged 😅

src/tcgwars/logic/impl/gen3/DragonFrontiers.groovy
axpendix: I mean this condition :) 

```suggestion
            if (holder.effect.target.player == self.owner && holder.object.is(POKEMON))
```

axpendix: Otherwise all cards (including energy and trainer) will be DELTA due to how this getter works.

axpendix (approved):

src/tcgwars/logic/impl/gen3/DragonFrontiers.groovy
starg09: Oh. Sorry, didn't notice no check was being done my bad on that!

PR #924, opened 2020-08-19
Some more fixes

starg09:
* Fixed _**Timer Ball** (Guardians Rising)_ shuffling the deck even when getting double tails.
* Fixed _**Pokemon Research Lab** (Unified Minds)_ being usable even when out of space in one's bench.
* Fixed _**Metal Core Barrier** (Unbroken Bonds)_ being discarded even after cards like _**Noctowl** (Sword & Shield)_ shuffled it into the deck,
* Made _**Dark City** (Unified Minds)_ override other retreat cost modifiers, such as _**Absol** (Team Up)_.

Also included a small line change not committed in time into #923.


src/tcgwars/logic/impl/gen7/UnbrokenBonds.groovy
```diff
@@ -4314,23 +4314,31 @@ public enum UnbrokenBonds implements LogicCardInfo {
         return pokemonTool (this) {
           text "If this card is attached to 1 of your Pokémon, discard it at the end of your opponent's turn." +
             "The [M] Pokémon this card is attached to takes 70 less damage from your opponent's attacks (after applying Weakness and Resistance)."
-          def eff
+          def eff1, eff2
           onPlay {reason->
-            eff=delayed{
+            eff1 = delayed {
               before APPLY_ATTACK_DAMAGES,{
```
luongthomas, line 4320: Does this reduce the 70 hp after applying weakness and resistance?  Just curious

starg09: Yeah, `before APPLY_ATTACK_DAMAGES` reduces after W/R and `before APPLY_ATTACK_DAMAGES` and `after PROCESS_ATTACK_EFFECTS` would reduce before it.

luongthomas: Huh interesting.  Thanks for clarifying

PR #931, opened 2020-08-28
New GET_GIVEN_PRIZES getter, fixes for cards with Prize Increase/Denial

starg09:
These cards have for the most part used direct calls to TakePrize for the extra prizes (or BlockingEffect for denying them). This has caused a number of issues in the past, some persisting to this day (main one leading to this situation: Galarian Cursola with a Vessel of Life Shedinja would deny the prize caused by Cursola's Ability, with no clean way of avoiding it).

The addition of this getter modifies the amount drawn during Knockout itself. In addition, the initial check now properly handles Pokémon LEGEND (Learned this just recently, but only the ones with pairs give two prizes! Both "Lugia LEGEND" and "Ho-Oh LEGEND" only give 1 prize, something that is now covered in code)

Depends on axpendix/tcgone#64


axpendix (approved):
Looks great

PR #932, opened 2020-08-28
Some card fixes

starg09:


starg09:
> Looks good

Oh, that was fast, thanks 😅

PR #934, opened 2020-08-30
Remove duplicates from discard if scooped after knockout

flagrama:


flagrama:

axpendix (approved):
Approved but I wonder if there could be a cleaner solution (:

flagrama:
You're probably right. I'll leave a FIXME on that comment so that it can be revisited in the future.

PR #938, opened 2020-09-02
Update short name for some Collections, reorder TRAINER and ENERGY in CardType

starg09:
* Updated the short name of some Collections/Sets:
  - Furious Fists (FUF -> FFI) and Primal Clash (PCL -> PRC) now use the official shortname used in other sites, including the official ones.
  - BW Black Star Promos (BLWP -> BWP) and SWSH Black Star Promos (SSHP -> SWSHP) are now consistent with other Black Star Promos such as the XY and SM ones, improving consistency across them.
  - Shiny Vault doesn't have an official short name. The set code used is the one from the original set so a custom, clearer one like the one in RUMBLE is used now (SV would've been preferable, but it's taken by Supreme Victors; SHINYV guarantees it won't conflict with any future set, hopefully)

* Reordered the TRAINER and ENERGY top CardTypes
  - This is needed for some Deck Import/Export changes to work consistently with other sites.
  - Would need confirmation on this not breaking something else in the process.

### DIFF
--- a/src/tcgwars/logic/card/CardType.java
+++ b/src/tcgwars/logic/card/CardType.java
@@ -14,8 +14,8 @@ public enum CardType {
   TOP(-100),
 
   POKEMON(10),
-  ENERGY(20),
-  TRAINER(30),
+  TRAINER(20),
+  ENERGY(30),
 
   MEGA_POKEMON(-6, "MEGA"),
   BREAK(-5, "BREAK"),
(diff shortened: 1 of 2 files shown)


axpendix (commented):
Would this change break any deck imports in `txt` form?

axpendix (approved):

starg09:
> Would this change break any deck imports in `txt` form?

Will check in a bit. But when running UI no crash seemed to happen (unlike when renaming SM to SUM), so I'd assume no.

starg09:
That last comment was about set name changes.

On the energy/trainer reorder, I see txt lists using both orders from before this change. Specifically, the imported e-Card decks seem to have old order and still import correctly after this change.

PR #939, opened 2020-09-03
Update Devolve Method to Handle Lv.X

luongthomas:
This PR introduces some changes to the `devolve()` method in `TcgStatics` so that it allows the caller to describ the destination of where the devolved card should be moved to.

Doing so will allow Lv.X cards to be devolved correctly and have the Lv.X card be moved to the same location as the highest Non Lv.X card.

Screen Shot 2020-09-03 at 2 08 28 PM


flagrama:
I still request that the moveTo is checked to ensure it wasn't blocked (primarily by scoop-up block)

luongthomas:
> I still request that the MoveCard is checked to ensure it wasn't blocked (primarily by scoop-up block)

Just to clarify the behavior, if MoveCard is blocked, we shouldn't allow the devolve to occur.  Is this correct?

flagrama:
Correct.

src/tcgwars/logic/groovy/TcgStatics.groovy
```diff
@@ -505,17 +505,24 @@ class TcgStatics {
...
     bg().em().run(new Devolve(pcs));
+    bg().em().run(new MoveCard(card, newLocation));
     if (all.contains(pcs)) { //not dead yet.
+      bc "$card Devolved"
       bg().em().run(new RemoveFromPlay(pcs, new CardList(card)));
       bg().em().run(new CantEvolve(pcs, bg().getTurnCount()));
-//			bg().em().run(new ActivateAbilities(pcs.getTopPokemonCard(), pcs, ActivationReason.OTHER));
+
+      // Remove the highest stage non-level up card if devolved from level-up
+      if (card.cardTypes.is(LEVEL_UP) && pcs.cards.filterByType(POKEMON).size() > 1) {
```
starg09, line 522: Not sure of where it'd need to go, but a special check for Actual Pokémon cards should be implemented for stuff like this and `isEvolution()` from PCS.java, as to cover cards like "Ditto (Boundaries Crossed)" (I feel there was some other card like this, but doesn't come to mind right now)

src/tcgwars/logic/impl/pokemod/PokemodBaseSet.groovy
```diff
@@ -1619,8 +1619,7 @@ public enum PokemodBaseSet implements LogicCardInfo {
           def stage = pkmn.size()>1 ? pkmn.select("Choose stage to devolve to").first() : pkmn.first()
```
starg09, line 1619: Same as the other comment, Also, don't think lv.xs are planned for pokemod (yet) but might be worth adding checks for it here (and def would need to be added in BaseSetNG for Unlimited)

starg09 (commented):
Some todo details, but the PR itself looks great 👍

luongthomas:
> Correct.

I've modified `devolve` so that it prevents devolution from occurring when MoveCard is blocked.

flagrama:
And can you test one thing for me? Dialga LOT's attack "Turn Back Time" was giving me the Dialga as the target prior to this PR. If that is still the case here, I think it's best to target the pcs directly in the devolve static.

luongthomas:
> And can you test one thing for me? Dialga LOT's attack "Turn Back Time" was giving me the Dialga as the target when trying to block the devolution prior to this PR. If that is still the case here, I think it's best to target the pcs directly in the devolve static.

Screen Shot 2020-09-03 at 4 49 21 PM


It seems to behave the way I expect it to here.  Is this how you wanted it to be tested?

flagrama:
Yup, that looks pretty good to me! Thank you!

axpendix (approved):

PR #944, opened 2020-09-11
Fix at least devolution not using the proper target

flagrama:
The attempt to retain the targeted Pokémon through effects didn't work properly. It would get confused if, say, you devolved a undamaged Pokémon, then a damaged one with your opponent having Scoop-Up Block active. It would not block the second one properly and allow you to scoop it up anyway.

Instead, I use the tried and true method of "just find the card in one of the opponent's PCSes, and use that PCS as the target"


flagrama:

flagrama:
I'm particularly curious if anyone knows of a card that uses the same MoveCard to move cards from multiple PCS at once, cause that would cause issues with this.

starg09 (approved):
> I'm particularly curious if anyone knows of a card that uses the same MoveCard to move cards from multiple PCS at once, cause that would cause issues with this.

The couple cases I could think of that might do this handle it one by one, so no issue should happen there.

I'd be on the lookout for potential reports after rolling this, but should work fine based on that (and a quick lookaround for potential uses of moveTo that worked like this)

PR #946, opened 2020-09-14
Even more scoop up

flagrama:
Even some Gen4 stuff because I'm sick of this refactor and never want to have to look at scoop up again :laughing:

### DIFF
--- a/src/tcgwars/logic/groovy/TcgStatics.groovy
+++ b/src/tcgwars/logic/groovy/TcgStatics.groovy
@@ -1535,13 +1535,15 @@ class TcgStatics {
    * Scoop up PokemonCardSets if not blocked by Scoop-Up Block
    *
    * @param params Optional settings map
-   *  + pokemonOnly: boolean - if true, scoops up all cards, else all pokemon cards. discards the rest.
+   *  + pokemonOnly: boolean - if true, scoops up all pokemon cards and discards the rest.
    *  + only: CardList - only scoops up them, discards the rest.
    * @param target PokemonCardSet to work on
    * @param delegate Effect delegate used to determine most sources automatically, and to get the card name for the Scoop-Up Block message
    * @param source Allows you to specify the source of the scoop up. Use intended manually setting SRC_ABILITY.
+   *
+   * @return boolean suceeded
    */
-  static void scoopUpPokemon(params=[:], PokemonCardSet target, Object delegate, Source source=null) {
+  static boolean scoopUpPokemon(params=[:], PokemonCardSet target, Object delegate, Source source=null) {
     if (source == null) {
       if (delegate.thisObject.cardTypes.is(TRAINER)) source = TRAINER_CARD
       if (delegate.thisObject.cardTypes.is(POKEMON)) source = ATTACK
@@ -1549,7 +1551,7 @@ class TcgStatics {
     }
     if (bg.em().retrieveObject("ScoopUpBlock_Count$target.owner.opposite") && target.numberOfDamageCounters && !hasThetaStop(target)) {
       bc "Scoop-Up Block prevents $delegate.thisObject.name's effect."
-      return
+      return false
     }
     targeted(target, source) {
       CardList toHand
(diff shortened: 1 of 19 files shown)


flagrama:

src/tcgwars/logic/groovy/TcgStatics.groovy
```diff
@@ -1577,6 +1579,7 @@ class TcgStatics {
       target.owner.pbg.discard.addAll(toDiscard2)
       toDiscard.getExcludedList(toDiscard2).discard()
       removePCS(target)
+      return true
     }
   }
```
starg09, line 1584: Not sure if relevant, but wouldn't a `return false` be needed after the targeted, in case it gets blocked?

src/tcgwars/logic/impl/gen4/DiamondPearlPromos.groovy
```diff
@@ -629,6 +630,13 @@ public enum DiamondPearlPromos implements LogicCardInfo {
           pokePower "Leap Away", {
             text "Once during your turn , if Toxicroak is your Active Pok�mon, you may flip a coin. If heads, return Toxicroak and all cards attached to it to your hand. This power can�t be used if Toxicroak is affected by a Special Condition."
             actionA {
+              assert self.active : "$self is not your Active Pokémon"
+              checkLastTurn()
+              checkNoSPC()
+              powerUsed()
```
starg09, line 636: Minor suggestion, but since implementing: Would maybe add a confirm when it's the only pokemon in play before using the power, similar to Super Scoop Up.

Might be worth its own PR though, maybe adding a static assert for all cards like this, that checks for no bench and in that case does the confirm.

src/tcgwars/logic/impl/gen4/HeartgoldSoulsilver.groovy
```diff
@@ -274,8 +274,7 @@ public enum HeartgoldSoulsilver implements LogicCardInfo {
             }
```
starg09, line 274: Line above, should be usable with no bench. See: review comment on Toxicroak.

src/tcgwars/logic/impl/gen4/Platinum.groovy
```diff
@@ -2660,9 +2663,13 @@ public enum Platinum implements LogicCardInfo {
       case POKE_TURN_118:
         return itemCard (this) {
           text "Return 1 of your Pokémon SP and all cards attached to it to your hand."
+          // TODO: Check for Pokémon SP in onPlay and playRequirement
           onPlay {
+            def pcs = my.bench.findAll { true }.select("Which Pokémon SP to return to hand?")
+            scoopUpPokemon(pcs, delegate)
           }
           playRequirement{
+            my.bench.any { true }
```
starg09, line 2672: Can be played on your only Pokémon in play. See: Super Scoop Up, review comment on Toxicroak.

src/tcgwars/logic/groovy/TcgStatics.groovy
flagrama: Default value of a boolean is false, so if it reaches the end of the method it returns false. I assume the return inside the targeted will not be run if targeted is blocked. Is that incorrect?

src/tcgwars/logic/impl/gen4/DiamondPearlPromos.groovy
flagrama: I'd prefer just getting these to work "good enough" for testing purposes. Primarily just don't want to have to review literally every Gen 4 PR just in case it needs to use scoop up.

src/tcgwars/logic/groovy/TcgStatics.groovy
flagrama: I did it just in case either way.

src/tcgwars/logic/impl/gen4/DiamondPearlPromos.groovy
flagrama: The rest of these should all be part of said scoop-up check PR. The existing ones, and the ones I added, avoid the issue by just not allowing the move be used. I don't think it's closely related enough to this PR to handle here.

starg09 (approved):

PR #947, opened 2020-09-15
Update Deckbuilder to Accomodate Lv.X Limit

luongthomas:
This PR is tied to https://github.com/axpendix/tcgone/pull/82 and https://github.com/axpendix/tcgone/pull/108

But also includes some changes to correct the class of some of the Lv.X cards from `evolution` to `levelUp`.

### DIFF
--- a/src/tcgwars/logic/util/PokemonCardSet.java
+++ b/src/tcgwars/logic/util/PokemonCardSet.java
@@ -285,6 +285,8 @@ public class PokemonCardSet implements PokemonStack, Serializable {
     }
   }
 
+  public String getWholeName() { return getTopPokemonCard().getWholeName(); }
+
   public boolean isActive() {
     return TcgStatics.bg().getPBG(owner).getActive() == this;
   }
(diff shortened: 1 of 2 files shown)


src/tcgwars/logic/util/PokemonCardSet.java
```diff
@@ -285,6 +285,8 @@ public String getName() {
     }
   }
 
+  public String getWholeName() { return getTopPokemonCard().getWholeName(); }
```
starg09, line 288: For the method right above this one, maybe it could be simplified into using `getName()` since that'd make the replace already?

starg09: Also not sure why that try catch is exactly needed, but if it is would it also be needed in the Card implementation of `getName()`?

luongthomas: I've removed the extra string stripping from the `getName` method above this one and also have a try/except block in `getWholeName`

starg09 (approved):

axpendix (approved):

PR #883, opened 2020-07-20
Add SWSH Promos 033-042 + Updates for Darkness Ablaze based on "official" set list

starg09:
Leaving as a draft, but this should in theory cover the main differences regarding ordering, numbering, and trainer/energy name changes. Some abilities got updated, but some may still need updating. And a similar update needs to be done on [carddb](https://github.com/tcgone/carddb/).

Some coded-and-tested cards were removed, which can be taken from here whenever SSH3.5 is worked on (e.g. Gardevoir V/VMAX). Eevee and Flareon are now added as SWSH Promos (alongside the 4 prerelease cards, and some reprints).

Also updated the label in `Collection.java`, as to use the official one.

Source for changes: https://www.pokebeach.com/2020/07/complete-darkness-ablaze-english-set-list

PR #948, opened 2020-09-17
Add Initial Champions Path Groovy Template

luongthomas:
This PR adds the Groovy template for the Champions Path set which is releasing on September 25th 2020.


src/tcgwars/logic/impl/gen8/ChampionsPath.groovy
```diff
@@ -0,0 +1,1073 @@
...
+      return copy (PIERS_58, this);
+      case CHARIZARD_V_79:
+      return copy (DarknessAblaze.CHARIZARD_V_19, this);
+      case STRANGE_CANNED_FOOD_80:
+      return copy (STRANGE_CANNED_FOOD_66, this);
+      default:
+      return null;
+    }
+  }
+}
```
flagrama, line 1073: Empty newline at end please?

flagrama (commented):
I don't have the time to double check that everything matches the relevant cards, but looks pretty good to me.

As a reminder when implementing, some of these cards already have code that I believe has been tested already. https://github.com/axpendix/tcgone-engine-contrib/pull/883/files

Should speed up dev on anything we thought was going to be in DAA but wasn't.

src/tcgwars/logic/impl/gen8/ChampionsPath.groovy
luongthomas: You got it 👍

PR #950, opened 2020-09-17
Correct return value of scoopUpPokemon static

flagrama:
Returning in a closure returns the value of that closure. Need to return the result of the closure instead. Since run returns `false` for success, need to invert it as well.


flagrama:

luongthomas (approved):

PR #951, opened 2020-09-17
Implement Pluspower (Base Set) in groovy

starg09:
Fixes a duplication bug.

### DIFF
--- a/src/tcgwars/logic/impl/gen1/BaseSetNG.groovy
+++ b/src/tcgwars/logic/impl/gen1/BaseSetNG.groovy
@@ -1943,7 +1943,45 @@ public enum BaseSetNG implements LogicCardInfo {
           }
         }
       case PLUSPOWER:
-        return copy(Unleashed.PLUSPOWER_80, this);
+        // This print increases after W/R, do not use it as copy for DP-era prints that do it before W/R (and affects both actives, while this one only affects the defending Pokémon).
+        return basicTrainer (this) {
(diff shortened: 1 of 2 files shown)


src/tcgwars/logic/impl/gen1/BaseSetNG.groovy
```diff
@@ -1943,7 +1943,45 @@ public enum BaseSetNG implements LogicCardInfo {
...
+                }
+              }
+              before BETWEEN_TURNS, {
+                discard thisCard
+              }
+              after REMOVE_FROM_PLAY, pcs, null, {
+                if(ef.removedCards.contains(thisCard)) {
+                  eff.unregister()
+                }
+              }
```
flagrama, line 1981: Does this not work if done in a `onRemoveFromPlay` closure?

src/tcgwars/logic/impl/gen1/BaseSetNG.groovy
```diff
@@ -1943,7 +1943,45 @@ public enum BaseSetNG implements LogicCardInfo {
...
+                      it.dmg += hp(10)
+                    }
+                  }
+                }
+              }
+              after DISCARD, {
+                if(ef.card == thisCard){
+                  eff.unregister()
+                }
+              }
```
flagrama, line 1973: Discard should inherently call RemoveFromPlay since it's pretty much just a MoveCard. Is this definitely necessary?

PR #957, opened 2020-09-19
Champions path

flagrama:
Added documentation to `attachEnergyFrom` and updated some documentation in TcgStatics because it was hard to read in IDEA's intellisense-like feature. Would like to find a way to actually display map keys in said UI, but knowing that it is a map should be enough to look up what is needed.

Implemented the rest of the cards in the set. Seems like another basic and boring set like Hidden Fates, so wasn't too difficult 🤣 

Tried to specify copies a bit better. Was confusing which ones were actually in the current file and which ones were actually in other files. At some point we need to make sure we don't have any duplicate code for reprints. e.g. Some Poké Ball cards copy from FireRedLeafGreen, some copy from EmergingPowers.

### DIFF
--- a/src/tcgwars/logic/impl/gen8/ChampionsPath.groovy
+++ b/src/tcgwars/logic/impl/gen8/ChampionsPath.groovy
@@ -1,4 +1,8 @@
-package tcgwars.logic.impl.gen8;
+package tcgwars.logic.impl.gen8
+
+import tcgwars.logic.impl.gen3.FireRedLeafGreen
+import tcgwars.logic.impl.gen5.BlackWhite
+import tcgwars.logic.impl.gen5.EmergingPowers;
 
 import static tcgwars.logic.card.HP.*;
 import static tcgwars.logic.card.Type.*;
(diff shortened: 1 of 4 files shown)


src/tcgwars/logic/impl/gen8/ChampionsPath.groovy
```diff
@@ -218,31 +222,32 @@ public enum ChampionsPath implements LogicCardInfo {
...
           text "80 damage. Your opponent’s Active Pokemon is now Poisoned."
           energyCost G
-          attackRequirement {}
           onAttack {
             damage 80
+            apply POISONED
           }
         }
       };
       case ELDEGOSS_5:
```
luongthomas, line 231: Could you modify the ID here to be `ELDEGOSS_V_5`, I must've missed it here

flagrama, line 231:
```suggestion
      case ELDEGOSS_V_5:
```

luongthomas (approved):

PR #960, opened 2020-09-21
thisAbility.name -> thisAbility

starg09:
toString method was added by @flagrama so this is no longer needed.

### DIFF
--- a/src/tcgwars/logic/impl/gen4/MysteriousTreasures.groovy
+++ b/src/tcgwars/logic/impl/gen4/MysteriousTreasures.groovy
@@ -1234,7 +1234,7 @@ public enum MysteriousTreasures implements LogicCardInfo {
               checkLastTurn()
               checkNoSPC()
               powerUsed()
-              def chosenPCS = opp.all.select("Choose which Pokémon should have its types copied by Ninetales' $thisAbility?")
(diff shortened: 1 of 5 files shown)


src/tcgwars/logic/impl/gen4/MysteriousTreasures.groovy
```diff
@@ -1234,8 +1234,8 @@ public enum MysteriousTreasures implements LogicCardInfo {
               checkLastTurn()
               checkNoSPC()
               powerUsed()
-              def chosenPCS = opp.all.select("Choose which Pokémon should have its types copied by Ninetales' $thisAbility.name?")
-              bc "$thisAbility.name - $chosenPCS was chosen"
+              def chosenPCS = opp.all.select("Choose which Pokémon should have its types copied by Ninetales' $thisAbility?")
```
flagrama, line 1237: I'd probably change Ninetales to a `$self` as well.

flagrama (approved):
Just one small thing I'd change myself, but otherwise looks good.

src/tcgwars/logic/impl/gen4/MysteriousTreasures.groovy
starg09: Good call 👍

PR #964, opened 2020-10-02
Update Stantler's Push Away Ability

luongthomas:
This PR introduces a change to Stantler's Push Away move so that it will be more convenient to use by removing the need to show the same card list prompt twice

### DIFF
--- a/src/tcgwars/logic/impl/gen3/UnseenForces.groovy
+++ b/src/tcgwars/logic/impl/gen3/UnseenForces.groovy
@@ -1207,7 +1207,7 @@ public enum UnseenForces implements LogicCardInfo {
             damage 20
             afterDamage {
               if (opp.hand) {
-                def list = opp.hand.shuffledCopy().showToMe("Opponent's hand").filterByType(TRAINER)
+                def list = opp.hand.shuffledCopy().filterByType(TRAINER)
                 if(list){
                   list.select("Select a Trainer card to discard").discard()
                 }
--- a/src/tcgwars/logic/impl/gen3/UnseenForces.groovy
+++ b/src/tcgwars/logic/impl/gen3/UnseenForces.groovy
@@ -1206,7 +1206,7 @@ public enum UnseenForces implements LogicCardInfo {
           onAttack {
             damage 20
             afterDamage {
-              if (opp.hand && opp.hand.filterByType(TRAINER)) {
+              if (opp.hand) {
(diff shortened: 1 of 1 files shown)


src/tcgwars/logic/impl/gen3/UnseenForces.groovy
```diff
@@ -1207,7 +1207,7 @@ public enum UnseenForces implements LogicCardInfo {
             damage 20
             afterDamage {
               if (opp.hand) {
-                def list = opp.hand.shuffledCopy().showToMe("Opponent's hand").filterByType(TRAINER)
+                def list = opp.hand.shuffledCopy().filterByType(TRAINER)
```
starg09, line 1210: While not doing a show, then select sounds good, the player should still be able to see the whole hand.

This could be replaced with a select that makes use of a filter, to confirm you're selecting a trainer or skip. If player picked a card, it is discarded.

starg09 (changes requested):

flagrama:
Aren't the two lists completely different? The first is the Opponent's entire hand, the second is only the Trainer cards in it? I think it'd be better to show the whole hand, but make the whole hand show in the `select` and only allow trainers to be shown.

luongthomas:
> Aren't the two lists completely different? The first is the Opponent's entire hand, the second is only the Trainer cards in it? I think it'd be better to show the whole hand, but make the whole hand show in the `select` and only allow trainers to be shown.

Valid point, I'll make the modifications to do so

src/tcgwars/logic/impl/gen3/UnseenForces.groovy
```diff
@@ -1206,11 +1206,9 @@ public enum UnseenForces implements LogicCardInfo {
           onAttack {
             damage 20
             afterDamage {
-              if (opp.hand) {
-                def list = opp.hand.shuffledCopy().showToMe("Opponent's hand").filterByType(TRAINER)
-                if(list){
-                  list.select("Select a Trainer card to discard").discard()
-                }
+              if (opp.hand && opp.hand.filterByType(TRAINER)) {
```
flagrama, line 1209:
```suggestion
              if (opp.hand) {
```
You can't know if the opponent has a trainer in their hand unless you look at their hand.

src/tcgwars/logic/impl/gen3/UnseenForces.groovy
luongthomas: Good point, with my latest change I've made it so it shows the whole hand, only allowing selection of the trainer cards.

flagrama (approved):

starg09 (approved):

PR #965, opened 2020-10-03
Make two new statics and implement Galarian Sirfetch'd V promo

flagrama:
One for selecting any number of energies from your Pokemon in play
One for doing additional damage for each energy discarded using the first method

I made them separately because the Galarian Sirfetch'd V's ability can use the first one as well (with some slight additional modification).

Are there any other cards that would use these, or could use these with additional modifications?

### DIFF
--- a/src/tcgwars/logic/groovy/TcgStatics.groovy
+++ b/src/tcgwars/logic/groovy/TcgStatics.groovy
@@ -1602,4 +1602,59 @@ class TcgStatics {
     }
   }
 
+  /**
+   * Do additional damage per Energy card discarded from owner's PokemonCardSets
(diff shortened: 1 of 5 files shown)


flagrama:
[Create a static for "Discard any amount of Energy from your Pokémon." effects](https://trello.com/c/u53NEC3u/830-create-a-static-for-discard-any-amount-of-energy-from-your-pok%C3%A9mon-effects)

flagrama:
[SWSH Black Star Promos](https://trello.com/c/AU4zxAhk/749-swsh-black-star-promos)

luongthomas (approved):

PR #966, opened 2020-10-03
Implement SWSH promos 2020 oct 03

flagrama:
The [first commit](https://github.com/axpendix/tcgone-engine-contrib/commit/8528eb8bc8460c529c3f3f39632d01f7b3a3f971) contains all of the known promos for future reference. I pruned it down to only the released ones in the second commit so as not to break street dates, cause trouble, broken card images, etc. Galarian Farfetch'd V's ability relies on #965 so I didn't fully implement it, but it also doesn't release for a couple weeks anyway.

Is there actually any cases where the TODO on Eevee actually is relevant though? I'm pretty sure I've asked before, but don't remember the answer.


flagrama:
[SWSH Black Star Promos](https://trello.com/c/AU4zxAhk/749-swsh-black-star-promos)

luongthomas (approved):

flagrama:
Based on comments from Tegull, Some of the other promos have already been released, so I'll add them soon. Galarian Farfetch'd V is one of them, so I'll do that after the other PR.

starg09 (approved):

PR #967, opened 2020-10-04
Fix the scoops some more because the nightmare never ends

flagrama:
This should handle knockouts and changed implementations more cleanly.


luongthomas (approved):

flagrama:
@luongthomas could you give axpendix/tcgone#89 a review too please?

PR #974, opened 2020-10-18
Add method to only get the live expansions

flagrama:
This gives us the opportunity to add sets to the server without making them visible in the deck editor and play server card database.

Just add a new item to the list of dev-only IDs in the private method and the public one should return the list of all other expansions.

### DIFF
--- a/src/tcgwars/logic/card/Collection.java
+++ b/src/tcgwars/logic/card/Collection.java
@@ -1,11 +1,10 @@
 package tcgwars.logic.card;
 
 import org.apache.commons.lang.WordUtils;
+import tcgwars.logic.GameFormat;
+import tcgwars.logic.GameType;
 
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.List;
-import java.util.Locale;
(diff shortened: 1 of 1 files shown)


axpendix (commented):
-

axpendix (changes requested):
Hey, I think this could be better handled with an auto-logic like this: run through all active formats for each `GameType`, then get `Collection`s for each `GameFormat`, put them in a set, the remaining `Collections` are dev-only.

flagrama:
That certainly sounds like a better idea. I suppose it would also let us set up formats for Dev for testing if any rule changes occur which could be nice. Would the code still be here in Collection, or do you think there is a more proper place for it?

axpendix:
> That certainly sounds like a better idea. I suppose it would also let us set up formats for Dev for testing if any rule changes occur which could be nice. Would the code still be here in Collection, or do you think there is a more proper place for it?

👍 `Collection` should be a fine place for this

axpendix (approved):

PR #975, opened 2020-10-19
Fix selecting energy using the selectEnergyFromPokemon static

flagrama:


flagrama:
[Electrode](https://trello.com/c/ImpBGCrK/833-electrode)

starg09 (commented):
Just tested this locally and seems to work fine. Minor detail, would change the pokemon selection message from "Cancel to choose none." to "Cancel to stop.", since that's what it'd do.

axpendix (approved):

PR #980, opened 2020-10-28
Amazing volt tackle [G]

flagrama:
Shiftry's ability is a bit complicated. Skipping for the moment.

Also going to try breaking this set down more than Legendary Heartbeat, hoping to get it reviewed quicker.

### DIFF
--- a/src/tcgwars/logic/impl/gen8/AmazingVoltTackle.groovy
+++ b/src/tcgwars/logic/impl/gen8/AmazingVoltTackle.groovy
@@ -257,9 +257,9 @@ public enum AmazingVoltTackle implements LogicCardInfo {
         move "Absorb", {
           text "10 damage. Heal 10 damage from this Pokémon."
           energyCost G
-          attackRequirement {}
           onAttack {
             damage 10
+            heal 10, self
           }
         }
       };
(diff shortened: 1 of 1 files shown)


src/tcgwars/logic/impl/gen8/AmazingVoltTackle.groovy
```diff
@@ -237,16 +235,19 @@ public enum AmazingVoltTackle implements LogicCardInfo {
...
         }
         move "Fan Tornado", {
           text "110 damage. You may have your opponent switch their Active Pokémon with 1 of their Benched Pokémon."
           energyCost G, C
-          attackRequirement {}
           onAttack {
             damage 110
+            bg.deterministicCurrentThreadPlayerType=bg.currentTurn.opposite
+            switchYourActive()
+            bg.clearDeterministicCurrentThreadPlayerType()
```
axpendix, line 250: I'd refrain from using the deterministicCurrentThreadPlayerType setting, it should be meant for internal engine use only.

flagrama: Right. Whirlwind static is the proper thing to use here. I just couldn't remember that it existed when I was writing this. Will fix.

axpendix (approved):
kudos to the implementation, not the easiest card.  we need a consistent way to have "replacement effects" of attacks, abilities and trainers.

PR #982, opened 2020-10-30
Implement [W] Amazing Volt Tackle Pokemon and add static for "Armor" abilities

flagrama:
As with the Metronome static, I created one for Armor abilities since they are pretty common, and will use a separate future PR to implement it in all the appropriate places.

### DIFF
--- a/src/tcgwars/logic/impl/gen8/AmazingVoltTackle.groovy
+++ b/src/tcgwars/logic/impl/gen8/AmazingVoltTackle.groovy
@@ -502,13 +502,13 @@ public enum AmazingVoltTackle implements LogicCardInfo {
           onDeactivate {
             bg.em().run(new CheckAbilities())
           }
-          getterA IS_ABILITY_BLOCKED, {holder->
-            if (self.cards.any { it.name == "Memory Capsule" } && thisAbility.name != holder.effect.ability.name && holder.effect.ability instanceof BwAbility) {
-              holder.object = true
+          getterA GET_ABILITIES, BEFORE_LAST, {Holder holder->
+            if (self.cards.any { it.name == "Memory Capsule" } && holder.effect.target != self) {
+              holder.object.keySet().removeIf { it instanceof BwAbility }
             }
           }
-          getterA IS_GLOBAL_ABILITY_BLOCKED, {holder->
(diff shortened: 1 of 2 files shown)


src/tcgwars/logic/impl/gen8/AmazingVoltTackle.groovy
```diff
@@ -496,13 +496,34 @@ public enum AmazingVoltTackle implements LogicCardInfo {
...
+          getterA IS_ABILITY_BLOCKED, {holder->
+            if (self.cards.any { it.name == "Memory Capsule" } && thisAbility.name != holder.effect.ability.name && holder.effect.ability instanceof BwAbility) {
+              holder.object = true
+            }
+          }
+          getterA IS_GLOBAL_ABILITY_BLOCKED, {holder->
+            if (self.cards.any { it.name == "Memory Capsule" } && thisAbility.name != holder.effect.ability.name && holder.effect.ability instanceof BwAbility) {
+              holder.object = true
+            }
+          }
```
axpendix, line 514: I think "has no abilities" needs to be `GET_ABILITIES`

flagrama: Okay, I'll give that a try. I pretty much copied this from Garbodor BKP since it was pretty similar. Would this need changed for that as well?

axpendix: Not sure how recennt that is but I checked this onne from sun moonn and it seems fairly implemented 

```      case ALOLAN_MUK_58:
        return evolution (this, from:"Alolan Grimer", hp:HP120, type:PSYCHIC, retreatCost:4) {
          weakness PSYCHIC
          bwAbility "Power of Alchemy", {
            text "Each Basic Pokémon in play, in each player's hand, and in each player's discard pile has no Abilities."
            getterA (GET_ABILITIES, BEFORE_LAST) {h->
              if (h.effect.target.basic && h.effect.target != self) {
                h.object.keySet().removeIf{it instanceof BwAbility}
              }
            }
            getterA IS_GLOBAL_ABILITY_BLOCKED, {Holder h->
              if ((h.effect.target as Card).cardTypes.is(BASIC)) {
                h.object=true
              }
            }
            onActivate {
              new CheckAbilities().run(bg)
            }
            onDeactivate {
              new CheckAbilities().run(bg)
            }
          }
```

axpendix (approved):

PR #983, opened 2020-10-31
Implement [L] Amazing Volt Tackle Pokemon

flagrama:


src/tcgwars/logic/impl/gen8/AmazingVoltTackle.groovy
```diff
@@ -777,13 +794,34 @@ public enum AmazingVoltTackle implements LogicCardInfo {
...
         bwAbility "Thunderclap Awakening", {
           text "If this Pokémon has a Memory Capsule card attached to it, each player's Pokémon has no Abilities."
-          actionA {
+          onActivate {
+            bg.em().run(new CheckAbilities())
+          }
+          onDeactivate {
+            bg.em().run(new CheckAbilities())
+          }
+          getterA IS_ABILITY_BLOCKED, {holder->
```
axpendix, line 803: same here, please use `GET_ABILITIES`

PR #985, opened 2020-11-05
Pokemod rocket

TheAquaPiplup:
nowhere close to done


flagrama:
https://github.com/axpendix/tcgone-engine-contrib/pull/986

PR #994, opened 2020-11-10
Implement [D] Pokémon

flagrama:


src/tcgwars/logic/impl/gen8/AmazingVoltTackle.groovy
```diff
@@ -1642,17 +1641,39 @@ public enum AmazingVoltTackle implements LogicCardInfo {
...
+                    if (energy.energyCount() > 1) {
+                      def choices = 1..energy.energyCount()
+                      def choiceInfo = "How many Energy do you want this card to count as?"
+                      energyCount = choose(choices, choiceInfo)
+                    }
+                    finalCount += energyCount
+                    energy.moveTo my.hand
+                  }
+                }
+            }
```
axpendix, line 1674:
```suggestion
            discardSelfEnergy C, C
```

flagrama: This doesn't allow you to choose if you only want a multi energy card to count as a single energy.

> Q. Can I use Umbreon-GX's "Dark Call GX" attack to discard 2 Double Colorless Energy?
A. Yes, you can select 1 Energy of each Double Colorless Energy to discard both of them. This applies for discarding your own Energy for an attack and for retreating as well. So if you used Hoopa-EX's "Hyperspace Fury" attack or paid its Retreat Cost of 2, discarding 2 Double Colorless Energy would be legal options. (Jan 26, 2017 TPCi Rules Team; Jun 27, 2017 TPCi Rules Team)

While solving this in the energy selection would be ideal, and something I plan to do once I have some time to, I think it's better if the card can be used as close to how it can in the paper game until then.

flagrama: I'll work on fixing the energy counting in general ASAP so that these codeblocks can be replaced with just the discardSelfEnergy code, but for now I'm going to merge this as-is since it works correctly this way, and the set releases tomorrow so I'd like to get these things tested.

flagrama: I will fully admit it should discard instead of going to the hand though 😆

axpendix: Hi @flagrama I was confused with the implementation, it mentioned returning energy to hand :) 

I agree this feature should be built-in to `discardSelfEnergy` method call. I am glad you are on top of things, thanks!

PR #1000, opened 2020-11-12
VIV Release Day Promos

flagrama:
- Remaining Champions Path Pin Collection Promos
- VIV Prerelease Promos
- VIV Blister Pack Promos


axpendix:
1000!

PR #1005, opened 2020-11-17
Dialogue: Text consistency, punctuation at the ends

loonaticx:
I went ahead and replaced all instances of "pokemon" and "Pokemon" with the proper "Pokémon" name. Additionally fixing some grammatical issues prominent in some areas.


axpendix (approved):
Hi, thanks for this PR, looks good :)

flagrama:
@loonaticx the asserts already add periods to the end of messages, so adding one causes a double period at the end.

PR #1007, opened 2020-11-19
Refactor Holon-Energy Pokemon Global Ability

luongthomas:
This PR introduces the changes to move the GlobalAbility for holon energy-pokmon (Holon's Voltorb, Holon's Castform, etc) into something more reusable inside TcgStatics.

These changes also update the logic of how attachment works so that it will not interfere with other Ability-based energy attachments such as Meganium Ex's `Nurture and Heal` ability.

This PR is dependent on PR https://github.com/axpendix/tcgone/pull/109 which allows the PlayEnergy effect to have the energy attachment target defined by the caller.  

It is used in these changes to determine if the Holon's Energy-Pokemon can be attached or not.  (clarification: it should not be allowed if an energy has already been played for the turn)

### DIFF
--- a/src/tcgwars/logic/groovy/TcgStatics.groovy
+++ b/src/tcgwars/logic/groovy/TcgStatics.groovy
@@ -18,6 +18,7 @@ import static tcgwars.logic.effect.Source.*
 import static tcgwars.logic.effect.ability.Ability.ActivationReason.*
 import static tcgwars.logic.groovy.TcgBuilders.delayed
 import static tcgwars.logic.groovy.TcgBuilders.getter
+import static tcgwars.logic.groovy.TcgBuilders.specialEnergy
 import static tcgwars.logic.groovy.TcgStatics.*
 import tcgwars.logic.*
 import tcgwars.logic.card.*
@@ -44,7 +45,7 @@ class TcgStatics {
 //		}
   }
 
-  static Type C = COLORLESS, R = FIRE, F = FIGHTING, G = GRASS, W = WATER, P = PSYCHIC, L = LIGHTNING, M = METAL, D = DARKNESS;
+  static Type C = COLORLESS, R = FIRE, F = FIGHTING, G = GRASS, W = WATER, P = PSYCHIC, L = LIGHTNING, M = METAL, D = DARKNESS, Y = FAIRY;
 
   static Weakness weak (Type t, String f=Weakness.X2){
     new Weakness(t,f)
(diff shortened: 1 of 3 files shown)


flagrama:
Sorry, don't have time to read through it all at this exact moment, but why does PlayEnergy need changed instead of just using the `allowAttach` closure of a SpecialEnergy?

luongthomas:
> Sorry, don't have time to read through it all at this exact moment, but why does PlayEnergy need changed instead of just using the `allowAttach` closure of a SpecialEnergy?

From my understanding, the `allowAttach` closure only defines the limitations on what kind of Pokemon the energy can be attached to.  

The purpose of changing PlayEnergy is so that I can attach and energy if allowed but also run into a `CantPlayEnergy` if an energy has already been played for the turn.

flagrama:
Holon's Magnemite and Holon's Voltorb do not provide Rainbow, they provide C. They also do not require an already attached energy, nor return one to the hand.

src/tcgwars/logic/groovy/TcgStatics.groovy
```diff
@@ -982,6 +983,83 @@ class TcgStatics {
...
+                    def returningEnergy = pcs.cards.filterByType(ENERGY).select(count: 1)
+
+                    bg.em().run(new ChangeImplementation(energyCard, thisCard))
+                    def cannotPlayEnergy = bg.em().run(new PlayEnergy(energyCard, pcs))
+                    if (cannotPlayEnergy) {
+                      bg.em().run(new ChangeImplementation(thisCard, energyCard))
+                    } else {
+                      returningEnergy.moveTo(thisCard.player.pbg.hand)
+
+                      bg.em().storeObject("Holon_Pokemon_Energy", bg.turnCount)
```
flagrama, line 1025: Is this actually used for anything? I thought it was just a hacky way to make Holon's Pokémon to count as turn attachment. This shouldn't be needed if properly using PlayEnergy instead of skipping PlayEnergy and using AttachEnergy.

src/tcgwars/logic/groovy/TcgStatics.groovy
```diff
@@ -982,6 +983,83 @@ class TcgStatics {
     }
   }
 
+  static holon_pokemon_energy(Object delegate, Integer rainbowEnergyCount) {
+    delegate.globalAbility { Card thisCard->
+      delayed {
+          before PLAY_CARD, {
+            def validPokemon = thisCard.player.pbg.all.findAll {it.cards.energyCount()}
+            def holonAlreadyPlayed = bg.em().retrieveObject("Holon_Pokemon_Energy") == bg.turnCount
```
flagrama, line 991: These checks also shouldn't be needed if using PlayEnergy. `allowAttach` should work fine for checking valid Pokémon in my testing.

src/tcgwars/logic/groovy/TcgStatics.groovy
```diff
@@ -982,6 +983,83 @@ class TcgStatics {
...
+                    }
+                    energyCard.player = thisCard.player
+
+                    def pcs = validPokemon.select("Attach to?")
+
+                    // Select an energy before attachment so that thisCard doesn't show up as an option
+                    def returningEnergy = pcs.cards.filterByType(ENERGY).select(count: 1)
+
+                    bg.em().run(new ChangeImplementation(energyCard, thisCard))
+                    def cannotPlayEnergy = bg.em().run(new PlayEnergy(energyCard, pcs))
```
flagrama, line 1019: You should create a separate variable for referencing the PlayEnergy effect so that you can use it's properties after it has been run. I feel like this is why you were trying to add a pcs to PlayEnergy too, which I don't think is necessary.

```suggestion
                    def playEnergy = new PlayEnergy(energyCard)
                    bg.em().run(playEnergy)
                    def cannotPlayEnergy = !playEnergy.attached
```

src/tcgwars/logic/groovy/TcgStatics.groovy
```diff
@@ -982,6 +983,83 @@ class TcgStatics {
...
+            else if (!isFromAbility && isEnergyAlreadyPlayed && ef.card.name == thisCard.name) {
+              wcu "Cannot play any more energy this turn."
+              prevent()
+            }
+          }
+          after ATTACH_ENERGY, {
+            if (ef.card.name == thisCard.name) {
+              bg.em().storeObject("Holon_Pokemon_Energy", bg.turnCount)
+            }
+          }
```
flagrama, line 1058: These shouldn't be required at all when using PlayEnergy as well.

flagrama (changes requested):
I think this static is trying to do more than it needs to. I think it can be simplified quite a bit.

src/tcgwars/logic/groovy/TcgStatics.groovy
```diff
@@ -982,6 +983,83 @@ class TcgStatics {
...
+                      onRemoveFromPlay {
+                        bg.em().run(new ChangeImplementation(thisCard, energyCard))
+                      }
+                    }
+                    energyCard.player = thisCard.player
+
+                    def pcs = validPokemon.select("Attach to?")
+
+                    // Select an energy before attachment so that thisCard doesn't show up as an option
+                    def returningEnergy = pcs.cards.filterByType(ENERGY).select(count: 1)
```
flagrama, line 1016: As an aside, filterByType is good for checking if a card exists, but I don't think we should use it in selects. It can be helpful to see all of the cards attached to a PCS when using a select, and the filter of the select can be used to ensure only the correct cards can be selected.

```suggestion
                    def returningEnergy = playEnergy.attached.cards.select cardTypeFilter(ENERGY)
```

src/tcgwars/logic/groovy/TcgStatics.groovy
luongthomas: Good point, I've removed the code that uses this storedObject

src/tcgwars/logic/groovy/TcgStatics.groovy
luongthomas: Thanks for the suggestion, I've used `allowedAttach` and it appears to be working great

src/tcgwars/logic/groovy/TcgStatics.groovy
luongthomas: This has been entirely removed, you're correct that this is unneeded when using PlayEnergy

src/tcgwars/logic/impl/gen3/HolonPhantoms.groovy
```diff
@@ -276,7 +279,7 @@ public enum HolonPhantoms implements LogicCardInfo {
           actionA {
             assert bg.em().retrieveObject("Form_Change") != bg.turnCount : "You can’t use more than 1 Form Change Poké-Power each turn"
             checkLastTurn()
-            assert my.deck : "Deck is empty"
+            assert my.deck : "Deck is empty!"
```
flagrama, line 282: The assert wcu still adds a . at the end doesn't it? "Deck is empty!." will probably look awkward if so.

luongthomas: Yes that is correct, I've reverted this

flagrama (approved):
Looks a lot better now!

PR #1008, opened 2020-11-20
Add Shiny Star V Groovy Template

luongthomas:


flagrama (approved):
Could you add it to Collection with the other dev/testing sets too?

luongthomas:
> Could you add it to Collection with the other dev/testing sets too?

Added to Collections!

src/tcgwars/logic/card/Collection.java
```diff
@@ -177,6 +177,7 @@
   LEGENDARY_HEARTBEAT(1010, "JS3A", "gen8.LegendaryHeartbeat"),
   AMAZING_VOLT_TACKLE(1011, "JS4", "gen8.AmazingVoltTackle"),
   WIZARDS_BLACK_STAR_PROMOS_NG(111, "WBSP", "gen1.WizardsBlackStarPromosNG"),
+  SHINY_STAR_V(1012, "JS5", "gen8.ShinyStarV"),
```
flagrama, line 180:
```suggestion
  SHINY_STAR_V(1012, "JS4A", "gen8.ShinyStarV"),
```
Japanese cards have the set number in the bottom left so it's not a guessing game like it is with English set abbreviations 
image

PR #1010, opened 2020-11-21
Bug fixing

flagrama:
Shiftry's Ability Only While Active

Simplify Unown E (UF) Hidden Power attack
* Should no longer double prompt the player. Will show the full hand
whether there is a Basic in hand or not. Message changes based on
whether a Basic is available or not as well.

Rocket's Wobb (TRR) Simplify and fix Dark Aid
* Simplified some repetitive code. Fixed Dark Aid not checking for
Rocket's Secret Machine cards. Fixed Dark Aid not always allowing
you to shuffle the cards to deck.

Clefairy (CEC) fix scoop up timing

Shedinja (VIV) playable with ability turned off


luongthomas (approved):
Your Dark Aid refactor looks pretty clean!

flagrama:
Thanks!

PR #1012, opened 2020-11-23
Use Safeguard effect for EX-era implementations

flagrama:
The safeguard static is for BW-era EXs. The Safeguard effect already
exists for EX-era exs, so we can use it for the relevant cards. I'm not
sure if the existing pokeBody closure should be removed?


axpendix:
This is a very nice change, though I think the existing pokeBody definitions should better be replaced.

luongthomas (approved):

PR #1014, opened 2020-11-26
Discard energy after damage

flagrama:
Implements the new SelectEnergy effect to handle discarding/moving Energy after damage properly. Energy is selected before damage, damage is applied, then energy is discarded/moved when these statics are used.

Implemented primarily in the gen8 sets for now, and replacing some TODO's left from a former non-static, non-SelectEnergy implementation in other sets. If all seems well, another PR will be made to try to massively replace existing improperly handled Energy discards/moves.


axpendix (approved):
Looks great! Thanks!

PR #1019, opened 2020-11-29
Phione crash fix

ufodynasty:
Fixes Phione crashes. used sw2(opp.active, opp.bench.oppSelect(...),SRC_ABILITY) rather than sw2(opp.bench.oppSelect(...), null,SRC_ABILITY) as it should target your opponent's active Pokemon. @axpendix 
[phione.txt](https://github.com/axpendix/tcgone-engine-contrib/files/5611729/phione.txt)
https://forum.tcgone.net/t/br-phione-smp-sm220-lucario-melmetal-gx-smp-sm192-phio/11188

### DIFF
--- a/src/tcgwars/logic/impl/gen7/CosmicEclipse.groovy
+++ b/src/tcgwars/logic/impl/gen7/CosmicEclipse.groovy
@@ -1595,7 +1595,7 @@ public enum CosmicEclipse implements LogicCardInfo {
               assert self.benched
               assertOppBench()
               powerUsed()
-              if(sw2(opp.bench.oppSelect("Choose a new Active Pokémon"), SRC_ABILITY)){
+              if(sw2(opp.active, opp.bench.oppSelect("Choose a new Active Pokémon"), SRC_ABILITY){
                 self.cards.getExcludedList(self.topPokemonCard).discard()
                 self.cards.moveTo(my.deck)
                 removePCS(self)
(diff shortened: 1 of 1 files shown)


flagrama:
If the opponent chooses the Pokémon to switch, would whirlwind not work here? You should be able to modify it to return a boolean if it doesn't currently so that you can check it in an if clause. Just remember that `true` means that the effect was prevented and `false` means it was successful when using `bg.em().run()` or `bg.em().activate()`

src/tcgwars/logic/impl/gen7/CosmicEclipse.groovy
```diff
@@ -1595,7 +1595,7 @@ public enum CosmicEclipse implements LogicCardInfo {
               assert self.benched
               assertOppBench()
               powerUsed()
-              if(sw2(opp.bench.oppSelect("Choose a new Active Pokémon"), SRC_ABILITY)){
+              if(sw2(opp.active, opp.bench.oppSelect("Choose a new Active Pokémon"), SRC_ABILITY){
```
flagrama, line 1598:
```suggestion
              if(sw2(opp.active, opp.bench.oppSelect("Choose a new Active Pokémon"), SRC_ABILITY)){
```

flagrama (approved):

axpendix:
Hi @ufodynasty I am sorry for suggesting incorrect code in the previous PR.

PR #1020, opened 2020-11-29
whirlwind3 static

ufodynasty:
Im not happy with this but thought I would make it public in case there is a better way to do it.


flagrama:
For the record, this is missing a lot of the logic that `Whirlwind` has to check. It'd probably need to be a re-implementation of that method in it's entirety.

PR #1022, opened 2020-11-30
Bug Fixes

ufodynasty:


luongthomas (approved):

src/tcgwars/logic/impl/gen4/Unleashed.groovy
```diff
@@ -205,12 +205,13 @@ public enum Unleashed implements LogicCardInfo {
             energyCost P
             onAttack {
               int max = self.cards.energyCount(C)
+              def tar = opp.all.findAll{ it.evolution }
```
flagrama, line 208: How do you select the second Pokémon to devolve if you move this out of the loop?

src/tcgwars/logic/impl/gen4/Unleashed.groovy
```diff
@@ -205,12 +205,13 @@ public enum Unleashed implements LogicCardInfo {
...
               while (max-- > 0) {
-                def tar = opp.all.findAll{ it.evolution }
                 if(!tar) break
                 def pcs = tar.select("Choose which Pokémon to devolve", false)
                 if(!pcs) break
                 def top=pcs.topPokemonCard
+                tar.remove(pcs)
                 devolve(pcs, top, opp.hand)
               }
```
flagrama, line 216: We should probably use a `multiSelect` here and process it all at once. This loop seems unnecessary to me since it's just preventing you from devolving a Stage 2 to a Stage 1. (which is proper, but can be done differently)

flagrama (changes requested):

src/tcgwars/logic/impl/gen4/Unleashed.groovy
ufodynasty: The issue with using `multiSelect` here is that it doesn't have a minimum parameter.

flagrama: @axpendix Do you think it would be worthwhile to add an overload to `selectMultiPokemon` that takes a `min` Integer and sets the `required` property of the select to false once the min has been reached? Current one would call the new one with count as the min and the max. Count would probably be renamed to max for the new one as well.

There are several cards I can find real quick that could use this, including a couple recent ones:
(4) https://pkmncards.com/?s=%2Btext%3A%22Choose+a+number+of%22+%2Btext%3A%22Pokemon%22+-skarmory+-gengar&display=full&sort=date&order=
(5) https://pkmncards.com/?s=%2Btext%3A%22choose+up+to+2%22+%2Btext%3A%22benched+pokemon%22+-kyogre+-groudon+-empoleon&display=full&sort=date&order=
https://pkmncards.com/card/cofagrigus-lost-thunder-lot-100/
https://pkmncards.com/card/cramorant-vivid-voltage-viv-040/

axpendix: @flagrama sorry that I just saw this, I think that would be really useful. There are many cards that require that as you mentioned. 

There could be even a different variation that selects a pair of pokemon for like moving damage counter/energy around.

flagrama: I can't wrap my head around how to do the latter one right this moment, but I've made PRs for the former here: #1027 and https://github.com/axpendix/tcgone/pull/129

@ufodynasty after these are merged, would you mind changing the above mentioned cards to use it as well in this PR?

axpendix (approved):

PR #1023, opened 2020-12-01
Call of legends

ufodynasty:


src/tcgwars/logic/impl/gen4/CallOfLegends.groovy
```diff
@@ -786,598 +452,215 @@ public enum CallOfLegends implements LogicCardInfo {
...
-            onAttack {
-              damage 0
+      case BAYLEEF_40:
+        return copy(HeartgoldSoulsilver.BAYLEEF_35, this);
+      case CROCONAW_41:
+        return copy(HeartgoldSoulsilver.CROCONAW_38, this);
+      case DONPHAN_42:
+        return copy(HeartgoldSoulsilver.DONPHAN_40, this);
+      case FLAAFFY_43:
+        return copy(HeartgoldSoulsilver.FLAFFY_42, this);
```
flagrama, line 534:
```suggestion
        return copy(HeartgoldSoulsilver.FLAAFFY_42, this);
```

src/tcgwars/logic/impl/gen4/CallOfLegends.groovy
```diff
@@ -786,598 +452,215 @@ public enum CallOfLegends implements LogicCardInfo {
...
+        return basic (this, hp:HP050, type:GRASS, retreatCost:1) {
+          weakness R
+          resistance W, MINUS20
+          move "Nap", {
+            text "Remove 1 damage counter from Chikorita."
             energyCost C
             attackRequirement {}
             onAttack {
-              damage 0
+              heal 10
```
flagrama, line 620:
```suggestion
              heal 10, self
```
Target PCS is required.

src/tcgwars/logic/impl/gen4/CallOfLegends.groovy
```diff
@@ -1593,110 +788,80 @@ public enum CallOfLegends implements LogicCardInfo {
...
+        return stadium (this) {
           text "This card stays in play when you play it. Discard this card if another Stadium card comes into play. If another card with the same name is in play, you can’t play this card.\nOnce during each player’s turn, if this player’s opponent has 6 or more Pokémon in the Lost Zone, the player may choose to win the game."
+          def lastTurn=0
+          def actions=[]
           onPlay {
+            actions=action("Stadium: Lost World") {
+              assert lastTurn != bg().turnCount : "You've already used Lost World this turn."
+              bc "Used Lost World."
+              lastTurn = bg().turnCount
+              bg.getGame().endGame(self.owner, WinCondition.OTHER);
```
flagrama, line 834: A stadium card does not have a `self`. I think you may want to do something with the `getStadiumInfoStruct` here?
```suggestion
              bg.getGame().endGame(bg.getStadiumInfoStruct().stadiumCard.player, WinCondition.OTHER);
```

src/tcgwars/logic/impl/gen4/CallOfLegends.groovy
```diff
@@ -1,6 +1,8 @@
 package tcgwars.logic.impl.gen4
 
 import tcgwars.logic.impl.gen3.RubySapphire
+import tcgwars.logic.impl.gen3.TeamRocketReturns;
+import tcgwars.logic.impl.gen3.DeltaSpecies;
```
flagrama, line 5:
```suggestion
import tcgwars.logic.impl.gen3.DeltaSpecies;
import tcgwars.logic.WinCondition
```

src/tcgwars/logic/impl/gen4/CallOfLegends.groovy
```diff
@@ -1,6 +1,8 @@
...
 
 import static tcgwars.logic.card.HP.*;
 import static tcgwars.logic.card.Type.*;
```
flagrama, line 8:
```suggestion
import static tcgwars.logic.card.HP.*;
import static tcgwars.logic.card.Type.*;
import static tcgwars.logic.card.CardType.*
import static tcgwars.logic.effect.EffectType.*
import static tcgwars.logic.effect.Source.*
import static tcgwars.logic.effect.special.SpecialConditionType.*
```

src/tcgwars/logic/impl/gen4/CallOfLegends.groovy
```diff
@@ -786,598 +452,215 @@ public enum CallOfLegends implements LogicCardInfo {
...
+            text "30 damage. Flip a coin. If heads, the Defending Pokémon is now Paralyzed. If tails, put 1 Energy card attached to the Defending Pokémon in the Lost Zone."
+            energyCost G, C, C
             attackRequirement {}
             onAttack {
-              damage 0
+              damage 30
+              flip 1, {
+                apply PARALYZED
+              }, {
+                defending.cards.filterByType(ENERGY).select("Choose an energy to put into the Lost Zone").moveTo(opp.lostZone)
```
flagrama, line 491: I prefer using the filter for CardList `select`s on a Pokémon's cards. That way you can see all of the cards currently attached in case it could inform your decision.

src/tcgwars/logic/impl/gen4/CallOfLegends.groovy
```diff
@@ -786,598 +452,215 @@ public enum CallOfLegends implements LogicCardInfo {
...
+        return copy(Triumphant.MAGBY_41, this);
+      case MIME_JR__47:
+        return basic (this, hp:HP030, type:PSYCHIC, retreatCost:0) {
+          pokeBody "Sweet Sleeping Face", {
+            text "As long as Mime Jr. is Asleep, prevent all damage done to Mime Jr. by attacks."
+            delayedA {
+              before APPLY_ATTACK_DAMAGES, {
+                bg.dm().each {
+                  if(self.isSPC(ASLEEP) && it.to == self && it.dmg.value && it.notNoEffect) {
+                    bc "Sweet Sleeping Face: Damage was prevented"
```
flagrama, line 549:
```suggestion
                    bc "$thisAbility prevents damage"
```
This can help make it easier if this is referenced for a copy/paste or as a candidate for a new static method. Also, most of the strings for these types of effects are written in this format, so best to try to keep it consistent.

src/tcgwars/logic/impl/gen4/CallOfLegends.groovy
```diff
@@ -786,598 +452,215 @@ public enum CallOfLegends implements LogicCardInfo {
...
-          move "High Jump Kick", {
-            text "60 damage. "
-            energyCost F, F, C
+          move "Sleepy Lost", {
+            text "Put the top card of your opponent’s deck in the Lost Zone. Mime Jr. is now Asleep."
+            energyCost ()
             attackRequirement {}
             onAttack {
-              damage 0
+              opp.deck.first().moveTo(opp.lostZone)
```
flagrama, line 561: > Mime Jr. is now Asleep

src/tcgwars/logic/impl/gen4/CallOfLegends.groovy
```diff
@@ -786,598 +452,215 @@ public enum CallOfLegends implements LogicCardInfo {
...
-            text "20 damage. "
-            energyCost C, C
+          move "Poison Effect", {
+            text "20+ damage. If Seviper is Poisoned, this attack does 20 damage plus 60 more damage and remove the Special Condition Poisoned from Seviper."
+            energyCost P, C
             attackRequirement {}
             onAttack {
-              damage 0
+              damage 20
+              if(self.isSPC(POISONED)){
```
flagrama, line 601:
```suggestion
              if (self.isSPC(POISONED)) {
```
Whitespace readability

src/tcgwars/logic/impl/gen4/CallOfLegends.groovy
```diff
@@ -786,598 +452,215 @@ public enum CallOfLegends implements LogicCardInfo {
...
+                damage 60
+                afterDamage {clearSpecialCondition(self, ATTACK, [POISONED])}
```
flagrama, line 603:
```suggestion
                afterDamage { clearSpecialCondition(self, ATTACK, [POISONED]) }
```
Whitespace readability

src/tcgwars/logic/impl/gen4/CallOfLegends.groovy
```diff
@@ -1387,62 +670,28 @@ public enum CallOfLegends implements LogicCardInfo {
...
         return basic (this, hp:HP060, type:FIGHTING, retreatCost:2) {
           weakness W
           resistance L, MINUS20
           pokeBody "Ultra-Thick Skin", {
             text "As long as Phanpy has Energy attached to it, any damage done to Phanpy by attacks is reduced by 10 ."
-            delayedA {
+            delayed {
+              before APPLY_ATTACK_DAMAGES, {
+                bg.dm().each {if(it.to==self && it.from.owner==self.owner.opposite && it.notZero && it.notNoEffect){
+                  bc "Ultra-Thick Skin"
```
flagrama, line 691:
```suggestion
                  bc "$thisAbility -10"
```
Generalized string and make same format as other similar messages.

src/tcgwars/logic/impl/gen4/CallOfLegends.groovy
```diff
@@ -1584,7 +768,18 @@ public enum CallOfLegends implements LogicCardInfo {
...
+              flip {
+                delayed {
+                  before PLAY_TRAINER, {
+                    if (bg.currentTurn == self.owner.opposite) {
+                      wcu "Fake Tears prevents you from playing Trainer cards."
+                      prevent()
+                    }
+                  }
+                  unregisterAfter 2
+                }
```
flagrama, line 780: "Trainer Cards" in these formats actually means "ITEM" cards I'm pretty sure. Supporter cards and Stadium cards were not considered Trainers at the time and did not have TRAINER printed on their cards.

src/tcgwars/logic/impl/gen4/CallOfLegends.groovy
```diff
@@ -1593,110 +788,80 @@ public enum CallOfLegends implements LogicCardInfo {
...
-          playRequirement{
-          }
-        };
+        return copy(Unleashed.INTERVIEWER_S_QUESTIONS_77, this)
       case LOST_REMOVER_80:
         return basicTrainer (this) {
           text "Put 1 Special Energy card attached to 1 of your opponent’s Pokémon in the Lost Zone."
           onPlay {
+            def pcs = opp.all.findAll{it.cards.filterByType(SPECIAL_ENERGY)}.select("Remove special energy from which Pokémon?")
+            pcs.cards.filterByTYpe(SPECIAL_ENERGY).select("Put which special energy in the Lost Zone?").first().moveTo(opp.lostZone)
```
flagrama, line 818: Please use the filter parameter instead of `filterByType` here as well.

src/tcgwars/logic/impl/gen4/CallOfLegends.groovy
```diff
@@ -200,55 +205,17 @@ public enum CallOfLegends implements LogicCardInfo {
             energyCost M, M, M, M
             attackRequirement {}
             onAttack {
-              damage 0
+              damage 70
+              my.hand.moveTo(hidden:true, my.deck)
+              shuffleDeck()
```
flagrama, line 210: If it is changing game state (adding/removing from hand/deck/pokemoncardset's cards/etc), it should probably be done `afterDamage`. This applies to many effects.

src/tcgwars/logic/impl/gen4/CallOfLegends.groovy
```diff
@@ -786,598 +452,215 @@ public enum CallOfLegends implements LogicCardInfo {
...
-            energyCost C, C
+          move "Plow Over", {
+            text "30 damage. Flip a coin. If heads, the Defending Pokémon is now Paralyzed. If tails, put 1 Energy card attached to the Defending Pokémon in the Lost Zone."
+            energyCost G, C, C
             attackRequirement {}
             onAttack {
-              damage 0
+              damage 30
+              flip 1, {
+                apply PARALYZED
```
flagrama, line 489:
```suggestion
                applyAfterDamage PARALYZED
```
Special Conditions should be applied after damage to help cause certain other checks related to some Pokémon Powers to be handled better.

src/tcgwars/logic/impl/gen4/CallOfLegends.groovy
```diff
@@ -1593,110 +788,80 @@ public enum CallOfLegends implements LogicCardInfo {
...
+        return copy(HeartgoldSoulsilver.PROFESSOR_OAK_S_NEW_THEORY_101, this);
       case RESEARCH_RECORD_84:
         return basicTrainer (this) {
           text "Look at the top 4 cards of your deck and put as many of them as you like back on top of your deck in any order. Then, put the remaining cards on the bottom of your deck in any order."
           onPlay {
+            def tar = my.deck.subList(0,5)
+            def sel = tar.select(min:0, max:tar.size(), "Choose any number of cards to put on top of your deck(unselected cards will be put on the bottom of your deck)")
+            if(sel){
+              tar.remove(sel)
+              deck.setSublist(0,rearrange(sel,"Arrange the top of your deck"))
```
flagrama, line 853:
```suggestion
              def rearranged = rearrange(sel, "Arrange the top of your deck")
              deck.setSublist(0, rearranged)
```
It's usually easier to read if you store the result of a method call in a variable and then use it in another method call rather than try to include it all in the same line.

flagrama (changes requested):
Need to add some missing imports. I've included suggestions for the ones that seem to be required.

Many of these comments should be applied in many places in this file, but it's a bit hard to read through already with the number of comments, I don't want to make it completely unreadable.

Looks pretty good, and only a couple of logic errors I could find. Should be able to merge it after just a little bit more work.

src/tcgwars/logic/impl/gen4/CallOfLegends.groovy
ufodynasty: In my experience, filtering before selection increases the clarity of the effect as it only shows your valid choices. I prefer to use the filter parameter of select for effects that have you choose one of n cards that meets a criteria such as team up jirachi.

src/tcgwars/logic/impl/gen4/CallOfLegends.groovy
flagrama: Here, and line 483. Getting into the habit of using the CardList.select()'s `filter` parameter should keep us from causing double prompts like Jirachi from Team Up's.

`moveTo` is only applicable to a CardList, so you can't use `first()` on it before that.

```suggestion
            pcs.cards.select("Put which special energy in the Lost Zone?", cardTypeFilter(SPECIAL_ENERGY)).moveTo(opp.lostZone)
```

src/tcgwars/logic/impl/gen4/CallOfLegends.groovy
```diff
@@ -1593,110 +783,82 @@ public enum CallOfLegends implements LogicCardInfo {
...
       case LOST_WORLD_81:
-        return basicTrainer (this) {
+        return stadium (this) {
           text "This card stays in play when you play it. Discard this card if another Stadium card comes into play. If another card with the same name is in play, you can’t play this card.\nOnce during each player’s turn, if this player’s opponent has 6 or more Pokémon in the Lost Zone, the player may choose to win the game."
+          def lastTurn=0
+          def actions=[]
           onPlay {
+            actions=action("Stadium: Lost World") {
+              assert lastTurn != bg().turnCount : "You've already used Lost World this turn."
+              assert opp.lostZone.filterByType(POKEMON).size<6 : "Your opponent has fewer than 6 Pokémon in the Lost Zone"
```
flagrama, line 826:
```suggestion
              assert opp.lostZone.filterByType(POKEMON).size() < 6 : "Your opponent has fewer than 6 Pokémon in the Lost Zone"
```

flagrama (changes requested):
Just a few more small things.

src/tcgwars/logic/impl/gen4/CallOfLegends.groovy
```diff
@@ -1151,427 +579,179 @@ public enum CallOfLegends implements LogicCardInfo {
...
+        return copy(Undaunted.PINECO_63, this);
       case RELICANTH_69:
         return basic (this, hp:HP080, type:WATER, retreatCost:2) {
           weakness G
           move "Prehistoric Wisdom", {
             text "Choose a card from your hand and put it in the Lost Zone. Then, draw 3 cards."
             energyCost C
-            attackRequirement {}
+            attackRequirement {
+              assert my.deck|my.hand : "You have no cards in your hand or deck!"
```
flagrama, line 711: Missed this as well.

```suggestion
              assert my.deck || my.hand : "You have no cards in your hand or deck!"
```
I don't think a bitwise or would work here, best to use the logical or.

src/tcgwars/logic/impl/gen4/CallOfLegends.groovy
```diff
@@ -431,75 +320,44 @@ public enum CallOfLegends implements LogicCardInfo {
           move "Linear Attack", {
             text "Choose 1 of your opponent’s Pokémon. This attack does 30 damage to that Pokémon."
             energyCost W, C, C
-            attackRequirement {}
             onAttack {
-              damage 0
+              damage 30, opp.bench.select()
```
flagrama, line 324:
```suggestion
              damage 30, opp.all.select()
```

src/tcgwars/logic/impl/gen4/CallOfLegends.groovy
```diff
@@ -431,75 +320,44 @@ public enum CallOfLegends implements LogicCardInfo {
...
           weakness F
           resistance M, MINUS20
           pokePower "Self-Generation", {
-            text "Once during your turn, when you put Pachirisu from your hand onto your Bench, you may attach up to 2 Energy cards from your hand to Pachirisu."
-            actionA {
+            text "Once during your turn, when you put Pachirisu from your hand onto your Bench, you may attach up to 2 [L] Energy cards from your hand to Pachirisu."
+            onActivate {reason ->
+              if(reason == PLAY_FROM_HAND && my.hand.filterByEnergyType(L) && confirm("Use Self-Generation?")) {
+                powerUsed()
+                attachEnergyFrom(type: L, my.hand, self)
```
flagrama, line 349: Need to set max.

src/tcgwars/logic/impl/gen4/CallOfLegends.groovy
```diff
@@ -522,279 +383,79 @@ public enum CallOfLegends implements LogicCardInfo {
...
-        };
+        return copy(HeartgoldSoulsilver.SLOWKING_12, this);
       case SNORLAX_33:
         return basic (this, hp:HP100, type:COLORLESS, retreatCost:4) {
           weakness F
           move "Layabout", {
             text "Remove all damage counters from Snorlax. Snorlax can’t use Layabout during your next turn."
             energyCost C, C, C
-            attackRequirement {}
+            attackRequirement {self.numberOfDamageCounters}
```
flagrama, line 447:
```suggestion
            attackRequirement {assert self.numberOfDamageCounters : "$self not damaged}
```
Need to use the assert statement. Nice to use an error string to the player so they know why it failed.

src/tcgwars/logic/impl/gen4/CallOfLegends.groovy
```diff
@@ -804,343 +465,110 @@ public enum CallOfLegends implements LogicCardInfo {
           weakness R
           resistance W, MINUS20
           move "Grind", {
```
flagrama, line 467: This is one of the ones that it would be nice to have an attackRequirement on as well.

src/tcgwars/logic/impl/gen4/CallOfLegends.groovy
```diff
@@ -804,343 +465,110 @@ public enum CallOfLegends implements LogicCardInfo {
...
-          move "Flamethrower", {
-            text "60 damage. Discard an Energy attached to Quilava."
-            energyCost R, C, C
-            attackRequirement {}
+          move "Sleepy Lost", {
+            text "Put the top card of your opponent’s deck in the Lost Zone. Mime Jr. is now Asleep."
+            energyCost ()
             onAttack {
-              damage 0
+              opp.deck.first().moveTo(opp.lostZone)
```
flagrama, line 553:
```suggestion
              opp.deck.moveTo(opp.lostZone)
```

src/tcgwars/logic/impl/gen4/CallOfLegends.groovy
```diff
@@ -1151,427 +579,179 @@ public enum CallOfLegends implements LogicCardInfo {
...
             text "Choose a card from your hand and put it in the Lost Zone. Then, draw 3 cards."
             energyCost C
-            attackRequirement {}
+            attackRequirement {
+              assert my.deck||my.hand : "You have no cards in your hand or deck!"
+            }
             onAttack {
-              damage 0
+              if(my.hand) {
+                my.hand.select("Choose a card to put in the Lost Zone").first().moveTo(my.lostZone)
```
flagrama, line 715:
```suggestion
                my.hand.select("Choose a card to put in the Lost Zone").moveTo(my.lostZone)
```

src/tcgwars/logic/impl/gen4/CallOfLegends.groovy
```diff
@@ -1151,427 +579,179 @@ public enum CallOfLegends implements LogicCardInfo {
...
       case PHANPY_66:
         return basic (this, hp:HP060, type:FIGHTING, retreatCost:2) {
           weakness W
           resistance L, MINUS20
           pokeBody "Ultra-Thick Skin", {
             text "As long as Phanpy has Energy attached to it, any damage done to Phanpy by attacks is reduced by 10 ."
-            delayedA {
+            delayed {
+              before APPLY_ATTACK_DAMAGES, {
+                bg.dm().each {if(it.to==self && it.from.owner==self.owner.opposite && it.notZero && it.notNoEffect) {
```
flagrama, line 681:
```suggestion
                bg.dm().each {if(it.to==self && it.notZero && it.notNoEffect) {
```
If the text is accurate, does not need to be damage done by opponent.

src/tcgwars/logic/impl/gen4/CallOfLegends.groovy
```diff
@@ -1151,427 +579,179 @@ public enum CallOfLegends implements LogicCardInfo {
...
           move "Prehistoric Wisdom", {
             text "Choose a card from your hand and put it in the Lost Zone. Then, draw 3 cards."
             energyCost C
-            attackRequirement {}
+            attackRequirement {
+              assert my.deck||my.hand : "You have no cards in your hand or deck!"
+            }
             onAttack {
-              damage 0
+              if(my.hand) {
```
flagrama, line 714: Hand check shouldn't be required since attackRequirement handles this.

src/tcgwars/logic/impl/gen4/CallOfLegends.groovy
```diff
@@ -1593,110 +783,82 @@ public enum CallOfLegends implements LogicCardInfo {
...
       case LOST_WORLD_81:
-        return basicTrainer (this) {
+        return stadium (this) {
           text "This card stays in play when you play it. Discard this card if another Stadium card comes into play. If another card with the same name is in play, you can’t play this card.\nOnce during each player’s turn, if this player’s opponent has 6 or more Pokémon in the Lost Zone, the player may choose to win the game."
+          def lastTurn=0
+          def actions=[]
           onPlay {
+            actions=action("Stadium: Lost World") {
+              assert lastTurn != bg().turnCount : "You've already used Lost World this turn."
+              assert opp.lostZone.filterByType(POKEMON).size() < 6 : "Your opponent has fewer than 6 Pokémon in the Lost Zone"
```
flagrama, line 826:
```suggestion
              assert opp.lostZone.filterByType(POKEMON).size() >= 6 : "Your opponent has fewer than 6 Pokémon in the Lost Zone"
```
This assertion is backwards.

src/tcgwars/logic/impl/gen4/CallOfLegends.groovy
```diff
@@ -1593,110 +783,82 @@ public enum CallOfLegends implements LogicCardInfo {
...
       case RESEARCH_RECORD_84:
         return basicTrainer (this) {
           text "Look at the top 4 cards of your deck and put as many of them as you like back on top of your deck in any order. Then, put the remaining cards on the bottom of your deck in any order."
           onPlay {
+            def tar = my.deck.subList(0,5)
+            def sel = tar.select(min:0, max:tar.size(), "Choose any number of cards to put on top of your deck(unselected cards will be put on the bottom of your deck)")
+            if(sel) {
+              tar.remove(sel)
+              def rearranged = rearrange(sel,"Arrange the top of your deck")
+              deck.setSublist(0,rearranged)
```
flagrama, line 849:
```suggestion
              deck.setSubList(0,rearranged)
```

flagrama (changes requested):
This should be the last review before this can be merged. A couple more logical issues and typos found.

src/tcgwars/logic/impl/gen4/CallOfLegends.groovy
flagrama: I can see that, however other cards attached to the Pokémon may change which card you choose. The text string should provide clarify on what should be selected, and the other cards can help inform the decision IMO.

src/tcgwars/logic/impl/gen4/CallOfLegends.groovy
ufodynasty: I think the hand check is needed as a a legal play would to attack with no cards in hand and draw 3 cards without having put any into the lost zone.

flagrama: My bad. I forgot select doesn't check if the cardlist is empty before building the UI. I'll say this is good for now.

However, @axpendix is there any reason in the future we can't add a check if the list is empty to the cardlist select and just return an empty list in that case? This can cause issues where a player is stuck without the ability to do anything because there are no valid checks if it isn't properly guarded against, which also causes dual checks like these (attackRequirement checking your hand and onAttack checking your hand)

flagrama (approved):
LGTM!

src/tcgwars/logic/impl/gen4/CallOfLegends.groovy
axpendix: @flagrama that is a good idea. I was thinking to do it on the client level to affect all selections. A new business logic wrapping client class should handle auto selections, default values and so on. 
The second option is to do it on cardlist level as an simple 1 step fix. we already have the other use cases guarded by implementation logic anyway.

PR #1025, opened 2020-12-02
Implement Shiny Star V

flagrama:


src/tcgwars/logic/impl/gen8/LegendaryHeartbeat.groovy
```diff
@@ -1056,8 +1056,10 @@ public enum LegendaryHeartbeat implements LogicCardInfo {
           text " Both Active Pokémon are Knocked Out."
           energyCost M, M, C, C
           onAttack {
-            new Knockout(defending).run(bg)
-            new Knockout(self).run(bg)
+            afterDamage {
+              new Knockout(defending).run(bg)
+              new Knockout(self).run(bg)
+            }
```
axpendix, line 1062: I am curious what does this change?

src/tcgwars/logic/impl/gen8/ShinyStarV.groovy
```diff
@@ -107,13 +107,15 @@ public enum ShinyStarV implements LogicCardInfo {
         weakness R
         bwAbility "Familiar Land", {
           text "If you have any Stadium card in play, this Pokémon has no Retreat Cost."
-          actionA {
+          getterA GET_RETREAT_COST, BEFORE_LAST, self, { h->
+            if (bg.getStadiumInfoStruct().stadiumCard.player == self.owner) {
```
axpendix, line 111:
```suggestion
            if (bg.stadiumInfoStruct?.stadiumCard?.player == self.owner) {
```
otherwise it could produce NPE

luongthomas (approved):

src/tcgwars/logic/impl/gen8/LegendaryHeartbeat.groovy
flagrama: This was primarily an experiment that turned out to not have any effect. Committing it was an accident, but it shouldn't affect anything IIRC.

PR #1027, opened 2020-12-03
Use the new multiSelectPokemon Overload

flagrama:
Uses the new overload in places where it is appropriate to use in existing code.

Also adds text to ones missing it. Was done by just using the text property of the attack that is using it, but should provide better context.


src/tcgwars/logic/impl/gen3/FireRedLeafGreen.groovy
```diff
@@ -1249,7 +1249,7 @@ public enum FireRedLeafGreen implements LogicCardInfo {
             energyCost F, C
             onAttack {
               if (opp.bench) {
-                multiSelect(opp.bench, 2).each{
+                multiSelect(opp.bench, 1, 2, text).each{
```
ufodynasty, line 1252:
```suggestion
                multiSelect(opp.bench, 0, 2, text).each{
```
I think this would be better done with an attack requirement. Although we should try to prevent attacks doing nothing, once the attack has been initiated, I think the player should have the option to select 0 pokemon.

flagrama: I'm pretty sure the general ruling for "up to" is a minimum of 1. I do agree that an attack requirement for the bench existing should probably be done though.

axpendix (approved):

PR #1028, opened 2020-12-05
Bugfixes2

ufodynasty:


src/tcgwars/logic/groovy/TcgStatics.groovy
```diff
@@ -199,6 +199,12 @@ class TcgStatics {
     bg.em().activateEffect(ef)
     return ef.selectedCards ?: []
   }
+  static CardList selectOpponentEnergy(PokemonCardSet pcs, Type...types=C) {
+    def ef = new SelectEnergy(pcs.cards, types)
+    ef.playerType = pcs.owner.opposite
```
flagrama, line 204: This should not be the default for `discardDefendingEnergyAfterDamage`. Normally the attacking player chooses the energy. If the defender is supposed to choose it this should be a completely separate method.

ufodynasty: With the current implementation, the defending player is choosing the energy card. Thats why i made the bugfix.

flagrama: Oh, my bad, I guess I read it backwards. I see how that is an issue now. Is there a time where the player who's turn it is is not supposed to do the selection? Might be more appropriate to change `selectEnergy` to use `bg.currentTurn` instead of `pcs.owner`.

ufodynasty: `discardSelfEnergyAfterDamage` uses the same method `selectEnergy`. I could imagine adding a param to `selectEnergy` but im not sure where all it is used.

flagrama: That method should have the current turn player do the discard though no? So that change should work.

ufodynasty: Sry I didn't misunderstood your previous comment. The `selectEnergy` method gives the selection to the owner of the pcs that it targets, rather than the player who's turn it is.

src/tcgwars/logic/groovy/TcgStatics.groovy
```diff
@@ -193,15 +193,9 @@ class TcgStatics {
    * @param types {@link Type}s of energy to be selected. Default: C
    * @return {@link CardList} of the selected cards, can be empty CardList
    */
-  static CardList selectEnergy(PokemonCardSet pcs, Type...types=C) {
+  static CardList selectEnergy(PokemonCardSet pcs, Type...types=C, PlayerType playerType = pcs.owner) {
```
flagrama, line 196: This won't work. a `...` parameter needs to be at the end.

src/tcgwars/logic/groovy/TcgStatics.groovy
flagrama: Yeah, I think you should just change that part from `pcs.owner` to `bg.currentTurn`. No reason to make a new method, or to have a param for it.

src/tcgwars/logic/groovy/TcgStatics.groovy
```diff
@@ -232,7 +232,7 @@ class TcgStatics {
    */
   static discardDefendingEnergyAfterDamage(Type...types=C) {
     def pcs = Target.OPP_ACTIVE.getSingleTarget(bg)
-    def cards = selectEnergy(pcs, types)
+    def cards = selectEnergy(pcs, types, pcs.owner.opposite)
```
flagrama, line 235: Don't forget to put this back as well.

ufodynasty: Thank you :(

ufodynasty:
https://forum.tcgone.net/t/br-shaymin-ex-ros-106-bursting-balloon-bkp-97-i-sky-retu/11264
Q. Say my opponent used Turtonator GX's "Shell Trap" attack last turn, then on my turn I attack it with Shaymin EX's "Sky Return" attack. What happens to the 8 damage counters effect from "Shell Trap"?
A. Sky Return puts Shaymin EX into the hand as part of its attack before the damage counters are placed for Shell Trap's effect. (Jun 8, 2017 TPCi Rules Team)

src/tcgwars/logic/impl/gen8/RebelClash.groovy
```diff
@@ -3771,7 +3771,7 @@ public enum RebelClash implements LogicCardInfo {
               }
             }
             after APPLY_ATTACK_DAMAGES, {
-              if(attackDidDamage && self.cards.contains(thisCard)) { // this energy card is still attached
+              if(attackDidDamage && self.cards.contains(thisCard) && ef.attacker) { // this energy card is still attached and the attacking Pokemon is still in play
```
axpendix, line 3774: this `ef.attacker` condition will always be true, thus have no effect. are you trying to check whether the pokemon in question is still in the game? if yes, then it's better to use `ef.attacker.inPlay` instead

src/tcgwars/logic/groovy/TcgStatics.groovy
```diff
@@ -195,7 +195,7 @@ class TcgStatics {
    */
   static CardList selectEnergy(PokemonCardSet pcs, Type...types=C) {
     def ef = new SelectEnergy(pcs.cards, types)
-    ef.playerType = pcs.owner
+    ef.playerType = bg.currentTurn
```
axpendix, line 198: may I ask why this is needed? I would consider using `bg.currentThreadPlayerType` instead.

src/tcgwars/logic/impl/gen8/RebelClash.groovy
ufodynasty: Thanks for the suggestion, The bug that I was trying to fix was a null pointer exception on attacks that shuffle away the attacking pokemon. I would guess that ef.attacker.inPlay might throw the same exception if ef.attacker is null?

src/tcgwars/logic/groovy/TcgStatics.groovy
ufodynasty: The new method discardDefendingEnergyAfterDamage uses select energy, but builds the ui for the defending player rather than the attacking one. I discussed a few ways to do this with @flagrama, and we thought this was the best.

axpendix:
Hi @ufodynasty do you mind sharing your plans regarding this one? TY

ufodynasty:
@axpendix Sry I didn't clean this up sooner. @link--11 made a fix that did most of what this wanted to accomplish. An edit to his static as well as the other optimizations are in a new pr #1189

PR #1029, opened 2020-12-06
Triumphant

ufodynasty:
Still need to figure out how to write absol. I don't know how to access the played pokemon within an after PLAY_BASIC_POKEMON


axpendix (approved):
Thank you!

PR #1032, opened 2020-12-13
Call Of Legends Fixes

ufodynasty:
Will commit to this as more bugs are found.


src/tcgwars/logic/impl/gen4/CallOfLegends.groovy
```diff
@@ -483,9 +483,11 @@ public enum CallOfLegends implements LogicCardInfo {
               damage 30
               afterDamage{
                 flip 1, {
-                  apply PARALYZED
+                  applyAfterDamage PARALYZED
```
flagrama, line 486: This is already in an afterDamage closure. Is it not working properly?

axpendix (approved):
Thanks!

PR #1034, opened 2020-12-15
Removed the gen 3 supporter workaround

ufodynasty:
https://forum.tcgone.net/t/br-minun-dx-41-sniff-out-should-not-have-the-option-to-recov/11290


flagrama:
I think it may be better to do this properly than use a hack in a bunch of different places.

ufodynasty:
@flagrama changed the engine to support supporters remaining in play so the workaround is no longer necessary.

axpendix (approved):

PR #1053, opened 2020-12-30
SW -18

ufodynasty:


src/tcgwars/logic/impl/gen4/SecretWonders.groovy
```diff
@@ -513,9 +513,7 @@ public enum SecretWonders implements LogicCardInfo {
             energyCost F, C
             onAttack {
               damage 20 + 10 * self.numberOfDamageCounters
-              afterDamage {
-                heal 20, self
-              }
+              heal 20, self
```
axpendix, line 516: Hi, I thought heal was supposed to run after damage

PR #1063, opened 2021-01-01
Electrode Debugging

ufodynasty:


flagrama:
One thing I can say is please don't use attachEnergy. Use PlayEnergy like the Holon Pokemon do.

PR #1162, opened 2021-01-31
Always shuffle a hidden CardList on select

flagrama:


flagrama:
As Prize Cards shouldn't be shuffled, should this only be done for Hand specifically? Are the other CardLists ever likely to be hidden when selecting?

axpendix (approved):

PR #1166, opened 2021-02-17
Fix timing of effects that trigger when pokemon is damaged by an attack

link--11:
See #dev channel in the discord server. Currently effects that trigger when a Pokémon is damaged by an attack are activated before the damage and effects of that attack are applied. 
This is problematic if for example an attack heals the attacking Pokémon, that effect should resolve before an effect of an Ability that places damage on the attacking Pokémon. With the current implementation, a Pokémon with previously zero damage would be able to immediately heal the damage it shouldn't yet have during the resolution of the attack effect. Another relevant example is when an attack that discards an Energy faces against a triggered effect that sends back an Energy to the hand. The attacking player should choose the discard first, and only after that the defending players makes the choice on what to return, but currently it's the other way around.

I have no prior experience with this engine or the language, so not sure if this PR solves the issue completely, but based on other card implementations it should help I think, and at least provide a solid basis for changing all those same effects in one shared function.

### DIFF
--- a/src/tcgwars/logic/impl/gen7/LostThunder.groovy
+++ b/src/tcgwars/logic/impl/gen7/LostThunder.groovy
@@ -2406,15 +2406,24 @@ public enum LostThunder implements LogicCardInfo {
           resistance FIGHTING, MINUS20
           bwAbility "Mirror Counter" , {
             text "If this Pokémon is your Active Pokémon and is damaged by an attack from your opponent's Pokémon-GX or Pokémon-EX (even if this Pokémon is Knocked Out), put damage counters on the Attacking Pokémon equal to the damage done to this Pokémon."
+
+
             delayedA (priority:LAST) {
+              def counterDmg = 0
               before APPLY_ATTACK_DAMAGES, {
                 if(self.active && ef.attacker.owner != self.owner && (ef.attacker.pokemonGX || ef.attacker.pokemonEX)) {
                   bg.dm().each{
                     if(it.to == self && it.dmg.value) {
-                      bc "Mirror Counter countered ${ef.attacker}'s attack"
-                      directDamage(it.dmg.value, ef.attacker, Source.SRC_ABILITY)
+                      counterDmg = it.dmg.value
                     }
                   }
+                }
+              }
+              after APPLY_ATTACK_DAMAGES, {
+                if (counterDmg) {
+                  bc "Mirror Counter countered ${ef.attacker}'s attack"
+                  directDamage(counterDmg, ef.attacker, Source.SRC_ABILITY)
+                  counterDmg = 0
                 }
               }
             }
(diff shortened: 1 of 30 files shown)


src/tcgwars/logic/impl/gen7/UnifiedMinds.groovy
```diff
@@ -4398,10 +4372,19 @@ public enum UnifiedMinds implements LogicCardInfo {
...
+                  applyEffect = true
+                }
+              }
+              after APPLY_ATTACK_DAMAGES, {
+                if (applyEffect) {
+                  if (self.cards.contains(thisCard)) {
+                    bc "Giant Bomb activates."
+                    directDamage(100, ef.attacker, TRAINER_CARD)
+                  }
+                  applyEffect = false
```
flagrama, line 4387: Shouldn't this be using one of the new statics?

src/tcgwars/logic/impl/gen7/LostThunder.groovy
```diff
@@ -2406,17 +2406,26 @@ public enum LostThunder implements LogicCardInfo {
...
           bwAbility "Mirror Counter" , {
             text "If this Pokémon is your Active Pokémon and is damaged by an attack from your opponent's Pokémon-GX or Pokémon-EX (even if this Pokémon is Knocked Out), put damage counters on the Attacking Pokémon equal to the damage done to this Pokémon."
+
+
             delayedA (priority:LAST) {
+              def counterDmg = 0
               before APPLY_ATTACK_DAMAGES, {
                 if(self.active && ef.attacker.owner != self.owner && (ef.attacker.pokemonGX || ef.attacker.pokemonEX)) {
                   bg.dm().each{
                     if(it.to == self && it.dmg.value) {
```
flagrama, line 2416: 1
```suggestion
            delayedA (priority:LAST) {
              def counterDmg = 0
              before APPLY_ATTACK_DAMAGES, {
                bg.dm().each{
                  if(it.to == self && it.dmg.value) {
```

src/tcgwars/logic/impl/gen7/LostThunder.groovy
```diff
@@ -2406,17 +2406,26 @@ public enum LostThunder implements LogicCardInfo {
...
-                      bc "Mirror Counter countered ${ef.attacker}'s attack"
-                      directDamage(it.dmg.value, ef.attacker, Source.SRC_ABILITY)
+                      counterDmg = it.dmg.value
```
flagrama, line 2417: 2
```suggestion
                    counterDmg = it.dmg.value
```

src/tcgwars/logic/impl/gen7/LostThunder.groovy
```diff
@@ -2406,17 +2406,26 @@ public enum LostThunder implements LogicCardInfo {
...
                   }
                 }
               }
+              after APPLY_ATTACK_DAMAGES, {
+                if (counterDmg) {
+                  bc "Mirror Counter countered ${ef.attacker}'s attack"
+                  directDamage(counterDmg, ef.attacker, Source.SRC_ABILITY)
+                  counterDmg = 0
+                }
+              }
```
flagrama, line 2428: 3
Might as well use the static here too.

**However** I'd actually recommend adding another closure to the static that gets called before applyEffect gets set to True so you can affect both parts of the static freely.
```suggestion
              ifActiveAndDamagedByAttackAttached {
                if (counterDmg) {
                  bc "Mirror Counter countered ${ef.attacker}'s attack"
                  directDamage(counterDmg, ef.attacker, Source.SRC_ABILITY)
                  counterDmg = 0
                }
              }
```

flagrama (commented):
I'm not a huge fan of the closure being at the beginning instead of the end. I also think the statics should accept two closures, one for the `after` being the main effect, then after, an optional one for the `before` since I believe they will be used less often, but may be needed to gather additional "should this effect fire" info.

src/tcgwars/logic/impl/gen7/UnifiedMinds.groovy
axpendix: It has a >180 damage condition

PR #1168, opened 2021-02-20
Power Keepers: Sidney's Stadium (82): Fix onPlay

flagrama:
Should also remove the statuses when first put in play.


luongthomas (approved):
Looks good to me 👍

PR #1170, opened 2021-02-24
EX era bug fixes

TheAquaPiplup:
fixing mysterious fossil from sandstorm


TheAquaPiplup:
I forgot to label this most recent change
its making it so when blastoise ex uses energy rain it correctly detects the attachment is coming from the hand

luongthomas (approved):
Looks good, remember to test

PR #1171, opened 2021-02-26
Blastoise ex (FRLG): Fix Energy Rain

flagrama:
Should use the PLAY_FROM_HAND activation reason so that cards that rely on checking if
an energy was attached from the hand trigger properly.


flagrama:
This should have already been fixed in #1170 so I'm not sure why it's letting me make a PR. Not sure why it isn't reporting "no changes" or even a merge conflict.

flagrama:
It even says it was Closed with unmerged commits. I really don't understand this one.

PR #1177, opened 2021-03-07
fixed broken rusted shield and rusted sword

jlgrimes:
In all other instances where "hp" is used, the number is wrapped in parentheses. However they are not here, which is causing the effects of these tools to be broken.

### DIFF
--- a/src/tcgwars/logic/impl/gen8/ShinyStarV.groovy
+++ b/src/tcgwars/logic/impl/gen8/ShinyStarV.groovy
@@ -288,7 +288,7 @@ public enum ShinyStarV implements LogicCardInfo {
                 def conditions = it.from == self && it.to == opp.active
                 if (!conditions) return
                 bc "$thisCard +30"
-                it.dmg += hp 30
+                it.dmg += hp(30)
               }
             }
           }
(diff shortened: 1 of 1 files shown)


flagrama (approved):
Please remember to test this on the dev server once it is back up. https://dev.tcgone.net

flagrama:
> In all other instances where "hp" is used, the number is wrapped in parentheses. However they are not here, which is causing the effects of these tools to be broken.

Groovy actually has some unique rules for parenthesis that tends to make the code a little easier to read. This change didn't actually do anything because those rules mean that both the old code and the new code are equivalent. You can skip putting the parenthesis around the parameters of 1 method, however if you use methods as parameters, those require parenthesis still to avoid confusing the Groovy parser.

Thank you for trying to fix the issue though. Seems that it was actually because the DamageManager has no entries `before PROCESS_ATTACK_EFFECTS` and the effect should actually be triggerd `after PROCESS_ATTACK_EFFECTS` https://github.com/axpendix/tcgone-engine-contrib/blob/9f5012d750cac5a11528682a1938ae08978ac58d/src/tcgwars/logic/impl/gen8/ShinyStarV.groovy#L279-L284
https://github.com/axpendix/tcgone-engine-contrib/blob/9f5012d750cac5a11528682a1938ae08978ac58d/src/tcgwars/logic/impl/gen8/ShinyStarV.groovy#L285
https://github.com/axpendix/tcgone-engine-contrib/blob/9f5012d750cac5a11528682a1938ae08978ac58d/src/tcgwars/logic/impl/gen8/ShinyStarV.groovy#L286-L299

PR #1180, opened 2021-03-13
SM70: Owner should choose, not opponent

flagrama:
Fixes https://forum.tcgone.net/t/bug-report-shining-ho-oh-smp-70-ho-ohs-ability-lets/900

Also cleans up the interaction in general to be a bit more user-friendly.

### DIFF
--- a/src/tcgwars/logic/impl/gen7/SunMoonPromos.groovy
+++ b/src/tcgwars/logic/impl/gen7/SunMoonPromos.groovy
@@ -1145,10 +1145,14 @@ public enum SunMoonPromos implements LogicCardInfo {
             text "If this Pokémon is your Active Pokémon and is Knocked Out by damage from an opponent's attack, move up to 2 basic Energy cards from this Pokémon to your Benched Pokémon in any way you like."
             delayedA {
               before (KNOCKOUT,self) {
-                if(self.active && (ef as Knockout).byDamageFromAttack && bg.currentTurn==self.owner.opposite && self.owner.pbg.bench.notEmpty && self.cards.filterByType(ENERGY)) {
-                  bc "Golden Wing activates"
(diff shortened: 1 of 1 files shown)


src/tcgwars/logic/impl/gen7/SunMoonPromos.groovy
```diff
@@ -1145,10 +1145,14 @@ public enum SunMoonPromos implements LogicCardInfo {
             text "If this Pokémon is your Active Pokémon and is Knocked Out by damage from an opponent's attack, move up to 2 basic Energy cards from this Pokémon to your Benched Pokémon in any way you like."
             delayedA {
               before (KNOCKOUT,self) {
-                if(self.active && (ef as Knockout).byDamageFromAttack && bg.currentTurn==self.owner.opposite && self.owner.pbg.bench.notEmpty && self.cards.filterByType(ENERGY)) {
-                  bc "Golden Wing activates"
-                  moveEnergy(basic: true, self, self.owner.pbg.bench)
```
ufodynasty, line 1150: There is a playerType parameter built into moveEnergy. Would it be more appropriate to use `moveEnergy(basic: true, playerType: self.owner, self, self.owner.pbg.bench)`

ufodynasty (approved):

src/tcgwars/logic/impl/gen7/SunMoonPromos.groovy
flagrama: moveEnergy requires choosing an Energy from one Pokemon, and choosing another Pokemon to move it to, and you have to do it twice with this ability. I made these changes so you just select the two energies you want to move, then choose the Pokemon to attach each energy to. While it only reduces the number of clicks by one, I think it feels better this way.

I can go back to using two moveEnergies of course, that was my first attempt at fixing this. I just felt that the extra click was unnecessary.

PR #1181, opened 2021-03-13
CES109: Count the Energy, not the Cards

flagrama:
Ruling: https://compendium.pokegym.net/compendium-bw.html#601

Fixes https://forum.tcgone.net/t/bug-report-rayquaza-gri-106-venusaur-slg-3-the-gr/920


ufodynasty (approved):
Q. Does Venusaur's "Jungle Totem" Ability boost the damage output of Rayquaza GX's "Dragon Break" attack since "each Basic {G} Energy attached to your Pokemon provides {G}{G} Energy"?
A. Yes, just because the Basic Energy card is now providing two units of energy doesn't change the fact that it's still Basic Energy. (Aug 9, 2018 TPCi Rules Team)

PR #1186, opened 2021-03-15
Metronome static: Handle BREAK and LvlUp cards, allow skip, implement

flagrama:


src/tcgwars/logic/groovy/TcgStatics.groovy
```diff
@@ -1844,16 +1847,45 @@ class TcgStatics {
...
+    delegate.getterA GET_MOVE_LIST, delegate.self, {holder->
+      def moves = [] as Set
+      moves.addAll holder.object
+      target.call().each {
+        if (it == holder.effect.target) return
+        if (it instanceof Card) {
+          moves.addAll it.moves
+        }
+        if (it instanceof PokemonCardSet) {
+          moves.addAll bg.em().activateGetter(new GetMoveList(it))
```
axpendix, line 1884: I suspect this can cause endless loop if two pokemon target each other e.g. Arceus Omniscient. Perhaps only add `it.topPokemonCard?.moves`?

src/tcgwars/logic/impl/gen7/UnifiedMinds.groovy
```diff
@@ -1775,25 +1775,8 @@ public enum UnifiedMinds implements LogicCardInfo {
...
-                  if(!cardList.contains("${it}")){
-                    cardList.add("${it}")
-                    perfectionMoves.addAll(it.moves)
-                  }
-                }
-              }
-              holder.object.addAll(perfectionMoves)
-            }
+            metronomeA delegate, { my.bench.findAll { it.pokemonGX || it.pokemonEX } }
+            metronomeA delegate, { my.discard.findAll { it.cardTypes.isIn(POKEMON_EX, POKEMON_GX) } }
```
axpendix, line 1779: Impressive

src/tcgwars/logic/groovy/TcgStatics.groovy
axpendix: sounds ugly.. then perhaps add a self.active guard at the top somewhere..

flagrama: Makes sense. I've added a check that the target is active, otherwise it returns without changing anything. Seems to work fine with 4 Arceus LV X on board.

axpendix (approved):

PR #1189, opened 2021-03-24
Retooled bugfixes2

ufodynasty:
Minor fixes and optimizations

-Effects that return a pokemon to your hand no longer cause payback effects to crash the game

- discardDefendingEnergyAfterDamage now lets the attacking player select the energy instead 

- Lanturn's Energy Grounding allows the player who owns lanturn to select which energy is moved
- Zacian V allows you to select metal energies the first time it shows you your top 3 cards
- LOT beautifly only hits for 30 damage instead of 100 (kinda want to leave this one in)
- Desert Geyser is no longer spelled Desert Geizer

### DIFF
--- a/src/tcgwars/logic/impl/gen3/PowerKeepers.groovy
+++ b/src/tcgwars/logic/impl/gen3/PowerKeepers.groovy
@@ -620,6 +620,7 @@ public enum PowerKeepers implements LogicCardInfo {
             before KNOCKOUT, {
               if (bg.em().retrieveObject("Energy_Grounding") != bg.turnCount && (ef as Knockout).byDamageFromAttack && bg.currentTurn==self.owner.opposite && ef.pokemonToBeKnockedOut != self && ef.pokemonToBeKnockedOut.cards.filterByType(BASIC_ENERGY).energyCount(C)) {
                 moveEnergy(basic: true, playerType: self.owner, may: true, info: "Energy Grounding : You may move an energy from ${ef.pokemonToBeKnockedOut} to $self", ef.pokemonToBeKnockedOut,self,SRC_ABILITY)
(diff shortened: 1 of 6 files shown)


src/tcgwars/logic/impl/gen3/PowerKeepers.groovy
```diff
@@ -619,10 +619,7 @@ public enum PowerKeepers implements LogicCardInfo {
           delayedA{
             before KNOCKOUT, {
               if (bg.em().retrieveObject("Energy_Grounding") != bg.turnCount && (ef as Knockout).byDamageFromAttack && bg.currentTurn==self.owner.opposite && ef.pokemonToBeKnockedOut != self && ef.pokemonToBeKnockedOut.cards.filterByType(BASIC_ENERGY).energyCount(C)) {
-                if (confirm("Move an energy from ${ef.pokemonToBeKnockedOut} to $self?", self.owner)) {
-                  bg.em().storeObject("Energy_Grounding", bg.turnCount)
```
flagrama, line 623: Don't you still need this to avoid attaching energy from multiple knockouts in the same turn?

ufodynasty: I think storing the turn count will only allow one activation per turn or are you talking about something else?

flagrama: Where is it storing the turn count now that this line has been removed?

ufodynasty: TY! I'm actually blind.

axpendix (approved):
Great job

PR #1191, opened 2021-03-25
Attach tool

flagrama:
This should fix many issues with cards that disable tool effects, effects that cheat tools onto in-play Pokémon, and Pokémon that turn themselves into tool cards and their interactions.

Depends on https://github.com/axpendix/tcgone/pull/154


src/tcgwars/logic/impl/gen3/UnseenForces.groovy
```diff
@@ -3529,15 +3529,14 @@ public enum UnseenForces implements LogicCardInfo {
...
+            assert my.all.findAll({canAttachPokemonTool(it)}) : "Your Pokémon already have tools attached to them."
             assert my.deck : "Deck is empty."
           }
           onAttack {
             def tar = my.all.findAll({!(it.cards.filterByType(POKEMON_TOOL))})
             if(tar){
               my.deck.search(max : Math.min(2,tar.size()),"Search for up to 2 Pokémon tool",cardTypeFilter(POKEMON_TOOL)).each{
-                def pcs = my.all.findAll({!(it.cards.filterByType(POKEMON_TOOL))}).select()
-                my.deck.remove(it)
+                def pcs = my.all.findAll({canAttachPokemonTool(it)}).select()
```
axpendix, line 3539:
```suggestion
                def pcs = my.all.findAll{canAttachPokemonTool(it)}.select()
                my.deck.remove(it)
```

src/tcgwars/logic/impl/gen4/GreatEncounters.groovy
```diff
@@ -1836,10 +1835,17 @@ public enum GreatEncounters implements LogicCardInfo {
...
+                }
               }
               bg.em().run(new ChangeImplementation(toolCard, top))
               toolCard.player = top.player
-              attachPokemonTool(toolCard, my.all.select("Attach to?"))
+              def pcs = my.all.findAll {it!=self && canAttachPokemonTool(it)}.select("Attach to?")
+              removeFromPlay(self, [top] as CardList)
+              bg.em().run(new ChangeImplementation(toolCard, top))
+              attachPokemonTool(toolCard, pcs)
+              removePCS(self)
```
axpendix, line 1848: removePCS() -> attachPokemonTool() seems simpler to me. is there any reason for picking this solution?

src/tcgwars/logic/impl/gen7/LostThunder.groovy
```diff
@@ -1513,7 +1513,6 @@ public enum LostThunder implements LogicCardInfo {
               if(tar){
                 tar.each {pcs->
                   my.deck.search("Select a Pokémon Tool to attach to $pcs",cardTypeFilter(POKEMON_TOOL)).each{
-                    deck.remove(it)
```
axpendix, line 1516: ?

src/tcgwars/logic/impl/gen4/GreatEncounters.groovy
flagrama: I've made changes to the main-engine side of these changes that may be relevant. Basically, I am/want to use MoveCard instead of manually adding to/removing from card lists, because I've not seen that be anything but trouble while trying to implement certain cards.

src/tcgwars/logic/impl/gen7/LostThunder.groovy
flagrama: I've made changes to the main-engine side of these changes that may be relevant. Basically, I am/want to use MoveCard instead of manually adding to/removing from card lists, because I've not seen that be anything but trouble while trying to implement certain cards.

src/tcgwars/logic/impl/gen4/GreatEncounters.groovy
flagrama: If using MoveCard in AttachPokemonTool, the card still needs to exist to be moved. Removing the PCS will remove the PCS entirely so the card will no longer exist on the field to be moved if removePCS is done first.

axpendix (approved):

PR #1192, opened 2021-03-25
UPR89: Don't show useless dialog to click through

flagrama:
If there are no more M energy in the discard pile this card would
display the player's Pokémon to choose from that would have to be
clicked through to finish the attack. Instead, this will now just
skip any remaining checks if the discard pile no longer has any M
energy in it.

This should "fix" https://forum.tcgone.net/t/bug-report-solgaleo-prism-star-upr-89-attempting-to-u/2163


luongthomas (approved):
Good solution

PR #1194, opened 2021-03-26
BUS:80 Should only copy Basic Pokémon moves

flagrama:
Whoops. Seems to be the only metrnomeA I messed up, but could be others.


luongthomas (approved):
Looks like a straightforward fix 👍

PR #1195, opened 2021-03-26
EXP29: Use a getter instead of modifying energies

flagrama:
Should fix https://forum.tcgone.net/t/bug-report-arcanine-skr-3-typhlosion-exp-28-scree/2495


luongthomas (approved):
Looks good 👍

PR #1204, opened 2021-03-30
Corv vmax bst110

flagrama:
Went with a different approach than is usually used. Hoping this fixes most cases where using your own ability on the Pokémon doesn't trigger the ability.


flagrama:
[CORVIKNIGHT_VMAX_110](https://trello.com/c/G9JTixs9/1024-corviknightvmax110)

luongthomas (approved):
Looks promising

PR #1210, opened 2021-04-01
Battle Style Various Fixes

luongthomas:


src/tcgwars/logic/impl/gen8/BattleStyles.groovy
```diff
@@ -424,16 +424,12 @@ public enum BattleStyles implements LogicCardInfo {
...
-                [POKEMON_V, VMAX, POKEMON_GX, POKEMON_EX].contains(it)
-              }
-            }
+            def validTargets = my.all.findAll { !it.ruleBox }
             assert validTargets : "No valid targets"
             powerUsed()
             def selEnergy = my.hand.filterByBasicEnergyType(G).first()
             def pcs = validTargets.select("Attach to?")
-            attachEnergy(pcs, selEnergy, PLAY_FROM_HAND)
+            attachEnergy(pcs, selEnergy)
```
flagrama, line 432: This was correct. The card is from the hand, therefore it is being played from the hand.

src/tcgwars/logic/impl/gen8/BattleStyles.groovy
```diff
@@ -688,7 +684,9 @@ public enum BattleStyles implements LogicCardInfo {
           attackRequirement {}
           onAttack {
             damage 10
-            if (defending.topPokemonCard.cardTypes.is(POKEMON_V)) {
+            if (defending.topPokemonCard.cardTypes.any {
+              [POKEMON_V, VMAX].contains(it)
+            }) {
```
flagrama, line 689: Use the helper methods in PokemonCardSet
```suggestion
            if (defending.pokemonV) {
```

src/tcgwars/logic/impl/gen8/BattleStyles.groovy
```diff
@@ -2553,7 +2551,16 @@ public enum BattleStyles implements LogicCardInfo {
...
-            // TODO
+            checkLastTurn()
+            assert hand.filterByNameEquals("Aegislash")
+            powerUsed()
+            def card = hand.filterByNameEquals("Aegislash").select("Stance Change").first()
+            def old = self.topPokemonCard
+            self.cards.remove(old)
+            hand.remove(card)
+            self.cards.add(card)
+            hand.add(old)
```
flagrama, line 2562: This will not be handled properly by Scoop Up Block. Current Ageislash should be discarded instead of put to hand if Scoop Up Block is active, but the one in the hand should still replace it.

src/tcgwars/logic/impl/gen8/BattleStyles.groovy
```diff
@@ -2575,7 +2582,16 @@ public enum BattleStyles implements LogicCardInfo {
...
```
flagrama, line 2593: Same here re:Scoop Up Block

src/tcgwars/logic/impl/gen8/BattleStyles.groovy
```diff
@@ -2805,18 +2829,16 @@ public enum BattleStyles implements LogicCardInfo {
...
-//                if (it.from == self && my.prizeCardSet.takenCount > 0 && it.dmg.value && it.notNoEffect) {
-//                  def plusDmg = 20*my.prizeCardSet.takenCount
-//                  bc "Beast Boost +$plusDmg"
-//                  it.dmg += hp(plusDmg)
-//                }
-//              }
-//            }
-//          }
+          delayedA {
+            before APPLY_ATTACK_DAMAGES, {
```
flagrama, line 2833: APPLY_ATTACK_DAMAGES is after Weak/Resist.

```suggestion
            after PROCESS_ATTACK_EFFECTS, {
```

src/tcgwars/logic/impl/gen8/BattleStyles.groovy
```diff
@@ -2908,8 +2930,12 @@ public enum BattleStyles implements LogicCardInfo {
         onPlay {
           my.all.each {
             if (it.topPokemonCard.cardTypes.is(EVOLUTION) && it.numberOfDamageCounters) {
```
flagrama, line 2932:
```suggestion
            if (it.realEvolution && it.numberOfDamageCounters) {
```

src/tcgwars/logic/impl/gen8/BattleStyles.groovy
```diff
@@ -3145,7 +3190,7 @@ public enum BattleStyles implements LogicCardInfo {
         def eff
         onPlay {
           eff = getter (GET_RETREAT_COST) { holder ->
-            if ((holder.effect.target as PokemonCardSet).topPokemonCard.cardTypes.is(SINGLE_STRIKE)) {
+            if ((holder.effect.target as PokemonCardSet).topPokemonCard.cardTypes.is(RAPID_STRIKE)) {
```
flagrama, line 3193:
```suggestion
            if ((holder.effect.target as PokemonCardSet).rapidStrike) {
```

flagrama (commented):
If you are explicitly checking the topPokemonCard, you should just use the PokemonCardSet helper methods instead of explict CardType.

src/tcgwars/logic/impl/gen8/BattleStyles.groovy
luongthomas: I think the new code I've added will address this :)

flagrama: Now you're just discarding the top card instead of putting it in your hand from what I see. 

You should either try to `moveTo` the hand and check if the card is still in the `self.cards` list and discard it if it is, or try to `moveCard` and it should return `true` on failure and `false` on success and you can discard the card if it fails I think.

PR #1216, opened 2021-04-09
Implements Rising Rivals

ufodynasty:


src/tcgwars/logic/impl/gen4/RisingRivals.groovy
```diff
@@ -247,15 +267,23 @@ public enum RisingRivals implements LogicCardInfo {
           resistance P, MINUS20
           pokeBody "Eerie Aura", {
             text "Put 1 damage counter on each of your opponent’s Pokémon that remains Asleep between turns."
-            delayedA {
+            before BEGIN_TURN, {
+              if(my.active.isSPC(ASLEEP)){
+                directDamage 10, self, SRC_ABILITY
+              }
```
flagrama, line 273: This Body says "each of **your opponent's Pokémon**"

Nothing about your Pokémon

src/tcgwars/logic/impl/gen4/RisingRivals.groovy
```diff
@@ -266,17 +294,19 @@ public enum RisingRivals implements LogicCardInfo {
           move "Incite", {
             text "Search your discard pile for up to 2 Supporter cards, show them to your opponent, and put them into your hand."
             energyCost ()
-            attackRequirement {}
+            attackRequirement {
+              assert my.discard.filterByType(SUPPORTER) : "You have no supporter cards in your discard pile"
+            }
             onAttack {
-              damage 0
+              my.discard.select(max:2, "Search you discard pile for up to 2 Supporter cards",cardTypeFilter(SUPPORTER)).showToOpponent("Incite: These cards will be returned to your opponent's hand").moveTo(my.hand)
```
flagrama, line 301: You need a minimum of 1 if the text is "up to". Once you declare the attack you must do the effect.

src/tcgwars/logic/impl/gen4/RisingRivals.groovy
```diff
@@ -287,23 +317,29 @@ public enum RisingRivals implements LogicCardInfo {
...
           move "Sand Wall", {
             text "40 damage. Discard a Stadium card your opponent has in play. If you do, prevent all effects of an attack, including damage, done to Flygon during your opponent’s next turn."
             energyCost C, C
-            attackRequirement {}
             onAttack {
-              damage 0
+              if(bg.stadiumInfoStruct && bg.stadiumInfoStruct.stadiumCard.player == self.owner.opposite){
+                preventAllEffectsNextTurn()
+                afterDamage {
+                  discard bg.stadiumInfoStruct.stadiumCard
```
flagrama, line 333: There is a static for doing this which may be a better idea if stadium stuff ever needs to be changed in the future.

src/tcgwars/logic/impl/gen4/RisingRivals.groovy
```diff
@@ -335,21 +377,44 @@ public enum RisingRivals implements LogicCardInfo {
...
+                    h.object = h.object + 1
+                  }
+                }
+                unregister {
+                  eff.unregister()
+                }
+                unregisterAfter 1
+              }
+              def card = my.playedSupporter.select("Select a Supporter to copy its effect as this attack.").first()
+              discard card
```
flagrama, line 409: Isn't this already handled by PlayTrainer?

src/tcgwars/logic/impl/gen4/RisingRivals.groovy
```diff
@@ -362,14 +427,38 @@ public enum RisingRivals implements LogicCardInfo {
...
+                    if (it.from.active && it.to.active && it.dmg.value && flag) {
+                      for(Type t1:Type.values()){//Doing it this way in case of 4x weakness
+                        if(it.from.types.contains(t1) && it.to.weaknesses.find{it.type == t1}){
+                          it.dmg *= 2
+                        }
+                      }
+                    }
+                  }
+                }
+              }
```
flagrama, line 453: Use a GET_WEAKNESS getter for this. Check BaseSetNG PORYGON to see how you can create a new object with the X2 feature but the same weakness.

src/tcgwars/logic/impl/gen4/RisingRivals.groovy
```diff
@@ -362,14 +427,38 @@ public enum RisingRivals implements LogicCardInfo {
...
             }
           }
           move "Combo Throw", {
             text "30+ damage. Does 30 damage plus 10 more damage for each Energy attached to the Defending Pokémon."
             energyCost M, C, C
-            attackRequirement {}
             onAttack {
-              damage 0
+              damage
```
flagrama, line 460: That's not gonna work :P

src/tcgwars/logic/impl/gen4/RisingRivals.groovy
```diff
@@ -403,17 +496,21 @@ public enum RisingRivals implements LogicCardInfo {
           move "Psychic Removal", {
             text "10 damage. Flip 2 coins. If both of them are heads, discard all Energy attached to the Defending Pokémon."
             energyCost P
-            attackRequirement {}
             onAttack {
-              damage 0
+              damage 10
+              flip 2, {}, {}, [2:{
+                targeted (defending) {
+                  opp.active.cards.filterByType(ENERGY).discard()
```
flagrama, line 503: Should be afterDamage

src/tcgwars/logic/impl/gen4/RisingRivals.groovy
```diff
@@ -422,19 +519,32 @@ public enum RisingRivals implements LogicCardInfo {
...
           weakness G
           move "Trample", {
-            text ". If that coin flip is heads, this attack does 30 damage to that Pokémon."
+            text "Flip a coin for each Benched Pokémon. If that coin flip is heads, this attack does 30 damage to that Pokémon."
             energyCost F, C
             attackRequirement {}
             onAttack {
-              damage 0
+              my.bench.each {
+                bc"$it"
```
flagrama, line 527: The damage should already state the Pokémon. You could instead make this part of the info string for the flip though. Broadcasting just the Pokémon name here is useless log spam though.

src/tcgwars/logic/impl/gen4/RisingRivals.groovy
```diff
@@ -467,22 +581,48 @@ public enum RisingRivals implements LogicCardInfo {
...
             text "As long as Shiftry is your Active Pokémon, whenever your opponent flips a coin during his or her turn, treat it as tails."
             delayedA {
+              def doit = {
+                if (bg.currentThreadPlayerType != self.owner) {
+                  bc "Hidden Power forced the coinflip to be TAILS."
+                  bg.deterministicCoinFlipQueue.offer(false)
+                }
+              }
+              before COIN_FLIP, {doit()}
+              before COIN_FLIP_GETTER, {doit()}
```
flagrama, line 591: Should be a COIN_FLIP_GETTER getter.

src/tcgwars/logic/impl/gen4/RisingRivals.groovy
```diff
@@ -467,22 +581,48 @@ public enum RisingRivals implements LogicCardInfo {
...
+                applyAfterDamage CONFUSED
+              }
             }
           }
           move "Seal Off", {
             text "60 damage. The Defending Pokémon can’t use any Poké-Powers or Poké-Bodies during your opponent’s next turn."
             energyCost D, C, C
-            attackRequirement {}
             onAttack {
-              damage 0
```
flagrama, line 485: This attack should still do 60 damage.

src/tcgwars/logic/impl/gen4/RisingRivals.groovy
```diff
@@ -491,28 +631,45 @@ public enum RisingRivals implements LogicCardInfo {
...
+          customAbility {
+            delayed (priority: LAST) {
+              before APPLY_ATTACK_DAMAGES, {
+                if(bg().currentTurn==self.owner.opposite) {
+                  turnCount=bg.turnCount
+                  lastDamage=bg().dm().find({it.to==self && it.dmg.value>=0})?.dmg
+                }
+              }
+            }
+          }
```
flagrama, line 643: This really seems like something that should be on the PokemonCardSet somehow and not a customAbility.

src/tcgwars/logic/impl/gen4/RisingRivals.groovy
```diff
@@ -491,28 +631,45 @@ public enum RisingRivals implements LogicCardInfo {
...
-            attackRequirement {}
             onAttack {
-              damage 0
+              damage 40
+              if(my.deck && confirm("Discard the top card of your deck?")) {
+                afterDamage {
+                  discard my.deck.first()
+                }
+                heal 20, self
+                clearSpecialCondition(self)
```
flagrama, line 664: Shouldn't these also be in the afterDamage closure?

src/tcgwars/logic/impl/gen4/RisingRivals.groovy
```diff
@@ -521,16 +678,30 @@ public enum RisingRivals implements LogicCardInfo {
...
             text "30 damage. The Defending Pokémon is now Paralyzed and Poisoned. Ignore this effect if any of your Pokémon used Needle Shock during your last turn."
             energyCost G, C, C
-            attackRequirement {}
             onAttack {
-              damage 0
+              damage 30
+              afterDamage {
+                bc"${bg.em().retrieveObject("Needle_Shock")}"
+                bc"$bg.turnCount"
+                if(bg.em().retrieveObject("Needle_Shock")!=(bg.turnCount - 2)) {
```
flagrama, line 699: The key needs to include the player, otherwise your opponent using Needle Shock will overwrite you using it last turn.

src/tcgwars/logic/impl/gen4/RisingRivals.groovy
```diff
@@ -521,16 +678,30 @@ public enum RisingRivals implements LogicCardInfo {
...
           move "Needle Shock", {
             text "30 damage. The Defending Pokémon is now Paralyzed and Poisoned. Ignore this effect if any of your Pokémon used Needle Shock during your last turn."
             energyCost G, C, C
-            attackRequirement {}
             onAttack {
-              damage 0
+              damage 30
+              afterDamage {
+                bc"${bg.em().retrieveObject("Needle_Shock")}"
+                bc"$bg.turnCount"
```
flagrama, line 698: Debug statements should be removed.

src/tcgwars/logic/impl/gen4/RisingRivals.groovy
```diff
@@ -659,22 +850,37 @@ public enum RisingRivals implements LogicCardInfo {
           pokeBody "Sticky Hold", {
             text "If Gastrodon East Sea is switched or retreats to your Bench, move as many Energy cards attached to Gastrodon East Sea as you like to the new Active Pokémon."
             delayedA {
+              after FALL_BACK, self, {
+                if(self.cards.energyCount(C) && confirm("Use Sticky Hold?")) {
+                  def energyList = self.cards.select(max:self.cards.filterByType(ENERGY).size(),"Move any number of Energy cards attached from $self to ${self.owner.pbg.active}",cardTypeFilter(ENERGY))
+                  energyList.each {
+                    energySwitch(self,self.owner.pbg.active,it)
```
flagrama, line 857: I recommend suppressing the energySwitch log and implementing a custom `bc` message to avoid spamming the log.

src/tcgwars/logic/impl/gen4/RisingRivals.groovy
```diff
@@ -659,22 +850,37 @@ public enum RisingRivals implements LogicCardInfo {
...
+            attackRequirement {
+              assert my.deck : "Your deck is empty"
+              assert my.bench.notFull : "Your bench is full"
+            }
             onAttack {
-              damage 0
+              def max = Math.min(2,my.bench.freeBenchCount)
+              my.deck.search(max:max,"Search your deck for up to 2 Gastrodon to put onto your bench",{it.cardTypes.is(POKEMON) && it.name.contains("Gastrodon")}).each {
+                def pcs = benchPCS(it)
+                directDamage 20, pcs, SRC_ABILITY
```
flagrama, line 874: Kinda unrelated
### 
Things like this makes thinking of how to implement auto source really annoying. PokemonPowers are not Abilities, so manually setting the source to Ability for them is kind of problematic. I suppose when we do do that we should just ignore all manual setting currently being done.

flagrama, line 874: This is not an ability.

src/tcgwars/logic/impl/gen4/RisingRivals.groovy
```diff
@@ -247,15 +267,23 @@ public enum RisingRivals implements LogicCardInfo {
           resistance P, MINUS20
           pokeBody "Eerie Aura", {
             text "Put 1 damage counter on each of your opponent’s Pokémon that remains Asleep between turns."
-            delayedA {
+            before BEGIN_TURN, {
+              if(my.active.isSPC(ASLEEP)){
+                directDamage 10, self, SRC_ABILITY
```
flagrama, line 272: This is also not an Ability. Gen4 is where we'll have to start paying attention to the fact that PokemonPowers and Abilities are completely different and have completely different rulings.

src/tcgwars/logic/impl/gen4/RisingRivals.groovy
```diff
@@ -812,16 +1059,32 @@ public enum RisingRivals implements LogicCardInfo {
           weakness M
           resistance L, MINUS20
           pokeBody "Icy Aura", {
-            text "As long as Mamoswine is your Active Pokémon, put 1 damage counter on each Active Pokémon (excluding Pokémon) between turns."
+            text "As long as Mamoswine GL is your Active Pokémon, put 1 damage counter on each Active Pokémon (excluding W Pokémon) between turns."
             delayedA {
+              before BEGIN_TURN, {
+                if(self.active && !self.owner.opposite.pbg.active.types.contains(W)||!self.types.contains(W)) {
```
flagrama, line 1065: Don't check these in the if statement if you're gonna have another if statement checking them anyway. Makes it less readable for 0 gain.

src/tcgwars/logic/impl/gen4/RisingRivals.groovy
```diff
@@ -881,22 +1167,30 @@ public enum RisingRivals implements LogicCardInfo {
           pokeBody "Maternal Comfort", {
             text "At any times between turns, remove 1 damage counter from each of your Pokémon. You can’t use more than 1 Maternal Comfort Poké-Body between turns."
             delayedA {
+              before BEGIN_TURN, {
+                if(bg.em().retrieveObject("Maternal Comfort")!=bg.turnCount && my.all.find{it.numberOfDamageCounters}) {
+                  bc "Maternal Comfort Activates"
+                  my.all.each {
+                    heal 10, it, SRC_ABILITY
```
flagrama, line 1174: This is also not an ability.

src/tcgwars/logic/impl/gen4/RisingRivals.groovy
```diff
@@ -952,17 +1252,25 @@ public enum RisingRivals implements LogicCardInfo {
           move "Pick and Collect", {
             text "Search your discard pile for up to 4 basic Energy cards, show them to your opponent, and put them into your hand."
             energyCost ()
-            attackRequirement {}
+            attackRequirement {
+              assert my.discard.filterByType(BASIC_ENERGY) : "You have no basic Energy cards in your discard pile"
+            }
             onAttack {
-              damage 0
+              my.discard.select(max:4,"Search your discard pile for up to 4 basic Energy cards",cardTypeFilter(BASIC_ENERGY)).showToOpponent("Pick and Collect: Selected cards").moveTo(my.hand)
```
flagrama, line 1259: Make sure it has a minimum of 1, once you declare using the ability, you have to do everything in it, and "up to" means at least 1.

src/tcgwars/logic/impl/gen4/RisingRivals.groovy
```diff
@@ -1016,16 +1348,26 @@ public enum RisingRivals implements LogicCardInfo {
...
           weakness M, PLUS30
           pokePower "Gather Ice", {
-            text "Once during your turn , when you play Walrein from your hand to evolve 1 of your Pokémon, you may as many Energy cards from your hand to Walrein as you like."
-            actionA {
+            text "Once during your turn , when you play Walrein from your hand to evolve 1 of your Pokémon, you may attach as many [W] Energy cards from your hand to Walrein as you like."
+            onActivate{reason ->
+              if(reason == PLAY_FROM_HAND && my.hand.filterByEnergyType(W) && confirm("Use Gather Ice?")){
+                powerUsed()
+                my.hand.select(min:0,max:my.hand.filterByEnergyType(W).size(),"Attach as many [W] Energy cards from your hand to $self as you like",energyFilter(W)).each {
+                  attachEnergy(self,it)
```
flagrama, line 1356: Reason should be PLAY_FROM_HAND

src/tcgwars/logic/impl/gen4/RisingRivals.groovy
```diff
@@ -1016,16 +1348,26 @@ public enum RisingRivals implements LogicCardInfo {
...
           move "Cold Crush", {
             text "70 damage. Discard an Energy card attached to Walrein and then discard an Energy attached to the Defending Pokémon."
             energyCost W, W, C, C
             attackRequirement {}
             onAttack {
-              damage 0
+              damage 70
+              afterDamage {
+                discardSelfEnergy C
+                discardDefendingEnergy C
```
flagrama, line 1369: You must select the energy before damage, then discard it after.

src/tcgwars/logic/impl/gen4/RisingRivals.groovy
```diff
@@ -1058,10 +1400,11 @@ public enum RisingRivals implements LogicCardInfo {
           weakness P
           move "Recover", {
             text "Energy attached to Alakazam and remove 4 damage counters from Alakazam ."
-            energyCost P, P
+            energyCost P
             attackRequirement {}
             onAttack {
-              damage 0
+              discardSelfEnergy P
+              heal 40, self
```
flagrama, line 1407: How is this supposed to work if I metronome? Should it always heal even if I can't discard a P energy?

src/tcgwars/logic/impl/gen4/RisingRivals.groovy
```diff
@@ -1125,22 +1484,35 @@ public enum RisingRivals implements LogicCardInfo {
           pokeBody "Frost Wind", {
             text "As long as Glaceon is your Active Pokémon, any damage done to your Pokémon by your opponent’s attacks is reduced by 10 ."
             delayedA {
+              before APPLY_ATTACK_DAMAGES, {
+                bg.dm().each {
+                  if(self.active && it.to.owner == self.owner && it.from.owner == self.owner.opposite && it.dmg.value && it.notNoEffect){
+                    bc "Frost Wind -10"
```
flagrama, line 1490: I really recommend using $thisAbility and $thisMove where appropriate. Makes it easier to copy and paste when relevant.

src/tcgwars/logic/impl/gen4/RisingRivals.groovy
```diff
@@ -1150,16 +1522,27 @@ public enum RisingRivals implements LogicCardInfo {
           weakness W
           resistance L, MINUS20
           pokeBody "Sand Armor", {
-            text "If Hippowdon 4 has any Energy attached to it, any damage done to Hippowdon 4 by attacks is reduced by 10 ."
+            text "If Hippowdon E4 has any Energy attached to it, any damage done to Hippowdon 4 by attacks is reduced by 10 ."
             delayedA {
+              before APPLY_ATTACK_DAMAGES, {
+                bg.dm().each {
+                  if(it.to == self && it.from.owner == self.owner.opposite && it.notNoEffect && it.dmg.value){
```
flagrama, line 1529: This is "by attacks" not "by opponent's attacks"

src/tcgwars/logic/impl/gen4/RisingRivals.groovy
```diff
@@ -1214,22 +1603,38 @@ public enum RisingRivals implements LogicCardInfo {
           pokeBody "Energy Refresh", {
             text "As long as Leafeon is your Active Pokémon, whenever you attach an Energy card from your hand to 1 of your Pokémon, remove 2 damage counters from that Pokémon."
             delayedA {
+              after ATTACH_ENERGY, {
+                if (bg.currentTurn == self.owner.opposite && self.active) {
```
flagrama, line 1607: You must check that it is PLAY_FROM_HAND

src/tcgwars/logic/impl/gen4/RisingRivals.groovy
```diff
@@ -1214,22 +1603,38 @@ public enum RisingRivals implements LogicCardInfo {
...
             attackRequirement {}
             onAttack {
-              damage 0
+              damage 40
+              afterDamage {
+                if(my.hand.filterByType(BASIC_ENERGY)) {
+                  def card = my.hand.select(min:0,"You may attack a basic Energy card from your hand to 1 of your Pokémon",cardTypeFilter(BASIC_ENERGY)).first()
+                  if(card) {
+                    def tar = my.all.select("Attach $card to which Pokémon?")
+                    attachEnergy(tar,card)
```
flagrama, line 1625: Must have reason PLAY_FROM_HAND

src/tcgwars/logic/impl/gen4/RisingRivals.groovy
```diff
@@ -1304,23 +1729,30 @@ public enum RisingRivals implements LogicCardInfo {
           resistance P, MINUS20
           pokeBody "Energy Shark", {
             text "If Sharpedo is your Active Pokémon and is damaged by an opponent’s attack , flip a coin. If heads, discard an Energy card attached to the Attacking Pokémon."
-            delayedA {
+            ifActiveAndDamagedByAttackBody(delegate){
+              if (ef.attacker.cards.filterByType(ENERGY)) {
+                bc "Energy Shark Activates"
+                flip {
+                  ef.attacker.cards.select("Select an Energy to move to the Opponent's hand", cardTypeFilter(ENERGY), self.owner).discard()
```
flagrama, line 1736: Wrong text. Gets discarded, not moved to hand.

src/tcgwars/logic/impl/gen4/RisingRivals.groovy
```diff
@@ -1382,23 +1831,29 @@ public enum RisingRivals implements LogicCardInfo {
...
             }
           }
           move "Blessed Fruit", {
-            text "Pokémon."
-            energyCost G, G, G
-            attackRequirement {}
+            text "Remove all damage counters from one of your benched [G] Pokémon."
+            energyCost G, G
+            attackRequirement {
+              assert my.all.find{it.types.contains(G) && it.numberOfDamageCounters} : "All of your benched [G] Pokémon are healthy"
```
flagrama, line 1844: Text isn't necessarily accurate. Could also be missing the Energy.

src/tcgwars/logic/impl/gen4/RisingRivals.groovy
```diff
@@ -1454,17 +1922,43 @@ public enum RisingRivals implements LogicCardInfo {
...
           move "Primal Breath", {
             text "30 damage. Your opponent can’t play any Pokémon from his or her hand to Evolve or to Level Up the Defending Pokémon during his or her next turn."
             energyCost F, C
             attackRequirement {}
             onAttack {
-              damage 0
+              damage 30
+              afterDamage {
+                targeted (defending) {
+                  bc "During ${opp.owner.getPlayerUsername(bg)}'s next turn, Energy can't be attached from their hand to the Defending ${defending}. (This effect can be removed by evolving or benching ${defending}.)"
```
flagrama, line 1940: Text is wrong.

You're also missing several EVOLVE effects you need to handle https://github.com/axpendix/tcgone-engine-contrib/blob/b9adba8e4aa995503c40134400f67ce87db63210/src/tcgwars/logic/impl/gen8/DarknessAblaze.groovy#L1322-L1347

src/tcgwars/logic/impl/gen4/RisingRivals.groovy
```diff
@@ -1573,15 +2078,18 @@ public enum RisingRivals implements LogicCardInfo {
...
-            text "40 damage. Energy attached to Flareon ."
-            energyCost R, C, R
+            text "40 damage. Flip a coin. If tails, discard a [R] Energy attached to Flareon E4."
+            energyCost R, C
             attackRequirement {}
             onAttack {
-              damage 0
+              damage 40
+              flip {
+                discardSelfEnergyAfterDamage(R)
```
flagrama, line 2091: This should be on tails, not heads.

src/tcgwars/logic/impl/gen4/RisingRivals.groovy
```diff
@@ -1591,19 +2099,33 @@ public enum RisingRivals implements LogicCardInfo {
...
-            text "can’t use Shell Scatter during your next turn."
+            text "This attack does 10 damage to each of your opponent's Benched Pokémon. Foretress G can’t use Shell Scatter during your next turn."
             energyCost C
-            attackRequirement {}
+            attackRequirement {
+              assert opp.bench : "Your opponent has no Benched Pokémon"
+            }
             onAttack {
-              damage 0
+              opp.bench.each {
```
flagrama, line 2108: Does not say Benched in the actual card text.

src/tcgwars/logic/impl/gen4/RisingRivals.groovy
```diff
@@ -1660,15 +2190,19 @@ public enum RisingRivals implements LogicCardInfo {
             energyCost C
             attackRequirement {}
             onAttack {
-              damage 0
+              damage 10
+              whirlwind()
```
flagrama, line 2194: I think whirlwind is actually backwards here. Which Pokémon gets switched with which is pretty important.

src/tcgwars/logic/impl/gen4/RisingRivals.groovy
```diff
@@ -1797,9 +2370,12 @@ public enum RisingRivals implements LogicCardInfo {
           move "Errand-Running", {
             text "Search your deck for a Trainer card, show it to your opponent, and put it into your hand. Shuffle your deck afterward."
             energyCost C, C
-            attackRequirement {}
+            attackRequirement {
+              assert my.deck : "Your deck is emtpy"
+            }
             onAttack {
-              damage 0
+              my.deck.search("Search your deck for a trainer card",cardTypeFilter(ITEM)).showToOpponent("Errand-Running: Selected card").moveTo(my.hand)
```
flagrama, line 2377: Maybe clarify that it should be the equivalent of an Item card here? This wording is going to be annoying either way though. I expect many invalid bug reports :laughing:

src/tcgwars/logic/impl/gen4/RisingRivals.groovy
```diff
@@ -1812,15 +2388,18 @@ public enum RisingRivals implements LogicCardInfo {
             energyCost P
             attackRequirement {}
             onAttack {
-              damage 0
+              damage 10
             }
           }
           move "Get Help", {
```
flagrama, line 2394: Card Text says "Offer Help"

src/tcgwars/logic/impl/gen4/RisingRivals.groovy
ufodynasty: I may be wrong, but I though select always has a minimum of 1 by default. Is there a case where not explicitly setting a minimum would allow for 0 cards to be selected?

src/tcgwars/logic/impl/gen4/RisingRivals.groovy
ufodynasty: Idk if im just blind, but the only statics I see that interacts with stadiums are whirlwind 2, discardStadium and the stadium can be affected check. As this effect depends both on whether or not a stadium card is in play, and who played the stadium, I don't see how any of those would work here.

src/tcgwars/logic/impl/gen4/RisingRivals.groovy
ufodynasty: :P

src/tcgwars/logic/impl/gen4/RisingRivals.groovy
ufodynasty: I agree, but I'm not sure it should be on the PCS as this effect and every other similar effect i've looked at don't allow you to reflect damage done to pre-evolutions

src/tcgwars/logic/impl/gen4/RisingRivals.groovy
flagrama: I think you might be right. I always remember in the back of my head that Deck searching has a minimum of 0, so I always try to double check when I see a select. I think this is fine though.

src/tcgwars/logic/impl/gen4/RisingRivals.groovy
flagrama: You can keep the check, but discard using the discard static :P

I also know this effect is used a couple of times, you can probably make a static for the whole effect too.

src/tcgwars/logic/impl/gen4/RisingRivals.groovy
flagrama: I'm not sure I understand. You're not checking for pre-evolutions anywhere here anyway?

Just keeping a variable on the PCS that gets updated with the last damage value actually done to it seems like a much better way to handle this than a customAbility too.

src/tcgwars/logic/impl/gen4/RisingRivals.groovy
flagrama: Looks like this is fine based on unofficial ruling.

src/tcgwars/logic/impl/gen4/RisingRivals.groovy
flagrama: Actually, it seems that WotC cards may have had it backwards? Whirlwind should definitely be double checked that it works right then. Might want to confirm with Jason if this order matters in WotC formats.

src/tcgwars/logic/impl/gen4/RisingRivals.groovy
flagrama: See how the static `wasSwitchedOutThisTurn` works, and I'd say try to emulate that.

src/tcgwars/logic/impl/gen4/RisingRivals.groovy
```diff
@@ -1928,11 +2520,14 @@ public enum RisingRivals implements LogicCardInfo {
...
-            text "40 damage. with 1 of your Benched Pokémon."
+            text "40 damage. You may switch Quagsire GL with 1 of your Benched Pokémon."
             energyCost W, C, C
             attackRequirement {}
             onAttack {
-              damage 0
+              damage 40
+              if(my.bench && confirm("Switch $self with one of your benched pokémon?")){
+                sw self,my.bench.select("New active")
+              }
```
flagrama, line 2530: switchYourActive static will handle all of this properly so would be preferred.

src/tcgwars/logic/impl/gen4/RisingRivals.groovy
```diff
@@ -1941,19 +2536,24 @@ public enum RisingRivals implements LogicCardInfo {
...
-            energyCost W
-            attackRequirement {}
+            text "Search your deck for up to 2 [W] Energy cards, show them to your opponent, and put them into your hand. Shuffle your deck afterward."
+            energyCost ()
+            attackRequirement {
+              assert my.deck : "Your deck is empty"
+            }
             onAttack {
-              damage 0
+              my.deck.search(max:2,"Search your deck for up to 2 [W] Energy cards",energyFilter(W)).moveTo(my.hand)
```
flagrama, line 2545: Shuffle that deck :laughing: 

Question for the future though, is there any reason that DeckCardList.search shouldn't Shuffle after each search? Maybe we could add that there if there aren't effects that would be affected by such a change.

src/tcgwars/logic/impl/gen4/RisingRivals.groovy
```diff
@@ -1967,15 +2567,29 @@ public enum RisingRivals implements LogicCardInfo {
             energyCost C
             attackRequirement {}
             onAttack {
-              damage 0
+              delayed {
```
flagrama, line 2570: I would definitely make this BEFORE_LAST to ensure calculation happens in the right order.

src/tcgwars/logic/impl/gen4/RisingRivals.groovy
```diff
@@ -2009,41 +2629,70 @@ public enum RisingRivals implements LogicCardInfo {
             energyCost ()
             attackRequirement {}
             onAttack {
-              damage 0
+              flip 2, {
+                def pcsList = my.all.findAll{it.numberOfDamageCounters}
+                if(pcsList) {
+                  heal 10, pcsList.select("Remove 1 damage counter from 1 of your Pokémon")
+                }
+              }
```
flagrama, line 2637: You're supposed to heal up to two damage from a single Pokémon, you should not be able to remove a damage counter from two different Pokémon. That is my understanding of the card text.

src/tcgwars/logic/impl/gen4/RisingRivals.groovy
```diff
@@ -2134,137 +2799,196 @@ public enum RisingRivals implements LogicCardInfo {
...
             def energyList = src.cards.filterByType(ENERGY).select(min:1, "Choose energy cards to move")
-            tar = my.all.findAll{it != src}.select("Choose Pokémon to move energy cards to").first()
+            def tar = my.all.findAll{it != src}.select("Choose Pokémon to move energy cards to")
             energyList.each {
               energySwitch(src, tar, it)
             }
           }
           playRequirement{
-            assert my.all.filterByType(ENERGY) : "You don't have any energy in play"
+            assert my.all.find{it.cards.energyCount(C)} : "You don't have any energy in play"
```
flagrama, line 2902: You also need a second Pokémon

src/tcgwars/logic/impl/gen4/RisingRivals.groovy
```diff
@@ -2134,137 +2799,196 @@ public enum RisingRivals implements LogicCardInfo {
...
+            }
+            eff2 = delayed {
+              before APPLY_RESISTANCE, {
+                bg.dm().each {
+                  if (it.from.types.contains(L) && it.to.active) {
+                    prevent()
+                  }
+                }
+              }
+            }
```
flagrama, line 2950: Does this work right? I think I'd prefer in instead `after PROCESS_ATTACK_EFFECTS` if you added the NO_RESISTANCE flag instead (see Phoebe from BattleStyles for reference adding a flag)

src/tcgwars/logic/impl/gen4/RisingRivals.groovy
```diff
@@ -2134,137 +2799,196 @@ public enum RisingRivals implements LogicCardInfo {
...
           onPlay {
+            my.hand.getExcludedList(thisCard).select("Choose a card to put back into your deck").moveTo(hidden:true,my.deck)
+            def card = my.deck.search("Search your deck for a Pokémon SP",cardTypeFilter(POKEMON_SP))
+            if(card) {
+              card.showToOpponent("SP Radar: Selected Card").moveTo(my.hand)
+            }
+            shuffleDeck()
           }
           playRequirement{
+            assert my.hand.getExcludedList(thisCard) : "You have no other cards in your hand"
```
flagrama, line 2977: Do you have any reference to ruling saying you can do this if you have no cards in your deck currently?

src/tcgwars/logic/impl/gen4/RisingRivals.groovy
```diff
@@ -2134,137 +2799,196 @@ public enum RisingRivals implements LogicCardInfo {
...
       case VOLKNER_S_PHILOSOPHY_98:
-        return basicTrainer (this) {
+        return supporter (this) {
           text "You can play only one Supporter card each turn. When you play this card, put it next to your Active Pokémon. When your turn ends, discard this card.\nYou may discard a card from your hand. Then, draw cards until you have 6 cards in your hand. (If you can’t draw any cards, you can’t play this card.)"
           onPlay {
+            my.hand.getExcludedList(thisCard).select(min:0,"You may discard a card from your hand").discard()
+            draw 6 - my.hand.getExcludedList(thisCard).size()
           }
           playRequirement{
+            assert my.hand.getExcludedList(thisCard).size() <= 6
```
flagrama, line 2990: You also need to be able to discard a card

src/tcgwars/logic/impl/gen4/RisingRivals.groovy
```diff
@@ -2301,31 +3027,73 @@ public enum RisingRivals implements LogicCardInfo {
...
           }
         };
       case FLOATZEL_GL_LV_X_104:
         return levelUp (this, from:"Floatzel GL", hp:HP100, type:WATER, retreatCost:0) {
           weakness L
           pokeBody "Water Rescue", {
-            text "Whenever any of your Pokémon (excluding any Floatzel) is Knocked Out by damage from your opponent’s attack, you may put that Pokémon and all cards that were attached to it from your discard pile into your hand."
-            delayedA {
+            text "Whenever any of your [W] Pokémon (excluding any Floatzel) is Knocked Out by damage from your opponent’s attack, you may put that Pokémon and all cards that were attached to it from your discard pile into your hand."
+            delayedA (priority: LAST) {
```
flagrama, line 3058: Can you confirm this works correctly with a Fossil card underneath the W Pokémon? The scoopUpPokemon static does a bit of heavy lifting to handle ChangeImplementation cards like Fossils. You may need to use some of that logic here.

src/tcgwars/logic/impl/gen4/RisingRivals.groovy
```diff
@@ -2336,31 +3104,45 @@ public enum RisingRivals implements LogicCardInfo {
...
         };
       case GALLADE_E4_LV_X_106:
         return levelUp (this, from:"Gallade 4", hp:HP100, type:PSYCHIC, retreatCost:1) {
           weakness P
           pokePower "Blade Storm", {
-            text "Once during your turn , when you put Gallade LV. from your hand onto your Active Gallade , you may put 1 damage counter on each of your opponent’s Pokémon."
-            actionA {
+            text "Once during your turn , when you put Gallade LV.X from your hand onto your Active Gallade , you may put 1 damage counter on each of your opponent’s Pokémon."
+            onActivate {r->
+              if (r==PLAY_FROM_HAND && opp.bench && confirm('Use Blade Storm?')) {
```
flagrama, line 3132: Why does opponent having a Bench matter?

src/tcgwars/logic/impl/gen4/RisingRivals.groovy
```diff
@@ -2371,56 +3153,75 @@ public enum RisingRivals implements LogicCardInfo {
...
             text "Once during your turn (before your attack), when you put Luxray GL LV.X from your hand onto your Active Luxray GL, you may switch the Defending Pokémon with 1 of your opponent's Benched Pokémon"
-            energyCost ()
-            attackRequirement {}
-            onAttack {
-              if(reason==PLAY_FROM_HAND && opp.bench && confirm("Use Bright Look?")) {
+            onActivate {r->
+              if(r==PLAY_FROM_HAND && opp.bench && confirm("Use Bright Look?")) {
                 powerUsed()
-                sw opp.active, opp.bench.select("Choose your opponent's new active Pokémon"), SRC_ABILITY
+                sw2 opp.active, opp.bench.select("Choose your opponent's new active Pokémon"), SRC_ABILITY
```
flagrama, line 3214: Why sw2 instead of sw?

src/tcgwars/logic/impl/gen4/RisingRivals.groovy
```diff
@@ -2431,14 +3232,22 @@ public enum RisingRivals implements LogicCardInfo {
...
           move "Darkness Magic", {
             text "Count the number of cards in your hand. Put that many damage counters on the Defending Pokémon. You can’t put more than 8 damage counters in this way."
             energyCost P, P, C
-            attackRequirement {}
+            attackRequirement {
+              assert my.hand : "Your hand is empty"
+            }
             onAttack {
-              damage 0
+              def count = Math.max(my.hand.size(),8)
```
flagrama, line 3249: You want min here.

src/tcgwars/logic/impl/gen4/RisingRivals.groovy
```diff
@@ -2448,14 +3257,26 @@ public enum RisingRivals implements LogicCardInfo {
           pokePower "Big Appetite", {
             text "Once during your turn , if Snorlax is your Active Pokémon, you may draw cards until you have 6 cards in your hand. If you do, Snorlax is now Asleep. This power can’t be used if Snorlax is affected by a Special Condition."
             actionA {
+              checkLastTurn()
+              checkNoSPC()
+              assert my.deck : "Your deck is empty"
+              assert my.hand.size() < 6 : "You already have 6 cards in your hand"
+              powerUsed()
+              draw 6 - my.hand.size()
+
```
flagrama, line 3266: "Snorlax is now Asleep"

src/tcgwars/logic/impl/gen4/RisingRivals.groovy
```diff
@@ -2605,24 +3491,49 @@ public enum RisingRivals implements LogicCardInfo {
...
             onAttack {
-              damage 0
+              damage 50
+              flip {
+                afterDamage {
+                  opp.all.each {
+                    if(it.cards.filterByType(ENERGY)) {
+                      it.cards.select("Choose an Energy card to discard",cardTypeFilter(ENERGY)).discard()
+                    }
+                  }
```
flagrama, line 3534: Make a new discardEnergyAfterDamage static if you need to, but this is incorrect behaviour I'm pretty sure.

src/tcgwars/logic/impl/gen4/RisingRivals.groovy
ufodynasty: That's a really good idea!

src/tcgwars/logic/impl/gen4/RisingRivals.groovy
ufodynasty: This does work, but I agree using a NO_RESISTANCE flag is better here

PR #1224, opened 2021-04-10
Rising rivals code review fixes

ufodynasty:
❤️ 🏴 ❤️

### DIFF
--- a/src/tcgwars/logic/impl/gen4/RisingRivals.groovy
+++ b/src/tcgwars/logic/impl/gen4/RisingRivals.groovy
@@ -3070,7 +3070,7 @@ public enum RisingRivals implements LogicCardInfo {
                   delayed(inline: true,priority: LAST){
                     after KNOCKOUT, pcs, {
                       if(confirm("Use Water Rescue?",self.owner)) {
-                        bc "W$thisAbility activates"
+                        bc "$thisAbility activates"
                         CardList returnList = []
                         cardList.each {
                           if(self.owner.pbg.discard.contains(it)) {
--- a/src/tcgwars/logic/impl/gen4/RisingRivals.groovy
+++ b/src/tcgwars/logic/impl/gen4/RisingRivals.groovy
@@ -2537,7 +2537,7 @@ public enum RisingRivals implements LogicCardInfo {
             }
             onAttack {
               my.deck.search(max:2,"Search your deck for up to 2 [W] Energy cards",energyFilter(W)).moveTo(my.hand)
-              shuffleDeck
+              shuffleDeck()
             }
           }
           move "Sheer Cold", {
@@ -3049,7 +3049,7 @@ public enum RisingRivals implements LogicCardInfo {
           weakness L
           pokeBody "Water Rescue", {
             text "Whenever any of your [W] Pokémon (excluding any Floatzel) is Knocked Out by damage from your opponent’s attack, you may put that Pokémon and all cards that were attached to it from your discard pile into your hand."
-            delayedA (priority: LAST) {
+            delayedA (priority: BEFORE_LAST) {
               before KNOCKOUT, {
                 def pcs=ef.pokemonToBeKnockedOut
                 if((ef as Knockout).byDamageFromAttack && pcs != self && pcs.owner == self.owner && pcs.types.contains(W)) {
@@ -3059,7 +3059,7 @@ public enum RisingRivals implements LogicCardInfo {
                     after KNOCKOUT, pcs, {
                       if(confirm("Use Water Rescue?",self.owner)) {
                         bc "$thisAbility activates"
-                        self.owner.pbg.discard.getExcludedList(dscrd).moveTo(self.owner.pbg.hand)
+                        self.owner.pbg.discard.getExcludedList(dscrd).moveTo(self.owner.pbg.hand)//This feels naive, but I haven't been able to break it yet
                       }
                       owner.delegate.unregister()
                     }
--- a/src/tcgwars/logic/impl/gen4/RisingRivals.groovy
+++ b/src/tcgwars/logic/impl/gen4/RisingRivals.groovy
@@ -3595,8 +3595,9 @@ public enum RisingRivals implements LogicCardInfo {
             }
             shuffleDeck()
             delayed {
-              before BETWEEN_TURNS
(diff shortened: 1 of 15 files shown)


src/tcgwars/logic/impl/gen4/RisingRivals.groovy
```diff
@@ -1929,6 +1912,15 @@ public enum RisingRivals implements LogicCardInfo {
...
           }
+          delayedA {
+            def warnAndPrevent = {
+              if (self.active && bg.currentTurn == self.owner.opposite) {
+                wcu "$self's Primal Law prevents playing Pokémon from your hand to evolve your Pokémon"
+                prevent()
+              }
+            }
+
+          }
```
flagrama, line 1923: I think you were using this for reference and forgot to remove it?

src/tcgwars/logic/impl/gen4/RisingRivals.groovy
```diff
@@ -1937,19 +1929,33 @@ public enum RisingRivals implements LogicCardInfo {
               damage 30
               afterDamage {
                 targeted (defending) {
-                  bc "During ${opp.owner.getPlayerUsername(bg)}'s next turn, Energy can't be attached from their hand to the Defending ${defending}. (This effect can be removed by evolving or benching ${defending}.)"
+                  bc "During ${opp.owner.getPlayerUsername(bg)}'s next turn, ${opp.owner.getPlayerUsername(bg)} can't play any Pokémon from their hand to Evolve or Level Up the defending $defending. (This effect can be removed by evolving or benching ${defending}.)"
```
flagrama, line 1932: Is this broadcast needed since a warning explaining the issue pops up when the affected player tries to play a relevant card?

flagrama, line 1932: Is this broadcast needed since a warning explaining the issue pops up when the affected player tries to play a relevant card?

ufodynasty: I like it personally for two reasons. Firstly it makes the effect more visible if it is ever involved in a bug report, and it makes scenarios where a player will search their deck for a card they can't play less likely.

src/tcgwars/logic/impl/gen4/RisingRivals.groovy
ufodynasty: Indeed :(

src/tcgwars/logic/impl/gen4/RisingRivals.groovy
flagrama: Hmm, I won't argue against it for now. Maybe when we get effect targeting in we can also add a list of effects affecting the player and their Pokémon where we'd put descriptions like these instead.

I don't think it'd come up in bug reports. The warning would make it pretty obvious why it was prevented, and if the reporter describes it well enough or we see the move that was used near the end of the log we'd be able to deduce the "issue" pretty quick.

src/tcgwars/logic/impl/gen4/RisingRivals.groovy
```diff
@@ -3607,6 +3603,11 @@ public enum RisingRivals implements LogicCardInfo {
               checkFaint()
             }
             shuffleDeck()
+            delayed {
+              before BETWEEN_TURNS
+              new CardList(thisCard).moveTo(thisCard.player.owner.pbg.hand)
+              unregisterAfter 1
+            }
```
flagrama, line 3610: What is this?

src/tcgwars/logic/impl/gen4/RisingRivals.groovy
```diff
@@ -3607,6 +3603,11 @@ public enum RisingRivals implements LogicCardInfo {
               checkFaint()
             }
             shuffleDeck()
+            delayed {
```
flagrama, line 3606: Regarding the above TODO. Maybe we could (if the Rotom is active)
1. Make the bench 1 bigger
2. Create a Fake PCS on the bench
3. Switch with the Fake PCS
4. Switch back to Active
5. Remove the Fake PCS
6. Reduce the bench size back

Most effects unregister after FALL_BACK after all.

src/tcgwars/logic/impl/gen4/RisingRivals.groovy
ufodynasty: This is what happens when you blow past the Ballmer peak

src/tcgwars/logic/impl/gen4/RisingRivals.groovy
```diff
@@ -2563,7 +2562,7 @@ public enum RisingRivals implements LogicCardInfo {
             energyCost C
             attackRequirement {}
             onAttack {
-              delayed {
+              delayed PRIORITY:BEFORE_LAST, {
```
flagrama, line 2565: Are you sure this works? PRIORITY: should not be a map key I'm pretty sure
```suggestion
              delayed BEFORE_LAST, {
```
should just work

src/tcgwars/logic/impl/gen4/RisingRivals.groovy
```diff
@@ -3055,20 +3053,13 @@ public enum RisingRivals implements LogicCardInfo {
               before KNOCKOUT, {
```
flagrama, line 3053: Water Rescue has some rulings that you should double check work properly and don't cause engine errors:
https://compendium.pokegym.net/compendium-lvx.html#410

src/tcgwars/logic/impl/gen4/RisingRivals.groovy
ufodynasty: This could work except there are other effects that trigger on retreat/switch. This is only my personal preference, but if the card will have bugs either way, I would prefer if they stemmed from an unimplemented feature rather than a janky implementation.

src/tcgwars/logic/impl/gen4/RisingRivals.groovy
ufodynasty: This does work.

src/tcgwars/logic/impl/gen4/RisingRivals.groovy
flagrama: Yeah, I agree. It was just an idea I had thought of, but you make a good point.

PR #1225, opened 2021-04-10
Bst fix

flagrama:


src/tcgwars/logic/impl/gen8/BattleStyles.groovy
```diff
@@ -384,6 +378,7 @@ public enum BattleStyles implements LogicCardInfo {
           text "Once during your turn, you may draw cards until you have 3 cards in your hand. If this Pokémon is in the Active Spot, you may draw cards until you have 4 cards in your hand instead. You can't use more than 1 Exciting Stage Ability each turn."
           actionA {
             checkLastTurn()
+            assert bg.em().retrieveObject("$self $thisAbility") != bg.turnCount : "Only 1 $thisAbility can be used per turn."
```
ufodynasty, line 381: I don't think this works. Since you included the pcs in the key, wouldn't another ability from a different pcs be able to be used on the same turn?

src/tcgwars/logic/impl/gen8/BattleStyles.groovy
```diff
@@ -710,7 +680,7 @@ public enum BattleStyles implements LogicCardInfo {
           text "Attach up to 3 [R] Energy cards from your discard pile to your Pokémon in any way you like."
           energyCost COLORLESS
           attackRequirement {
-            assert discardPile.filterByEnergyType(R)
+            assert discardPile.filterByEnergyType(R) : "No $R Energy in your Discard Pile."
```
ufodynasty, line 683: This is just my personal preference, but I prefer [R] here to $R. [X] seems to be used almost everywhere else.

src/tcgwars/logic/impl/gen8/BattleStyles.groovy
```diff
@@ -2913,6 +2737,7 @@ public enum BattleStyles implements LogicCardInfo {
           }
         }
         playRequirement{
+          assert my.hand || my.deck : "Both Hand and Deck are empty"
```
ufodynasty, line 2740: my.hand.getExcludedList(thisCard) should be used here.

src/tcgwars/logic/impl/gen8/BattleStyles.groovy
```diff
@@ -2904,7 +2728,7 @@ public enum BattleStyles implements LogicCardInfo {
         }
         onPlay {
           shuffleDeck(hand.getExcludedList(thisCard))
-          hand.removeAll(hand.getExcludedList(thisCard))
+          hand.getExcludedList(thisCard).discard()
```
ufodynasty, line 2731: This duplicates all cards other than bruno in the player's hand.

src/tcgwars/logic/impl/gen8/BattleStyles.groovy
flagrama: In this case, $self is the string "Krikitune V (BST 6)" which is the same for all Krikitune V right now, but that would be an issue for a reprint yeah, I'll fix that.

src/tcgwars/logic/impl/gen8/BattleStyles.groovy
flagrama: I prefer using the string representation because FIRE is less confusing than R (I totally put $F the first time)

src/tcgwars/logic/impl/gen8/BattleStyles.groovy
flagrama: No it doesn't. Discard moves cards, it does not make new ones.

flagrama: Oh it's supposed to move them to deck? I'll double check this one. Using add/remove/removeAll on a card list for a proper effect should not be done though.

ufodynasty: moveTo followed by shuffleDeck is what I would use here.

flagrama: Agreed.

ufodynasty (approved):

PR #1228, opened 2021-04-12
Implement Dusknoir Lv.X

flagrama:
Doesn't really handle interactions based on rulings currently. Does generally work though. `after KNOCKOUT` may actually be the correct place to handle the Stadium implementation change? Needs testing though.


ufodynasty (approved):

luongthomas (approved):

flagrama:
I had wanted axpendix to look at this one.

src/tcgwars/logic/impl/gen4/Stormfront.groovy
```diff
@@ -2926,38 +2926,53 @@ public enum Stormfront implements LogicCardInfo {
...
+                  stadiumCard.player = thisCard.player
+                  bg.em().run(new ChangeImplementation(stadiumCard, pkmnCard))
+                  if (bg.stadiumInfoStruct) {
+                    discardStadium()
+                  }
+
+                  PlayStadium.StadiumInfoStruct stadiumInfoStruct = new PlayStadium.StadiumInfoStruct()
+                  stadiumInfoStruct.stadiumCard = stadiumCard
+                  bg.setStadiumInfoStruct(stadiumInfoStruct)
+                  bg.em().activateEffect(stadiumCard)
```
axpendix, line 2971: I am afraid there are some additional logic in `PlayStadium` such as remove from play registration which won't work with this setup. Why don't you run it directly?

flagrama: Does this not handle the removeFromPlay stuff?
https://github.com/flagrama/tcgone-engine-contrib/blob/062226ec23c5073af344c2f1a4800ad49fa593fa/src/tcgwars/logic/impl/gen4/Stormfront.groovy#L2946-L2954

axpendix: You are quite right, That ought to handle that. Apologies

axpendix (commented):
I would still try to reuse PlayStadium with this but no big deal. It works if it works :)

PR #1232, opened 2021-04-17
Add C to Energy cards that provide energy of Every type

ufodynasty:
Q. Spinda has a Rainbow Energy attached, which counts as every type of energy, and uses its "Synchro Removal" attack. Can it discard a Call Energy from the defending Luxray [GL]?
A. Yes, Call Energy may be discarded by this attack, because the types of energy attached to Spinda includes colorless, the same type as that of Call Energy. (Sep 3, 2009 PUI Rules Team)

### DIFF
--- a/src/tcgwars/logic/impl/gen4/SecretWonders.groovy
+++ b/src/tcgwars/logic/impl/gen4/SecretWonders.groovy
@@ -1019,7 +1019,7 @@ public enum SecretWonders implements LogicCardInfo {
                       def pkmnCard = pcs.topPokemonCard
                       def tar = pcs.owner.pbg.all.findAll{it != self}.select("Choose a pokemon to attach $self to",pcs.owner)
                       def energyCard
-                      energyCard = specialEnergy(new CustomCardInfo(ELECTRODE_26).setCardTypes(ENERGY, SPECIAL_ENERGY), [[R, D, F, G, W, Y, L, M, P, C]]) {
+                      energyCard = specialEnergy(new CustomCardInfo(ELECTRODE_26).setCardTypes(ENERGY, SPECIAL_ENERGY), [valuesBasicEnergy()]) {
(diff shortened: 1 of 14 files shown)


src/tcgwars/logic/impl/gen4/SecretWonders.groovy
```diff
@@ -1019,7 +1019,7 @@ public enum SecretWonders implements LogicCardInfo {
                       def pkmnCard = pcs.topPokemonCard
                       def tar = pcs.owner.pbg.all.findAll{it != self}.select("Choose a pokemon to attach $self to",pcs.owner)
                       def energyCard
-                      energyCard = specialEnergy(new CustomCardInfo(ELECTRODE_26).setCardTypes(ENERGY, SPECIAL_ENERGY), [[R, D, F, G, W, Y, L, M, P]]) {
+                      energyCard = specialEnergy(new CustomCardInfo(ELECTRODE_26).setCardTypes(ENERGY, SPECIAL_ENERGY), [[R, D, F, G, W, Y, L, M, P, C]]) {
```
flagrama, line 1022: I'm curious about this one. Most of these energies explicitly state "This card provides C Energy" and have the little colorless icon in the top right. Does Electrode provide C here too because of "every type of energy" or does it not because it does not explicitly add C?

flagrama: This question also applies to older Rainbow energies with the icon from Team Rocket instead of a Colorless Symbol (pre-HG/SS)

flagrama:
While we're here, there are methods in the Type class that should really be used for these instead.

src/tcgwars/logic/impl/gen4/SecretWonders.groovy
flagrama: Clarified on Discord

PR #1233, opened 2021-04-19
Fall through case for broken enum

ufodynasty:
This is an attempt to allow discord user Andreloupgarou to fix their account.


flagrama:
I assume you've tested that this actually fixes the issue. Does it duplicate the card in any way though? Having a duplicate might be a bit annoying, especially since we should probably find a way to make it so that over time all decks end up using the right one and the database could be checked to see if any are left still before being able to remove the incorrect one.

ufodynasty:
This does fix the bug, but also adds a duplicate to the deck editor. It does not allow any illegal decks. I was able to recreate the bug Andreloupgarou encountered by creating a deck with the duplicate card, then removing the duplicate from the file. Adding the duplicate back allowed the deck editor to be accessed again. Unfortunately, even after deleting the decks with the duplicated card, reverting the file caused the original error. This seems like something only @axpendix can fix properly, but this has been an issue for this user for over a month, and I think that the partial fix is justified until a permanent solution can be found.

axpendix:
thanks for this. previously I had only fixed only a couple of decks with ash's pikachu. just now I redid the fix it for all ash's pikachus. therefore this won't be needed anymore.

PR #1234, opened 2021-04-19
Supreme victors

ufodynasty:
Implements the Supreme Victors expansion


src/tcgwars/logic/impl/gen4/Platinum.groovy
```diff
@@ -1141,7 +1141,7 @@ public enum Platinum implements LogicCardInfo {
             energyCost G, C, C
             onAttack {
               damage 60
-              if(defending.hasPokePower || defending.hasPokeBody) {
```
flagrama, line 1144: This one is complicated, but anything that affects PokéPower AND PokéBody should affect Pokémon Powers. (In Japanese, Pokémon Powers were actually errata'd to either a PokéBody or a PokéPower, so we may actually need to figure out what we want to do regarding that at some points since it's kind of important for Unlimited.)

src/tcgwars/logic/impl/gen4/GreatEncounters.groovy
```diff
@@ -2960,34 +2954,34 @@ public enum GreatEncounters implements LogicCardInfo {
...
+                        }]
+                        prevent()
                       }
+                      after CLEAR_SPECIAL_CONDITION, pcs, {
+                        if (ef.types.contains(ASLEEP)) {
+                          unregister()
+                        }
+                      }
+                      after FALL_BACK, pcs, { unregister() }
+                      after KNOCKOUT, pcs, { unregister() }
```
flagrama, line 2980: Effects unregister on (D)Evolve and LevelUp as well.

src/tcgwars/logic/impl/gen4/GreatEncounters.groovy
```diff
@@ -2960,34 +2954,34 @@ public enum GreatEncounters implements LogicCardInfo {
...
+                      before ASLEEP_SPC, null, null, BEGIN_TURN, {
+                        flip "Asleep (Endless Darkness)", 2, {}, {}, [2: {
+                          ef.unregisterItself(bg.em());
+                        }, 1:{
+                          bc "$ef.target is still asleep."
+                        }, 0:{
+                          bc "$pcs is knocked out by $thisMove."
+                          new Knockout(pcs).run(bg)
+                        }]
+                        prevent()
```
flagrama, line 2972: This seems like it could have some weird timing issue at some point and could end up applying when it shouldn't. Should definitely try weird combinations for testing.

src/tcgwars/logic/impl/gen4/SupremeVictors.groovy
```diff
@@ -229,30 +245,32 @@ public enum SupremeVictors implements LogicCardInfo {
...
                 delayed {
                   before BETWEEN_TURNS, {
-                    if (bg.currentTurn == self.owner.opposite && all.contains(pcs)) {
-                      bc "Doom News' effect occurs."
-                      new Knockout(pcs).run(bg)
+                    if(turnCount + 1 <= bg.turnCount){
+                      if(all.contains(pcs)){
+                        bc "Doom News kicks in"
+                        new Knockout(pcs).run(bg)
+                      }
```
flagrama, line 273: You should not be doing the whole attack in the attackRequirement

src/tcgwars/logic/impl/gen4/SupremeVictors.groovy
```diff
@@ -262,345 +280,488 @@ public enum SupremeVictors implements LogicCardInfo {
...
-            text "Choose 1 of your opponent’s Benched Pokémon. This attack does 40 damage to that Pokémon. Apply Weakness and Resistance for this attack."
+            text "Choose 1 of your opponent's Benched Pokémon. This attack does 40 damage to that Pokémon. Apply Weakness and Resistance."
             energyCost P, C, C
-            attackRequirement {}
+            attackRequirement {
+              assert opp.bench : "Your opponent has no Benched Pokémon"
+            }
             onAttack {
-              damage 0
+              def pcs = opp.all.select()
```
flagrama, line 333: Should be bench, not all

src/tcgwars/logic/impl/gen4/SupremeVictors.groovy
```diff
@@ -262,345 +280,488 @@ public enum SupremeVictors implements LogicCardInfo {
...
+              new ResolvedDamage(hp(40), self, pcs, ATTACK, DamageManager.DamageFlag.FORCE_WEAKNESS_RESISTANCE).run(bg)
+              bg.dm().applyWeakness()
+              bg.dm().applyResistance()
+              def damage = bg.dm().getTotalDamage(self, pcs)
+              bg.dm().clearDamages()
+              bg.em().run(new DirectDamage(damage, pcs))
```
flagrama, line 339: Why is this so much more complicated than when Galvantula STS applies Weak/Resist to the bench? It looks like it may be partially based on Elekid N1? However, that is only because Elekid's is not an attack.

src/tcgwars/logic/impl/gen4/SupremeVictors.groovy
```diff
@@ -262,345 +280,488 @@ public enum SupremeVictors implements LogicCardInfo {
...
-              damage 0
+              def energyCards = my.hand.select(min:0,max:2,text,cardTypeFilter(ENERGY))
+              draw 2 * energyCards.size()
+              energyCards.discard()
             }
           }
           move "Electric Current", {
-            text "40 damage. to 1 of your Benched Pokémon."
-            energyCost L, C, C, L
+            text "40 damage. Move a [L] Energy card attached to Electivire FB to 1 of your Benched Pokémon."
```
flagrama, line 361: Might want to get a ruling from someone whether you are supposed to select on declaration, move after damage, like with discarding. I'm pretty sure you should.

src/tcgwars/logic/impl/gen4/SupremeVictors.groovy
```diff
@@ -262,345 +280,488 @@ public enum SupremeVictors implements LogicCardInfo {
...
+                }
+              }
             }
           }
-
         };
       case GARCHOMP_5:
-        return evolution (this, from:"Gabite", hp:HP130, type:COLORLESS, retreatCost:0) {
-          weakness C, PLUS30
+        return evolution (this, from:"Gabite", hp:HP130, type:C, retreatCost:null) {
```
flagrama, line 377: Are you sure about the retreatCost of null here?

flagrama (commented):
Taking a break here. Why did you change all the PLUS30s to '+30'? The ENUM value should generally be used, these are basically "magic numbers" now.

ufodynasty:
> 
> 
> Taking a break here. Why did you change all the PLUS30s to '+30'? The Enum value should generally be used, these are basically "magic numbers" now.

I regenerated the template because the old one didn't handle SP symbols or energy symbols in attack text properly. I will go through and fix all of these, but perhaps a closer look should be taken at the template generator? It seems to have had a lot of issue as of recently.

src/tcgwars/logic/impl/gen4/SupremeVictors.groovy
ufodynasty: This was taken directly from galvantula UNM 62. The STS code does look much better than this.

src/tcgwars/logic/impl/gen4/SupremeVictors.groovy
ufodynasty: This does work and is used in many of the new groovy implementations, but I agree that a cost of 0 is more readable. This is another artifact of regenerating the template.

flagrama:
You're probably right that it needs looked at. Did you fix the issues with Lv. X Pokémon already? Would probably be good to PR that if you did.

src/tcgwars/logic/impl/gen4/SupremeVictors.groovy
```diff
@@ -229,378 +245,519 @@ public enum SupremeVictors implements LogicCardInfo {
...
+              damage 60 + 10 * self.cards.energyCount(F)
             }
           }
-
         };
       case VENUSAUR_13:
-        return evolution (this, from:"Ivysaur", hp:HP140, type:GRASS, retreatCost:4) {
-          weakness R
+        return evolution (this, from:"Ivysaur", hp:HP140, type:G, retreatCost:4) {
+          weakness R, '+40'
```
flagrama, line 646: Still some strings missed.

src/tcgwars/logic/impl/gen4/SupremeVictors.groovy
```diff
@@ -229,378 +245,519 @@ public enum SupremeVictors implements LogicCardInfo {
...
-            text "20+ damage. Energy attached to Yanmega to your hand. If you do, this attack does 20 damage plus 20 more damage for each Energy card you returned."
-            energyCost C, C, G
+            text "20+ damage. You may return all [G] Energy attached to Yanmega to your hand. If you do, this attack does 20 damage plus 20 more damage for each Energy card you returned."
+            energyCost C, C
             attackRequirement {}
             onAttack {
-              damage 0
+              damage 20
+              if(confirm("Return all [G] Energy attached to $self in order to do 20 additional damage?")) {
+                damage 20
```
flagrama, line 712: This doesn't return the cards, nor multiply by the number of cards returned to the hand.

src/tcgwars/logic/impl/gen4/SupremeVictors.groovy
```diff
@@ -629,132 +795,159 @@ public enum SupremeVictors implements LogicCardInfo {
...
+            text "Choose 1 of your opponent's Pokémon. Count the amount of Energy attached to that Pokémon. Put that many damage counters on the Pokémon."
             energyCost P
-            attackRequirement {}
+            attackRequirement {
+              assert opp.all.find{it.cards.energyCount(C)} : "None of your opponent's Pokémon have any Energy attached"
+            }
             onAttack {
-              damage 0
+              def tar = opp.all.findAll{it.numberOfDamageCounters()}.select("Choose 1 of your oppopnent's Pokémon")
+              directDamage 10*tar.numberOfDamageCounters(), tar
```
flagrama, line 914: This should be checking and multiplying by energyCount, not numberOfDamageCounters

src/tcgwars/logic/impl/gen4/SupremeVictors.groovy
```diff
@@ -629,132 +795,159 @@ public enum SupremeVictors implements LogicCardInfo {
...
+        return evolution (this, from:"Baltoy", hp:HP090, type:P, retreatCost:1) {
           weakness P, PLUS20
           move "Antique Magic", {
-            text "in any way you like."
+            text "Put 3 damage counters on any Pokémon (both yours and your opponent's) in any way you like."
             energyCost C, C
             attackRequirement {}
             onAttack {
-              damage 0
+              putDamageCountersOnOpponentsPokemon(3,all)
```
flagrama, line 926: > any Pokémon (both yours and your opponent's)

src/tcgwars/logic/impl/gen4/SupremeVictors.groovy
```diff
@@ -763,113 +956,143 @@ public enum SupremeVictors implements LogicCardInfo {
...
-        return evolution (this, from:"Seel", hp:HP100, type:WATER, retreatCost:3) {
+        return evolution (this, from:"Seel", hp:HP100, type:W, retreatCost:3) {
           weakness M, PLUS20
           pokeBody "Thick Fat", {
-            text "Any damage done to Dewgong by attacks from Pokémon and Pokémon is reduced by 30 ."
+            text "Any damage done to Dewgong by attacks from [R] Pokémon and [W] Pokémon is reduced by 30 (after applying Weakness and Resistance)."
             delayedA {
+              before APPLY_ATTACK_DAMAGES, {
+                bg.dm().each {
+                  if (self.active && it.to == self && it.from.owner == self.owner.opposite && (it.from.types.contains(R) || it.from.types.contains(W)) && it.dmg.value && it.notNoEffect) {
```
flagrama, line 973: It doesn't say "from your Opponent's Pokémon" nor that it needs to be active.

src/tcgwars/logic/impl/gen4/SupremeVictors.groovy
```diff
@@ -763,113 +956,143 @@ public enum SupremeVictors implements LogicCardInfo {
...
-            text "30 damage. Pokémon, this attack’s base damage is 80 instead of 30."
-            energyCost W, C, F
+            text "30 damage. If the Defending Pokémon is a [F] Pokémon, this attack's base damage is 80 instead of 30."
+            energyCost W, C
             attackRequirement {}
             onAttack {
-              damage 0
+              damage 30
+              if(defending.types.contains(F)) {
+                damage 30
```
flagrama, line 988: Should change from 30 damage to 80 damage, not 60 damage.
```suggestion
                damage 50
```

flagrama (commented):
Reviewed up to DODRIO_25

src/tcgwars/logic/impl/gen4/SupremeVictors.groovy
ufodynasty: The method is poorly named imo. It accepts any PCSList as its target space, and in this case all allows you to choose both your Pokémon and your opponents Pokémon.

src/tcgwars/logic/impl/gen4/SupremeVictors.groovy
```diff
@@ -229,378 +245,519 @@ public enum SupremeVictors implements LogicCardInfo {
...
+                  if (true){
+                    target = bg.em().retrieveObject("Gravitation_target")
+                    source = bg.em().retrieveObject("Gravitation_source")
+                    if(!target.contains(pcs)){
+                      h.object -= hp(20)
+                      target.add(pcs)
+                      bg.em().storeObject("Gravitation_target", target)
+                      source.add(self)
+                      bg.em().storeObject("Gravitation_source", source)
+                    } else if(source.get(target.indexOf(pcs)) == self){
```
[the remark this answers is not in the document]
ufodynasty: This is the most important part of this implementation. source.get(target.indexOf(pcs)) determines which metagross is the source of the effect on the pcs, and thus only lets the source metagross apply the -20 hp.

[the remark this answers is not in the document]
flagrama: But is that necessary if you don't clear the lists every time?

src/tcgwars/logic/impl/gen4/SupremeVictors.groovy
flagrama: Neat. wasn't aware there was an all() static.

src/tcgwars/logic/impl/gen4/SupremeVictors.groovy
```diff
@@ -708,8 +706,11 @@ public enum SupremeVictors implements LogicCardInfo {
             attackRequirement {}
             onAttack {
               damage 20
-              if(confirm("Return all [G] Energy attached to $self in order to do 20 additional damage?")) {
-                damage 20
+              if(confirm("Return all [G] Energy attached to $self in order to do aditional damgae?")) {
+                damage 20 * self.cards.filterByEnergyType(G)
```
flagrama, line 710: Do the `filterBy` methods return a size by default? Or will this cause an incompatible types exception?

src/tcgwars/logic/impl/gen4/SupremeVictors.groovy
```diff
@@ -911,7 +912,7 @@ public enum SupremeVictors implements LogicCardInfo {
             }
             onAttack {
               def tar = opp.all.findAll{it.numberOfDamageCounters()}.select("Choose 1 of your oppopnent's Pokémon")
```
flagrama, line 914: The findAll is still wrong.

src/tcgwars/logic/impl/gen4/SupremeVictors.groovy
```diff
@@ -229,378 +245,518 @@ public enum SupremeVictors implements LogicCardInfo {
...
           move "Flame Ball", {
-            text "60 damage. Energy card attached to Magmortar to 1 of your Benched Pokémon."
-            energyCost R, C, C, R
+            text "60 damage. You may move a [R] Energy card attached to Magmortar to 1 of your Benched Pokémon."
+            energyCost R, C, C
             attackRequirement {}
             onAttack {
-              damage 0
+              damage 60
+              if(self.cards.energyCount(R) && my.bench && confirm("MOve a [R] Energy card attached to $self to 1 of yoru Benched Pokémon")) {
```
luongthomas, line 429: Misspelt "your" here as well as "Move"

src/tcgwars/logic/impl/gen4/SupremeVictors.groovy
```diff
@@ -229,378 +245,518 @@ public enum SupremeVictors implements LogicCardInfo {
...
+                def energyCard = self.cards.select("Choose a [L] Energy to move",energyFilter(L)).fist()
```
luongthomas, line 430: This should be changed to `R` type

luongthomas, line 430: use `.first()` rather than `.fist()`

src/tcgwars/logic/impl/gen4/SupremeVictors.groovy
```diff
@@ -229,378 +245,518 @@ public enum SupremeVictors implements LogicCardInfo {
...
           }
           move "Rainbow Lariat", {
-            text "20× damage. you have in play."
+            text "20x damage. Does 20 damage times the number of different types of Pokémon SP you have in play."
             energyCost C, C, C, C
             attackRequirement {}
             onAttack {
-              damage 0
+              for(Type t1:Type.values()){
+                if(my.bench.find{it.pokemonSP && it.types.contains(t1)})
```
luongthomas, line 540: This attack mentions different types of Pokemon SP, maybe create an array each time damage is done, and for each subsequent iteration, check if it's in the array before doing damage

src/tcgwars/logic/impl/gen4/SupremeVictors.groovy
```diff
@@ -229,378 +245,518 @@ public enum SupremeVictors implements LogicCardInfo {
...
           move "Overrun", {
-            text "40 damage. Does 20 damage to 1 of your opponent’s Benched Pokémon."
+            text "40 damage. Does 20 damage to 1 of your opponent's Benched Pokémon. (Don't apply Weakness and Resistance for Benched Pokémon.)"
             energyCost R, C, C
             attackRequirement {}
             onAttack {
-              damage 0
+              damage 40
+              if(opp.bench) {
+                damage 20, opp.bench.selet("Does 20 damage to 1 of your opponent's Benched Pokémon")
```
luongthomas, line 742: `selet` to `select`

src/tcgwars/logic/impl/gen4/SupremeVictors.groovy
flagrama: I don't understand this one. It's already iterating over all the types and if a pokemonSP with that type is on the bench it adds damage which seems right to me. It should also check the active spot too since Regigigas is also a pokemonSP though.

luongthomas: Ah that makes sense.  The way it is currently coded is correct then, except for `my.bench` which I agree should be `my.all`

flagrama (commented):
Reviewed through ARCANINE_50. Skipped reviewing Spinda. I'm not feeling well enough to follow a few for loops currently so will rely on that one just being tested for correctness.

src/tcgwars/logic/impl/gen4/SupremeVictors.groovy
```diff
@@ -1030,646 +1301,847 @@ public enum SupremeVictors implements LogicCardInfo {
...
+                  flag = true
+                }
+              }
+              after APPLY_ATTACK_DAMAGES, {
+                if(flag && self.slatedToKO) {
+                  self.damage = self.fullHP - hp(10)
+                  bc "$thisAbility saved $self!"
+                }
+                flag = false
+              }
```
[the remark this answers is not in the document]
ufodynasty: This did work in testing.

src/tcgwars/logic/impl/gen4/SupremeVictors.groovy
```diff
@@ -1030,646 +1301,847 @@ public enum SupremeVictors implements LogicCardInfo {
...
+                  shuffleDeck()
+                }
+              }
             }
           }
-
         };
       case SPINDA_46:
-        return basic (this, hp:HP070, type:COLORLESS, retreatCost:1) {
+        return basic (this, hp:HP070, type:C, retreatCost:1) {
```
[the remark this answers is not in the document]
ufodynasty: This was the reason for adding colorless energy to all of the rainbow energy cards. The logic is that it iterates through all types and if the type is shared among energy cards on both spinda and the defending pokemon, it adds every energy that has that type to the list of all energy cards that could be chosen to be discarded.

src/tcgwars/logic/impl/gen4/SupremeVictors.groovy
```diff
@@ -1030,646 +1301,847 @@ public enum SupremeVictors implements LogicCardInfo {
...
+                        }]
+                        prevent()
+                      }
+                      after CLEAR_SPECIAL_CONDITION, pcs, {
+                        if (ef.types.contains(ASLEEP)) {
+                          unregister()
+                        }
+                      }
+                      after FALL_BACK, pcs, { unregister() }
+                      after KNOCKOUT, pcs, { unregister() }
```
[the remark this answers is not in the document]
ufodynasty: Would that work? I'm not familiar with how delegation works, but it seems like passing the effect to the static to be unregistered would be more complicated than than the other places where delegate is used.

src/tcgwars/logic/impl/gen4/GreatEncounters.groovy
ufodynasty: I've done a lot of testing on this card and haven't been able to trigger it when it was not appropriate.

src/tcgwars/logic/impl/gen4/SupremeVictors.groovy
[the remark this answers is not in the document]
flagrama: I'm aware, I figure it was tested enough to not need to review it, but was mentioning that I was explicitly skipping it just in case it could be relevant in the future :P

src/tcgwars/logic/impl/gen4/SupremeVictors.groovy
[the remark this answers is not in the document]
flagrama: For delayed effects, the delegate is the `DelayedEffectGroovyImpl` itself. So everything you write in the implementation files inside of the (newest branch) `delayed` closure should be available to whatever you pass the delegate to in TcgStatics. You'd just call `delegate.unregister()` with the hard part being figuring out how to add `after`s and `before`s to something like that. However,  that isn't even hard, it just takes familiarity with Groovy. Let's use Escape Board as an example.

https://github.com/axpendix/tcgone-engine-contrib/blob/c5b758dcd607e6879fcf849f5f17f1ca7afe0d36/src/tcgwars/logic/impl/gen7/UltraPrism.groovy#L2952-L2970

If we change `eff2` to

```groovy
eff2 = delayed{
  delegate.before ASLEEP_SPC, self, null, RETREAT, {if(e.parentEvent.effect.retreater == self) prevent() }
  delegate.before PARALYZED_SPC, self, null, RETREAT, {if(e.parentEvent.effect.retreater == self) prevent() }
}
```

It will still work. `before` and `after` are just methods on the `DelayedEffectGroovyImpl` class and Groovy will parse this code just the same way as not including the `delegate.` parts for them. You can cast the delegate to get the IDE to recognize it and highlight it too.

image

so basically all you have to do in the static is something like
```groovy
static unregisterCommon(def delegate, def target) {
  delegate.after FALL_BACK, target, { delegate.unregister() }
}
```

but for all the relevant ones.

src/tcgwars/logic/impl/gen4/SupremeVictors.groovy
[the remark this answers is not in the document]
ufodynasty: Fair enough. Sorry I misunderstood your intent.

src/tcgwars/logic/impl/gen4/SupremeVictors.groovy
```diff
@@ -2697,107 +3308,130 @@ public enum SupremeVictors implements LogicCardInfo {
...
-            attackRequirement {}
+            attackRequirement {
+              assert !defending.topPokemonCard.moves.isEmpty() : "$defending has no attacks"
+            }
             onAttack {
-              damage 0
+              def move=choose(defending.topPokemonCard.moves, defending.topPokemonCard.moves.collect({it.name}), "Encore") as Move
+              def pcs = defending
+              bc "$pcs can only use ${move.name} next turn."
+              delayed{
```
flagrama, line 3320: I wonder if a getter here would be more user-friendly, but may take some time to work out, so I don't expect it in this PR.

src/tcgwars/logic/impl/gen4/SupremeVictors.groovy
```diff
@@ -2871,464 +3507,648 @@ public enum SupremeVictors implements LogicCardInfo {
...
           }
           move "Icy Wind", {
             text "20 damage. Flip a coin. If heads, the Defending Pokémon is now Asleep."
             energyCost W, C
             attackRequirement {}
             onAttack {
-              damage 0
+              damage 20
+              flip {
+                apply ASLEEP
```
flagrama, line 3547: Should be afterDamage, or use the `flipThenApplySC` static

src/tcgwars/logic/impl/gen4/SupremeVictors.groovy
```diff
@@ -2871,464 +3507,648 @@ public enum SupremeVictors implements LogicCardInfo {
...
           weakness P, PLUS10
           move "Poison Sting", {
             text "10 damage. Flip a coin. If heads, the Defending Pokémon is now Poisoned."
             energyCost P, C
             attackRequirement {}
             onAttack {
               damage 10
-              flip { apply POISONED }
+              flip {
+                apply POISONED
```
flagrama, line 3598: afterDamage or flipThen static

src/tcgwars/logic/impl/gen4/SupremeVictors.groovy
```diff
@@ -2871,464 +3507,648 @@ public enum SupremeVictors implements LogicCardInfo {
...
+          def eff
           onPlay {
+            eff = delayed {
+              boolean flag = false
+              before LEVEL_UP, {
+                flag = (ef.levelUpCard as Card).player.pbg.hand.contains(ef.levelUpCard)
+              }
+              after LEVEL_UP, { if(flag) {
+                bc "Battle Tower activates"
+                heal(30, ef.pokemonToLevelUp, TRAINER_CARD)
```
flagrama, line 3715:
```suggestion
                heal(40, ef.pokemonToLevelUp, TRAINER_CARD)
```

src/tcgwars/logic/impl/gen4/SupremeVictors.groovy
```diff
@@ -2871,464 +3507,648 @@ public enum SupremeVictors implements LogicCardInfo {
...
           }
         };
       case NIGHT_TELEPORTER_138:
-        return basicTrainer (this) {
+        return itemCard (this) {
           text "Flip a coin. If heads, put all cards in your hand on top of your deck. Then, search your deck for any 1 card and put it into your hand. Shuffle your deck afterward."
           onPlay {
+            flip {
+              my.hand.getExcludedList(thisCard).moveTo(hidden:true,my.deck)
+              my.deck.select("Search your deck for a card").moveTo(hidden:true,my.hand)
```
flagrama, line 3774: Could you link the ruling here? (and no, not because of the memes)
https://compendium.pokegym.net/compendium-lvx.html#420

src/tcgwars/logic/impl/gen4/SupremeVictors.groovy
```diff
@@ -2871,464 +3507,648 @@ public enum SupremeVictors implements LogicCardInfo {
...
-            actionA {
+            text "Once during your turn , when you put Absol LV.X from your hand onto your Active Absol , you may flip 3 coins. For each heads, put the top card from your opponent’s deck in the Lost Zone."
+            onActivate {r->
+              if (r==PLAY_FROM_HAND && opp.deck && confirm('Use Darkness Send?')) {
+                powerUsed()
+                def count
+                flip 3, {
+                  count ++
+                }
+                opp.deck.subList(0,count).moveTo(opp.lostZone)
```
flagrama, line 3808: I don't think it's necessary to keep a count, you can just move each card as the heads happens since this is an ability.

src/tcgwars/logic/impl/gen4/SupremeVictors.groovy
```diff
@@ -2871,464 +3507,648 @@ public enum SupremeVictors implements LogicCardInfo {
...
             energyCost R, R, C, C, C
             attackRequirement {}
             onAttack {
-              damage 0
+              damage 150
+              flip {
+                afterDamage {
+                  discardAllSelfEnergy(C)
+                }
+              }
```
flagrama, line 3898: Should be on Tails, not Heads.

flagrama (changes requested):
Everything except for the Lv X have been reviewed. A couple of the Lv. X have been reviewed, but I noticed that you didn't double check that the text was correct since these ones weren't able to be regenerated with the carddb tools script. You should probably manually update the text for those and fix the effects that are following the broken text.

src/tcgwars/logic/impl/gen4/SupremeVictors.groovy
```diff
@@ -1030,646 +1301,853 @@ public enum SupremeVictors implements LogicCardInfo {
...
-            text "Once during your turn, when you put Chatot from your hand onto your Bench, you may look at top 4 cards of your opponent’s deck. Put them back on top of your opponent’s deck in any order."
-            actionA {
+            text "Once during your turn, when you put Chatot G from your hand onto your Bench, you may look at the top 4 cards of your opponent's deck. Put them back on top of your opponent's deck in any order."
+            onActivate {r->
+              if (r==PLAY_FROM_HAND && confirm('Use Disrupting Spy?')) {
+                powerUsed()
+                def maxSize = Math.min(opp.deck.size(),4)
+                def list = rearrange(opp.deck.subList(0,maxSize), "Rearrange top $maxSize cards of your opponent's deck")
+                my.deck.setSubList(0, list)
+                bc "$thisAbility rearranged the top cards of ${self.owner.opposite.getPlayerUsername(bg)}'s deck"
```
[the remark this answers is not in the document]
ufodynasty: powerUsed should be used on every poke power as that is how use_ability hooks in. It is critical for powerspray/power cancel to work.

src/tcgwars/logic/impl/gen4/SupremeVictors.groovy
```diff
@@ -1030,646 +1301,853 @@ public enum SupremeVictors implements LogicCardInfo {
...
           move "Ice Beam", {
             text "20 damage. Flip a coin. If heads, the Defending Pokémon is now Paralyzed."
             energyCost C, C
             attackRequirement {}
             onAttack {
               damage 20
-              flip { apply PARALYZED }
+              flip {
+                applyAfterDamage PARALYZED
+              }
```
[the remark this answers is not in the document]
ufodynasty: I'll keep that in mind. Thanks.

src/tcgwars/logic/impl/gen4/SupremeVictors.groovy
```diff
@@ -1030,646 +1301,853 @@ public enum SupremeVictors implements LogicCardInfo {
...
           move "Fireworks", {
-            text "30 damage. Energy attached to Magmar."
-            energyCost R, C, R
+            text "30 damage. Flip a coin. If tails, discard a [R] Energy attached to Magmar."
+            energyCost R, C
             attackRequirement {}
             onAttack {
-              damage 0
+              damage 30
+              flip 1, {}, {
```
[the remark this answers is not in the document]
ufodynasty: I agree that the empty closures aren't aesthetically pleasing, but personally think that having no effect on heads and an effect on tails is more clear than using flip for heads and flipTails for tails.

src/tcgwars/logic/impl/gen4/SupremeVictors.groovy
```diff
@@ -1678,407 +2156,500 @@ public enum SupremeVictors implements LogicCardInfo {
...
-          move "", {
-            text "Search damage. "
-            energyCost L
-            attackRequirement {}
-            onAttack {
-              damage 0
-            }
+          move "Call for Family", {
+            text "Search your deck for up to 2 Lightning Basic Pokémon and put them onto your Bench. Shuffle your deck afterward."
+            energyCost ()
```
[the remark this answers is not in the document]
ufodynasty: I think it would work without an energy cost, however for the sake of clarity I think it makes sense to leave it in.

src/tcgwars/logic/impl/gen4/SupremeVictors.groovy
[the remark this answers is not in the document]
flagrama: In that case, maybe we can make the current powerUsed an overload that takes a custom string for the bc, and a new one that uses the existing string?

src/tcgwars/logic/impl/gen4/SupremeVictors.groovy
```diff
@@ -1678,407 +2156,500 @@ public enum SupremeVictors implements LogicCardInfo {
...
-            attackRequirement {}
-            onAttack {
-              damage 0
+            attackRequirement {
+              assert opp.all.find{it.cards.filterByType(POKEMON_TOOL)} : "Your opponent has no Pokémon Tools attached"
+            }
+            onAttack {
+              def src = opp.all.findAll{it.cards.filterByType(POKEMON_TOOL)}.select("Pokémon that has a Tool to move")
+              def tool = src.cards.filterByType(POKEMON_TOOL).select("Pokémon Tool to move").first()
+              def tar = opp.all.findAll{it.cards.filterByType(POKEMON_TOOL).empty}.select("Move $tool to which Pokémon")
```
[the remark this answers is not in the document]
ufodynasty: I interpreted (excluding Pokémon that already has a Pokémon Tool attached to it) to mean that this attack should ignore those abilities. However I suppose I should get a ruling.

src/tcgwars/logic/impl/gen4/SupremeVictors.groovy
```diff
@@ -2087,262 +2658,279 @@ public enum SupremeVictors implements LogicCardInfo {
...
           move "Pit Search", {
             text "Search your deck for a Stadium card, show it to your opponent, and put it into your hand. Shuffle your deck afterward."
             energyCost C
-            attackRequirement {}
+            attackRequirement {
+              assert my.deck : "Your deck is empty"
+            }
             onAttack {
-              damage 0
+              my.seck.search(test,cardTypeFilter(STADIUM)).showToOpponent("Pit Search: Selected card").moveTo(my.hand)
```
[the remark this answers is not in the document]
ufodynasty: I fat fingered this card to death. I'm not committing this as I also misspelled text as test a few characters over.

src/tcgwars/logic/impl/gen4/SupremeVictors.groovy
```diff
@@ -1678,407 +2156,500 @@ public enum SupremeVictors implements LogicCardInfo {
...
             onAttack {
-              damage 0
+              damage 20
+              delayed {
+                after PLAY_BASIC_POKEMON, {
+                  if(bg.currentTurn == self.owner.opposite){
+                    ef.place.damage += 20
+                  }
+                }
+                unregisterAfter 2
```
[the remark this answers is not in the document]
ufodynasty: This doesn't catch baby evolution for babies that evolve via pokebodys but does hit wizards era babys. I'm leaving this as a note that I need to change add a method to PlayBasicPokemon.java to determine whether the card was played as an evolution or not.

src/tcgwars/logic/impl/gen4/SupremeVictors.groovy
ufodynasty: I like that idea a lot.

src/tcgwars/logic/impl/gen4/SupremeVictors.groovy
ufodynasty: I did this to avoid chat log spam. Is there any reason to change this beyond aesthetic?

flagrama: That makes sense. No real reason, no.

axpendix:
Hi, I've fixed merge conflicts, is there anything missing on this one before merging?

src/tcgwars/logic/impl/gen4/SupremeVictors.groovy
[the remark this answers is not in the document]
axpendix: each closure whether empty (i.e. `{}`) or not takes up memory and disk space as each closure is compiled as a java class and they are all loaded to memory in runtime. that's the reason the server takes too much time to boot, it's loading hundreds of thousands of classes into memory. the solution is rather radical, rather than putting set implementation files in the same folder with java sources, they could be split into groovy script files for each card then loaded to a GroovyShell created specifically for each game, that would evaluate the card scripts per game, with a new classloader per shell, then throw it away at the end of game.

PR #1235, opened 2021-04-19
TcgStatics: Pass effect to closures

flagrama:


ufodynasty:
https://github.com/axpendix/tcgone-engine-contrib/pull/1028#discussion_r544677937

flagrama:
Weird, could have sworn it was `null` at that point.

luongthomas (approved):
Looks good

src/tcgwars/logic/groovy/TcgStatics.groovy
```diff
@@ -1913,11 +1913,11 @@ class TcgStatics {
           applyEffect = bg.currentTurn == self.owner.opposite && self.active && bg.dm().find({ it.to == self && it.dmg.value })
         }
         after APPLY_ATTACK_DAMAGES, {
-          if (applyEffect) {
+          if (applyEffect && ef.attacker.inPlay) {
```
axpendix, line 1916: could you explain why this is needed?

flagrama: It's been a while, but IIRC it has to do with attacks where the attacker is no longer in-play (Knocked Out due to self-damage or a Knockout effect running as part of the attack) since the closure with the effect will run after ApplyDamages does its CheckFaint.

PR #1239, opened 2021-04-24
Implement Pop Series 6

luongthomas:
Initial run through of the Pop Series 6 set

### DIFF
--- a/src/tcgwars/logic/impl/gen4/PopSeries6.groovy
+++ b/src/tcgwars/logic/impl/gen4/PopSeries6.groovy
@@ -151,9 +151,7 @@ public enum PopSeries6 implements LogicCardInfo {
               damage 10
               if (opp.bench) {
                 multiSelect(opp.bench, 2, text).each {
-                  targeted(it) {
-                    damage 10, it
-                  }
+                  damage 10, it
                 }
               }
             }
--- a/src/tcgwars/logic/impl/gen4/PopSeries6.groovy
+++ b/src/tcgwars/logic/impl/gen4/PopSeries6.groovy
@@ -103,7 +103,6 @@ public enum PopSeries6 implements LogicCardInfo {
               damage 0
             }
           }
-
         };
       case LUCARIO_2:
         return evolution (this, from:"Riolu", hp:HP090, type:FIGHTING, retreatCost:1) {
@@ -125,7 +124,6 @@ public enum PopSeries6 implements LogicCardInfo {
               damage 0
             }
           }
-
         };
       case MANAPHY_3:
         return basic (this, hp:HP070, type:WATER, retreatCost:1) {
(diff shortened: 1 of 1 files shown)


src/tcgwars/logic/impl/gen4/PopSeries6.groovy
```diff
@@ -87,358 +92,89 @@ public enum PopSeries6 implements LogicCardInfo {
...
-            text "Oran Berry. Remove 1 damage counter from Gible at the end of your turn."
-            energyCost ()
-            attackRequirement {}
-            onAttack {
-              damage 0
+          customAbility {
+            // Oran Berry: Remove 1 damage counter from Gible at the end of your turn.
+            delayedA {
+              before BEGIN_TURN,{
+                if (self.numberOfDamageCounters >= 1) {
```
ufodynasty, line 113: Oran berry should only activate at the end of its owner's turn

src/tcgwars/logic/impl/gen4/PopSeries6.groovy
```diff
@@ -87,358 +92,89 @@ public enum PopSeries6 implements LogicCardInfo {
...
-            text "Oran Berry. Remove 1 damage counter from Pikachu at the end of your turn."
-            energyCost ()
-            attackRequirement {}
-            onAttack {
-              damage 0
+          customAbility {
+            // Oran Berry: Remove 1 damage counter from Gible at the end of your turn.
+            delayedA {
+              before BEGIN_TURN,{
+                if (self.numberOfDamageCounters >= 1) {
```
ufodynasty, line 139: Oran berry should only heal at the end of its owner's turn.

src/tcgwars/logic/impl/gen4/PopSeries6.groovy
```diff
@@ -87,358 +92,89 @@ public enum PopSeries6 implements LogicCardInfo {
...
           move "Spark", {
             text "10 damage. Does 10 damage to 2 of your opponent’s Benched Pokémon."
             energyCost L, C
             attackRequirement {}
             onAttack {
-              damage 0
+              damage 10
+              if (opp.bench) {
+                multiSelect(opp.bench, 2, text).each {
+                  targeted(it) {
```
ufodynasty, line 154: targeted is not appropriate here. It will cause abilities that prevent all effects of attacks except damage to block the damage.

ufodynasty (commented):
Thank you so much for doing these.

src/tcgwars/logic/impl/gen4/PopSeries6.groovy
flagrama: BETWEEN_TURNS doesn't change anything. It's still happening at the end of both player's turns.

luongthomas:
Addressed review comments

flagrama (approved):

PR #1242, opened 2021-04-24
Implement Pop Series 9

luongthomas:
Initial run through of the set

### DIFF
--- a/src/tcgwars/logic/impl/gen4/PopSeries9.groovy
+++ b/src/tcgwars/logic/impl/gen4/PopSeries9.groovy
@@ -463,7 +463,7 @@ public enum PopSeries9 implements LogicCardInfo {
             onAttack {
               damage 40
 
-              if (my.all.findAll {it.name == "Chimchar" }) {
+              if (my.all.findAll {it.name == "Buizel" }) {
                 opp.bench.each {
                   damage 10, it
                 }
--- a/src/tcgwars/logic/impl/gen4/PopSeries9.groovy
+++ b/src/tcgwars/logic/impl/gen4/PopSeries9.groovy
@@ -247,7 +247,9 @@ public enum PopSeries9 implements LogicCardInfo {
           move "Knock Off", {
             text "Flip a coin. If heads, choose 1 card from your opponent’s hand without looking and discard it."
             energyCost C
-            attackRequirement {}
+            attackRequirement {
+              assert opp.hand : "Opponent's hand is empty"
+            }
             onAttack {
               flip {
                 discardRandomCardFromOpponentsHand()
--- a/src/tcgwars/logic/impl/gen4/PopSeries9.groovy
+++ b/src/tcgwars/logic/impl/gen4/PopSeries9.groovy
@@ -106,14 +106,10 @@ public enum PopSeries9 implements LogicCardInfo {
             onAttack {
               damage 100
 
-              if (confirm(text)) {
-                opp.bench.each {
-                  damage 10, it
-                }
-                afterDamage {
-                  discardSelfEnergy C,C
-                }
+              opp.bench.each {
+                damage 10, it
               }
+              discardSelfEnergyAfterDamage(C, C)
             }
           }
         };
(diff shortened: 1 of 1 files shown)


src/tcgwars/logic/impl/gen4/PopSeries9.groovy
```diff
@@ -425,18 +454,26 @@ public enum PopSeries9 implements LogicCardInfo {
...
           }
           move "Wavelet", {
             text "40 damage. If you have Buizel in play, this attack does 10 damage to each of your opponent’s Benched Pokémon."
             energyCost W, C, C
             attackRequirement {}
             onAttack {
-              damage 0
+              damage 40
+
+              if (my.all.findAll {it.name == "Chimchar" }) {
```
ufodynasty, line 470: This should check for Buizel instead of Chimchar.

ufodynasty (commented):
Thank you so much for doing these!

src/tcgwars/logic/impl/gen4/PopSeries9.groovy
```diff
@@ -94,39 +96,50 @@ public enum PopSeries9 implements LogicCardInfo {
...
             onAttack {
-              damage 0
+              damage 100
+
+              if (confirm(text)) {
+                opp.bench.each {
+                  damage 10, it
+                }
+                afterDamage {
+                  discardSelfEnergy C,C
```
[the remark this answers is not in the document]
flagrama: Specifically, use it outside of the afterDamage for proper order, it uses afterDamage properly within itself.

src/tcgwars/logic/impl/gen4/PopSeries9.groovy
```diff
@@ -199,12 +215,14 @@ public enum PopSeries9 implements LogicCardInfo {
           move "Poltergeist", {
             text "30+ damage. Look at your opponent’s hand. This attack does 30 damage plus 10 more damage for each Trainer, Supporter, and Stadium card in your opponent’s hand."
             energyCost P, C
-            attackRequirement {}
+            attackRequirement {
+              assert opp.hand : "Opponent's hand is empty"
+            }
             onAttack {
-              damage 0
+              def filteredHand = opp.hand.shuffledCopy().showToMe("Opponent's hand").filterByType(ITEM, SUPPORTER, STADIUM)
```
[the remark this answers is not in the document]
flagrama: Hidden selections are shuffled, though since those are the only ones that it matters for, yeah shuffling the copy you see should no longer be necessary.

src/tcgwars/logic/impl/gen4/PopSeries9.groovy
```diff
@@ -214,18 +232,20 @@ public enum PopSeries9 implements LogicCardInfo {
...
           }
           move "Super Fast", {
             text "30 damage. If you have Pachirisu in play, flip a coin. If heads, prevent all effects of an attack, including damage, done to Buizel during your opponent’s next turn."
             energyCost W, W
             attackRequirement {}
             onAttack {
-              damage 0
+              damage 30
+              if (my.all.findAll {it.name == "Pachirisu" }) {
+                flip { preventAllDamageNextTurn() }
```
flagrama, line 245:
```suggestion
                flip { preventAllEffectsNextTurn() }
```

src/tcgwars/logic/impl/gen4/PopSeries9.groovy
```diff
@@ -235,18 +255,23 @@ public enum PopSeries9 implements LogicCardInfo {
...
           }
           move "Nimble", {
             text "30 damage. If you have Turtwig in play, remove from Croagunk the number of damage counters equal to the damage you did to the Defending Pokémon."
             energyCost P, P
             attackRequirement {}
             onAttack {
-              damage 0
+              damage 30
+
+              if (my.all.findAll {it.name == "Turtwig" }) {
```
flagrama, line 270: You only need `find` or `any` here. These will both stop at finding one instead of collecting them all 

`find` will return the found value, `any` will return a boolean

luongthomas:
Addressed all the comments above

flagrama (approved):

PR #1244, opened 2021-04-28
Gen4fixes

ufodynasty:
Addresses bugs mentioned on the trello page

### DIFF
--- a/src/tcgwars/logic/impl/gen4/GreatEncounters.groovy
+++ b/src/tcgwars/logic/impl/gen4/GreatEncounters.groovy
@@ -2643,7 +2643,7 @@ public enum GreatEncounters implements LogicCardInfo {
                   holder.object.addAll(it.topPokemonCard.moves)
                 }
               }
-              self.owner.opposite.pbg.bench.findAll { it.name.contains("Unown") }.each {
+              self.owner.opposite.pbg.all.findAll { it.name.contains("Unown") }.each {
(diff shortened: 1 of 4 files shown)


src/tcgwars/logic/impl/gen4/GreatEncounters.groovy
```diff
@@ -2643,7 +2643,7 @@ public enum GreatEncounters implements LogicCardInfo {
                   holder.object.addAll(it.topPokemonCard.moves)
                 }
               }
-              self.owner.opposite.pbg.bench.findAll { it.name.contains("Unown") }.each {
```
flagrama, line 2646: I think it'd be easier to use the metronomeA static here.

src/tcgwars/logic/impl/gen4/SecretWonders.groovy
```diff
@@ -315,23 +317,23 @@ public enum SecretWonders implements LogicCardInfo {
...
                 }
               }, {
                 if(self.cards.filterByType(ENERGY).size()>=4){
                   damage 120
-                  afterDamage {
-                    self.cards.select(count:4,"Discard 4 Energy cards attached to $self",cardTypeFilter(ENERGY))
-                  }
+                  energyList = self.cards.select(count:4,"Discard 4 Energy cards attached to $self",cardTypeFilter(ENERGY))
                 } else {
                   bc "Blast Burn failed"
```
flagrama, line 331: Should probably have this else in the other if statement too.

PR #1245, opened 2021-04-28
Dark houndoom should be able to discard special energy.

ufodynasty:
Also fixed timing.

### DIFF
--- a/src/tcgwars/logic/impl/gen3/TeamRocketReturns.groovy
+++ b/src/tcgwars/logic/impl/gen3/TeamRocketReturns.groovy
@@ -336,11 +336,16 @@ public enum TeamRocketReturns implements LogicCardInfo {
             energyCost R, C, C
             onAttack {
               damage 40
-              if(confirm("discard a [R] Energy or [D] Energy attached to Dark Houndoom for 20 more damage?")){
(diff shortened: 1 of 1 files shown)


flagrama:
Is it better to have two prompts like this, or would it be better to make the select have a min of 0 and remove the confirm?

src/tcgwars/logic/impl/gen3/TeamRocketReturns.groovy
```diff
@@ -337,8 +337,11 @@ public enum TeamRocketReturns implements LogicCardInfo {
             onAttack {
               damage 40
               if(confirm("discard a [R] Energy or [D] Energy attached to Dark Houndoom for 20 more damage?")){
-                self.cards.filterByType(BASIC_ENERGY).findAll{it.asEnergyCard().containsTypePlain(D) || it.asEnergyCard().containsTypePlain(R)}.select().discard()
+                def card = self.cards.filterByType(ENERGY).findAll{it.asEnergyCard().containsTypePlain(D) || it.asEnergyCard().containsTypePlain(R)}.select()
                 damage 20
+                afterDamage {
+                  card.discard()
+                }
               }
```
axpendix, line 345: should better be written like

```suggestion
def list = self.cards.filterByType(ENERGY).findAll{it.containsType(D) || it.containsType(R)}
if(list && list.select(min:0, "Discard a [R] or [D] Energy attached to $self for 20 more damage?").discard()) {
  damage 20
}
```

PR #1246, opened 2021-04-30
Stormfront + sources

ufodynasty:
Fixes all bugs found in stormfront on trello.


flagrama:
Why didn't you just use
`
import static tcgwars.logic.effect.Source.*`?

ufodynasty:
intellij did auto import source and I was getting issues with POKEPOWER and POKEBODY defaulting to effect types instead of sources.

PR #1252, opened 2021-05-27
Card Fixes

starg09:
* Luxray (Legends Awakened)'s Poké-Body "Rivalry" should increase damage by 50, not 30.
* Politoed (Legends Awakened)'s Poké-Body "Enthusiasm" should only work when a Poliwag, a Poliwhirl and a Poliwrath are in play (not counting pre-evolution of evolved Pokémon in this). The damage increase should also not be ignored by shred attacks, should they happen.
* Rayquaza (EX Deoxys) should now stop causing a NPE when using it's "Dragon Aura" Poké-Body.
* EXP.ALL (EX FireRed & LeafGreen) should in theory prevent activation when a benched Pokémon is KO'd (couldn't reproduce, but logic could be improved anyway)
* Sidney's Stadium (EX Power Keepers) should now properly remove pre-existing Special Conditions.

### DIFF
--- a/src/tcgwars/logic/impl/gen4/LegendsAwakened.groovy
+++ b/src/tcgwars/logic/impl/gen4/LegendsAwakened.groovy
@@ -584,16 +584,20 @@ public enum LegendsAwakened implements LogicCardInfo {
             text "If you have Poliwag, Poliwhirl, and Poliwrath in play, each of these Pokémon's attacks does 60 more damage to the Defending Pokémon (before applying Weakness and Resistance)."
             delayedA {
               after PROCESS_ATTACK_EFFECTS, {
-                if (
-                  ef.attacker.owner == self.owner &&
-                  ef.attacker.name in ["Poliwag", "Poliwhirl", "Poliwrath"] &&
-                  ["Poliwag", "Poliwhirl", "Poliwrath"].every{
-                    toadName -> self.owner.pbg.all.any{ it.name == toadName }
-                  }
-                ) bg.dm().each {
-                  if (it.to.active && it.notZero) {
-                    bc "Enthusiasm +60"
-                    it.dmg += hp(60)
+                if (ef.attacker.owner == self.owner) bg.dm().each {
+                  if (it.to.active && it.to.owner != self.owner && it.notZero) {
+                    def attacker = it.from
+                    def enthusiasm_cond = {
+                      def toadNames = ["Poliwag", "Poliwhirl", "Poliwrath"]
+                      attacker.name in toadNames &&
+                        toadNames.every{toadName ->
+                          self.owner.pbg.all.any{ pcs -> pcs.name == toadName }
+                        }
+                    }
+                    if(enthusiasm_cond) {
+                      bc "Enthusiasm +60"
+                      it.dmg += hp(60)
+                    }
                   }
                 }
               }
--- a/src/tcgwars/logic/impl/gen4/LegendsAwakened.groovy
+++ b/src/tcgwars/logic/impl/gen4/LegendsAwakened.groovy
@@ -465,7 +465,7 @@ public enum LegendsAwakened implements LogicCardInfo {
             text "If your opponent has any Pokémon LV.X in play, each of Luxray's attacks does 50 more damage to the Active Pokémon (before applying Weakness and Resistance)."
             delayedA {
               after PROCESS_ATTACK_EFFECTS, {
-                if (ef.attacker == self && opp.all.any { it.isPokemonLevelUp() }) {
+                if (ef.attacker == self && opp.all.any { it.pokemonLevelUp }) {
                   bg.dm().each {
                     if (it.to.active && it.notZero) {
                       bc "Rivalry +50"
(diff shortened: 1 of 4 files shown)


starg09:
[POLITOED_12](https://trello.com/c/roxOK3iP/1117-politoed12)

starg09:
[LUXRAY_8](https://trello.com/c/YlFUWsp8/1113-luxray8)

starg09:
[RAYQUAZA_22](https://trello.com/c/R1HLCzGH/1103-rayquaza22)

starg09:
[EXP_ALL_91](https://trello.com/c/iqlnjgNS/1104-expall91)

starg09:
[SIDNEY_S_STADIUM_82](https://trello.com/c/7K685ZHa/1105-sidneysstadium82)

src/tcgwars/logic/impl/gen4/LegendsAwakened.groovy
```diff
@@ -465,11 +465,11 @@ public enum LegendsAwakened implements LogicCardInfo {
             text "If your opponent has any Pokémon LV.X in play, each of Luxray's attacks does 50 more damage to the Active Pokémon (before applying Weakness and Resistance)."
             delayedA {
               after PROCESS_ATTACK_EFFECTS, {
-                if (opp.all.any { it.isPokemonLevelUp() }) {
+                if (ef.attacker == self && opp.all.any { it.isPokemonLevelUp() }) {
```
flagrama, line 468: While we are here, might as well use Groovy convention here.
```suggestion
                if (ef.attacker == self && opp.all.any { it.pokemonLevelUp }) {
```

flagrama (approved):
Nothing major, looks pretty good.

src/tcgwars/logic/impl/gen4/LegendsAwakened.groovy
```diff
@@ -584,8 +584,14 @@ public enum LegendsAwakened implements LogicCardInfo {
...
             delayedA {
               after PROCESS_ATTACK_EFFECTS, {
-                bg.dm().each {
-                  if (it.from.name in ["Poliwag", "Poliwhirl", "Poliwrath"] && it.to.active && it.to.owner != self.owner && it.dmg.value && it.notNoEffect) {
+                if (
+                  ef.attacker.owner == self.owner &&
+                  ef.attacker.name in ["Poliwag", "Poliwhirl", "Poliwrath"] &&
+                  ["Poliwag", "Poliwhirl", "Poliwrath"].every{
+                    toadName -> self.owner.pbg.all.any{ it.name == toadName }
+                  }
```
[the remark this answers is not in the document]
starg09: Def an ugly card text. Having the check only when it's an attack into the defending sounds a lot less expensive, will edit in a bit 👍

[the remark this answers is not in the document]
starg09: Reworked in https://github.com/axpendix/tcgone-engine-contrib/pull/1252/commits/5ec1c96fda8da2a40d36816da3e1618bc4f8569f

Maybe a bit too verbose with the `def attacker`, but shouldn't affect much 😅

src/tcgwars/logic/impl/gen4/LegendsAwakened.groovy
```diff
@@ -584,10 +584,20 @@ public enum LegendsAwakened implements LogicCardInfo {
             text "If you have Poliwag, Poliwhirl, and Poliwrath in play, each of these Pokémon's attacks does 60 more damage to the Defending Pokémon (before applying Weakness and Resistance)."
             delayedA {
               after PROCESS_ATTACK_EFFECTS, {
-                bg.dm().each {
-                  if (it.from.name in ["Poliwag", "Poliwhirl", "Poliwrath"] && it.to.active && it.to.owner != self.owner && it.dmg.value && it.notNoEffect) {
-                    bc "Enthusiasm +60"
-                    it.dmg += hp(60)
+                if (ef.attacker.owner == self.owner) bg.dm().each {
```
flagrama, line 587: This looks like a broken if statement

starg09: ?
Seems to compile fine, it was originally a one-line-if (just with a horrible multi-line condition in the middle)

There is a broken check inside though, probably due to not using `return` inside enthusiasm_cond, will send it in a minute.

flagrama: I mean, it might compile, but it still _looks_ wrong. Maybe you could put the `bg.dm().each {` on the next line to make it easier to read. As it is, it's definitely easy to read past the fact it is doing anything with the damage manager if you are used to `if` at beginning of line, `{` at end of line = "everything in between is conditions"

starg09: That's fair, will move the each into a separate line. Usually move it, even if not adding {} brackets for the single-line if.

PR #1254, opened 2021-05-27
Fix Platinum Blissey_22

luongthomas:
Fix crash on Nurse call ability

### DIFF
--- a/src/tcgwars/logic/impl/gen4/Platinum.groovy
+++ b/src/tcgwars/logic/impl/gen4/Platinum.groovy
@@ -1009,11 +1009,11 @@ public enum Platinum implements LogicCardInfo {
               checkLastTurn()
               checkNoSPC()
               assert my.hand : "Your hand is empty"
-              assert my.all.find{it.numberOfDamageCounters} : "Your Pokémon are healthy"
+              assert my.all.find {it.numberOfDamageCounters} : "Your Pokémon are healthy"
               assert bg.em().retrieveObject("Nurse_Call") != bg.turnCount : "You cannot use Nurse Call more than once per turn"
               powerUsed()
               my.hand.select("Choose a card to discard").discard()
-              heal 20, my.all.findAll(it.numberOfDamageCounters).select("Heal which Pokémon"), Source.POKEPOWER
+              heal 20, my.all.findAll{it.numberOfDamageCounters}.select("Heal which Pokémon"), Source.POKEPOWER
             }
           }
           move "Return", {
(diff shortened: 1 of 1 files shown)


luongthomas:
[BLISSEY_22](https://trello.com/c/Vcy6BlhT/992-blissey22)

src/tcgwars/logic/impl/gen4/Platinum.groovy
```diff
@@ -1009,11 +1009,11 @@ public enum Platinum implements LogicCardInfo {
               checkLastTurn()
               checkNoSPC()
               assert my.hand : "Your hand is empty"
-              assert my.all.find{it.numberOfDamageCounters} : "Your Pokémon are healthy"
+              assert my.all.find {it.numberOfDamageCounters} : "Your Pokémon are healthy"
```
starg09, line 1012: > Q. Can I use Blissey's "Nurse Call" Poke-POWER to discard a card from my hand even if I have no damaged Pokemon in play?
>
> A. Yes, you can. (Jun 11, 2009 PUI Rules Team)

This shouldn't be asserted at all (heal below could be replaced with a filer, then doing the heal only if valid targets are found.

starg09, line 1012: > Q. Can I use Blissey's "Nurse Call" Poke-POWER to discard a card from my hand even if I have no damaged Pokemon in play?
>
> A. Yes, you can. (Jun 11, 2009 PUI Rules Team)

This shouldn't be asserted at all (heal below could be replaced with a filer, then doing the heal only if valid targets are found.

starg09 (changes requested):

src/tcgwars/logic/impl/gen4/Platinum.groovy
starg09: Not going to lie, I've no clue what I meant to say when I wrote "could be replaced with a filer" 😓 

But with no assert, there's a need for one more change, suggesting it in a second.

src/tcgwars/logic/impl/gen4/Platinum.groovy
```diff
@@ -1009,11 +1009,10 @@ public enum Platinum implements LogicCardInfo {
               checkLastTurn()
               checkNoSPC()
               assert my.hand : "Your hand is empty"
-              assert my.all.find{it.numberOfDamageCounters} : "Your Pokémon are healthy"
               assert bg.em().retrieveObject("Nurse_Call") != bg.turnCount : "You cannot use Nurse Call more than once per turn"
               powerUsed()
               my.hand.select("Choose a card to discard").discard()
-              heal 20, my.all.findAll(it.numberOfDamageCounters).select("Heal which Pokémon"), Source.POKEPOWER
+              heal 20, my.all.findAll{it.numberOfDamageCounters}.select("Heal which Pokémon"), Source.POKEPOWER
```
starg09, line 1015:
```suggestion
              if (my.all.any{it.numberOfDamageCounters}) 
                heal 20, my.all.findAll{it.numberOfDamageCounters}.select("Heal which Pokémon"), Source.POKEPOWER
```

starg09: Just to avoid a popup when that part can't be done.

starg09 (approved):

PR #1257, opened 2021-05-28
Groovy re-implementation for "Ruby & Sapphire - Base Set" (Cards 1-22)

starg09:
* Relocated some cards already implemented in Groovy, and redirected to the RubySapphireNG print. Affects:
  - Blaziken (RS 3, PK 5)
  - Delcatty (RS 5, PK 8)
  - Gardevoir (RS 7, PK 9)
  - Slaking (RS 12, PK 13)
  - Sceptile (RS 20, POP4 5)
* Besides that, this PR implements cards 1 through 22.

### DIFF
--- a/src/tcgwars/logic/impl/gen3/RubySapphireNG.groovy
+++ b/src/tcgwars/logic/impl/gen3/RubySapphireNG.groovy
@@ -407,17 +407,16 @@ public enum RubySapphireNG implements LogicCardInfo {
         move "Super Slap Push", {
           text "Does 20 damage to each Defending Pokémon."
           energyCost F
-          attackRequirement {}
           onAttack {
-
+            damage 20
           }
         }
         move "Mega Throw", {
           text "40+ damage. If the Defending Pokémon is a Pokémon-ex, this attack does 40 damage plus 40 more damage."
           energyCost F, C, C
-          attackRequirement {}
           onAttack {
             damage 40
+            if (defending.EX) damage 40
           }
         }
       };
--- a/src/tcgwars/logic/impl/gen3/RubySapphireNG.groovy
+++ b/src/tcgwars/logic/impl/gen3/RubySapphireNG.groovy
@@ -541,14 +541,21 @@ public enum RubySapphireNG implements LogicCardInfo {
         pokePower "Water Call", {
           text "Once during your turn (before your attack), you may attach a [W] Energy card from your hand to your Active Pokémon. This power can't be used if Swampert is affected by a Special Condition."
           actionA {
+            checkNoSPC()
+            checkLastTurn()
+            assert my.hand.filterByType(BASIC_ENERGY).filterByEnergyType(W) : "You have no [W] Energy cards in your hand"
+
+            powerUsed()
+            def pcs = my.all.select()
+            attachEnergyFrom(type:W, my.hand, pcs)
           }
         }
         move "Hypno Splash", {
           text "50 damage. The Defending Pokémon is now Asleep."
           energyCost W, W, C, C
-          attackRequirement {}
           onAttack {
             damage 50
+            applyAfterDamage(ASLEEP)
           }
         }
       };
--- a/src/tcgwars/logic/impl/gen3/RubySapphireNG.groovy
+++ b/src/tcgwars/logic/impl/gen3/RubySapphireNG.groovy
@@ -645,17 +645,23 @@ public enum RubySapphireNG implements LogicCardInfo {
         move "Invisible Hand", {
           text "If any of your opponent's Active Pokémon are Evolved Pokémon, search your deck for any 1 card and put it into your hand. Shuffle your deck afterward."
           energyCost C
-          attackRequirement {}
+          attackRequirement {
+            assert my.deck : "There are no more cards in your deck."
+            assert opp.active.evolution : "Your opponent's Active Pokémon isn't an Evolved Pokémon"
+          }
           onAttack {
-
+            my.deck.select().moveTo(hidden: true, my.hand)
+            shuffleDeck()
           }
         }
         move "Repulsion", {
           text "Flip a coin. If heads, your opponent returns the Defending Pokémon and all cards attached to it to his or her hand. (If your opponent doesn't have any Benched Pokémon or other Active Pokémon, this attack does nothing.)"
           energyCost C, C
-          attackRequirement {}
+          attackRequirement { assert opp.bench: "Opponent's bench is empty" }
           onAttack {
-
+            flip {
+              scoopUpPokemon(defending, delegate)
+            }
           }
         }
       };
--- a/src/tcgwars/logic/impl/gen3/RubySapphireNG.groovy
+++ b/src/tcgwars/logic/impl/gen3/RubySapphireNG.groovy
@@ -565,15 +565,14 @@ public enum RubySapphireNG implements LogicCardInfo {
         move "Take Down", {
           text "50 damage. Wailord does 20 damage to itself."
           energyCost C, C, C
-          attackRequirement {}
           onAttack {
             damage 50
+            damage 20, self
           }
         }
         move "Surf", {
           text "70 damage."
           energyCost W, W, W, C, C
-          attackRequirement {}
           onAttack {
             damage 70
           }
--- a/src/tcgwars/logic/impl/gen3/RubySapphireNG.groovy
+++ b/src/tcgwars/logic/impl/gen3/RubySapphireNG.groovy
@@ -206,15 +206,16 @@ public enum RubySapphireNG implements LogicCardInfo {
         move "Retaliate", {
           text "10x damage. Flip a coin. If heads, this attack does 10 damage times the number of damage counters on Aggron."
           energyCost C
-          attackRequirement {}
           onAttack {
             damage 10
+            flip {
+              damage 10 * self.numberOfDamageCounters
+            }
           }
         }
         move "Mega Punch", {
           text "40 damage."
           energyCost C, C, C
-          attackRequirement {}
           onAttack {
             damage 40
           }
@@ -222,9 +223,8 @@ public enum RubySapphireNG implements LogicCardInfo {
         move "Double Lariat", {
           text "70x damage. Flip 2 coins. This attack does 70 damage times the number of heads."
           energyCost M, M, C, C, C
-          attackRequirement {}
           onAttack {
-            damage 70
+            flip 2, {damage 70}
           }
         }
       };
--- a/src/tcgwars/logic/impl/gen3/RubySapphireNG.groovy
+++ b/src/tcgwars/logic/impl/gen3/RubySapphireNG.groovy
@@ -427,17 +427,17 @@ public enum RubySapphireNG implements LogicCardInfo {
         move "Attract Current", {
           text "10 damage. Search your deck for a [L] Energy card and attach it to 1 of your Pokémon. Shuffle your deck afterward."
           energyCost C
-          attackRequirement {}
           onAttack {
             damage 10
+            attachEnergyFrom(type:L, my.deck, my.all)
           }
         }
         move "Thunder Jolt", {
           text "50 damage. Flip a coin. If tails, Manectric does 10 damage to itself."
           energyCost L, L, C
-          attackRequirement {}
           onAttack {
             damage 50
+            flip 1, {}, {damage 10, self}
           }
         }
       };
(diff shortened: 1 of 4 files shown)


src/tcgwars/logic/impl/gen3/RubySapphireNG.groovy
```diff
@@ -432,15 +500,38 @@ public enum RubySapphireNG implements LogicCardInfo {
         weakness F
         pokeBody "Lazy", {
           text "As long as Slaking is your Active Pokémon, your opponent's Pokémon can't use any Poké-Powers."
-          delayedA {
+          getterA IS_ABILITY_BLOCKED, { Holder h->
+            if (self.active && h.effect.target.owner == self.owner.opposite && h.effect.ability instanceof PokePower) {
+              h.object=true
+            }
+          }
+          //TODO: Is this needed here?
```
flagrama, line 508: Currently yes. This is the getter that `checkGlobalAbility()` checks when you want to be able to block it. It may be possible to use the normal `IS_ABILITY_BLOCKED` getter instead there though, but that wouldn't work with things like `instanceof PokePower` and is something we probably need to add a way of checking to `IS_GLOBAL_ABILITY_BLOCKED` too.

src/tcgwars/logic/impl/gen3/RubySapphireNG.groovy
```diff
@@ -432,15 +500,38 @@ public enum RubySapphireNG implements LogicCardInfo {
...
-          attackRequirement {}
+          attackRequirement {
+            assert self.cards.filterByType(BASIC_ENERGY) : "$self has no Basic Energy attached"
+          }
           onAttack {
-            damage 100
+            if(self.cards.filterByType(BASIC_ENERGY)){
+              damage 100
+              afterDamage {
+                self.cards.filterByType(BASIC_ENERGY).select("Discard a basic energy from $self.").discard()
```
flagrama, line 531: Again, must select before Damage is applied.

src/tcgwars/logic/impl/gen3/RubySapphireNG.groovy
```diff
@@ -450,14 +541,21 @@ public enum RubySapphireNG implements LogicCardInfo {
         pokePower "Water Call", {
           text "Once during your turn (before your attack), you may attach a [W] Energy card from your hand to your Active Pokémon. This power can't be used if Swampert is affected by a Special Condition."
           actionA {
+            checkNoSPC()
+            checkLastTurn()
+            assert my.hand.filterByType(BASIC_ENERGY).filterByEnergyType(W) : "You have no [W] Energy cards in your hand"
```
flagrama, line 546: Theoretically this PokePower could allow a Special Energy card that provides W energy while not attached couldn't it? Should this force filtering to `BASIC_ENERGY`? `filterByEnergyType` should check the types provided to the Special Energy in its implementation which should all now be correct (empty for most Special Energy cards)

src/tcgwars/logic/impl/gen3/RubySapphireNG.groovy
```diff
@@ -487,17 +584,19 @@ public enum RubySapphireNG implements LogicCardInfo {
...
         }
         move "Flamethrower", {
           text "80 damage. Discard a [R] Energy card attached to Blaziken."
           energyCost R, C, C, C
-          attackRequirement {}
           onAttack {
             damage 80
+            afterDamage{
+              discardSelfEnergy(R)
+            }
```
flagrama, line 599:
```suggestion
            discardSelfEnergyAfterDamage R
```

src/tcgwars/logic/impl/gen3/RubySapphireNG.groovy
```diff
@@ -597,14 +700,24 @@ public enum RubySapphireNG implements LogicCardInfo {
         pokePower "Energy Trans", {
           text "As often as you like during your turn (before your attack), move a [G] Energy card attached to 1 of your Pokémon to another of your Pokémon. This power can't be used if Sceptile is affected by a Special Condition."
           actionA {
+            checkNoSPC()
+            assert my.all.findAll {it.cards.energyCount(G)>0} : "There are no Pokémon with [G] Energy cards"
+            assert my.all.size()>=2 : "There is only one Pokémon on the field"
```
flagrama, line 705: Nitpick: Couldn't you just check if there is a benched Pokémon?

src/tcgwars/logic/impl/gen3/RubySapphireNG.groovy
```diff
@@ -597,14 +700,24 @@ public enum RubySapphireNG implements LogicCardInfo {
         pokePower "Energy Trans", {
           text "As often as you like during your turn (before your attack), move a [G] Energy card attached to 1 of your Pokémon to another of your Pokémon. This power can't be used if Sceptile is affected by a Special Condition."
           actionA {
+            checkNoSPC()
+            assert my.all.findAll {it.cards.energyCount(G)>0} : "There are no Pokémon with [G] Energy cards"
```
flagrama, line 704: This should probably be either a `find`, `any`, or should store the results of the `findAll` before asserting it so that you don't have to repeat the `findAll` later.

flagrama (changes requested):
Looks pretty good, mostly some small nitpicks, things to discuss, and timing issues with selecting/discarding Energy.

src/tcgwars/logic/impl/gen3/RubySapphireNG.groovy
```diff
@@ -257,14 +267,27 @@ public enum RubySapphireNG implements LogicCardInfo {
...
           energyCost R, C, C
-          attackRequirement {}
           onAttack {
             damage 50
+            opp.bench.each {
+              damage 10, it
+            }
+            afterDamage {
+              discardSelfEnergy(R)
+            }
```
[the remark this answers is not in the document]
starg09: 👍 

This was copied 1:1 from what was in Power Keepers, but that sounds good. Seems there's still a need to mass-implement this new static in older sets?

src/tcgwars/logic/impl/gen3/RubySapphireNG.groovy
```diff
@@ -277,14 +300,20 @@ public enum RubySapphireNG implements LogicCardInfo {
...
           text "100 damage. Discard 2 basic Energy cards attached to Camerupt or this attack does nothing."
           energyCost R, R, C, C
-          attackRequirement {}
+          attackRequirement {
+            assert self.cards.filterByType(BASIC_ENERGY).size() >= 2 : "$self needs 2 or more Basic Energies attached"
+          }
           onAttack {
             damage 100
+            afterDamage {
+              self.cards.filterByType(BASIC_ENERGY).select(count: 2, "Select 2 Basic Energies to discard.").discard()
```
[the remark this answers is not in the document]
starg09: That sounds good. Will look into it later.

Would probably leave it as a more energy-specific static, potentially returning a boolean so it can be used as the main condition (thinking of whenever "using an attack that does nothing" is properly implemented):

```
if (forcedDiscardEnergy(amount, isBasic, energyTypeFilter)){
    damage 100
}
```

src/tcgwars/logic/impl/gen3/RubySapphireNG.groovy
```diff
@@ -294,14 +323,24 @@ public enum RubySapphireNG implements LogicCardInfo {
...
+            checkLastTurn()
+            assert my.hand.filterByType(ENERGY) : "No Energy in hand"
+            //Using LV.X Compendium Ruling instead of the EX: You can discard even if you don't have cards in deck.
+            powerUsed()
+
+            my.hand.filterByType(ENERGY).select("Discard").discard()
+            if (my.deck){
+              def maxDraw = Math.min(3, my.deck.size())
+              draw choose(1..maxDraw, "Draw how many cards?")
+            }
```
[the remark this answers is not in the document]
starg09: Don't remember the details (would need to check the git blame for PK, from before moving the card over here), but many of these cards need to do this check to avoid a NPE iirc.

Also, there's still the need in most EX-era cards to be able to select "up to x cards" to draw, unlike modern cards which usually force the max amount (or as much as possible).

src/tcgwars/logic/impl/gen3/RubySapphireNG.groovy
```diff
@@ -336,14 +380,24 @@ public enum RubySapphireNG implements LogicCardInfo {
...
           actionA {
+            checkLastTurn()
+            checkNoSPC()
+            assert my.deck : "Deck is empty"
+            powerUsed()
+
+            my.deck.search("Search for a [P] Energy card to attach to one of your Pokémon.", energyFilter(P)).each {
+              def tar = my.all.select("Attach $it to? That Pokémon will receive 2 damage counters.")
+              attachEnergy(tar, it)
+              directDamage 20, tar, SRC_ABILITY
```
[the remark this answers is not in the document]
starg09: Another 1:1 migrated card. Will correct, already using POKEPOWER in other sets but missed it here 👌 
```suggestion
              directDamage 20, tar, POKEPOWER
```

src/tcgwars/logic/impl/gen3/RubySapphireNG.groovy
starg09: 👍 

(another moved-over card, but thanks for responding to it anyway 😅 )

src/tcgwars/logic/impl/gen3/RubySapphireNG.groovy
starg09:
```suggestion
            assert my.hand.filterByEnergyType(W) : "You have no [W] Energy cards in your hand"
```

src/tcgwars/logic/impl/gen3/RubySapphireNG.groovy
starg09: Yup, fixing it (migrated card from POP)

```suggestion
            assert my.all.find{it.cards.energyCount(G)>0} : "There are no Pokémon with [G] Energy cards"
```

src/tcgwars/logic/impl/gen3/RubySapphireNG.groovy
starg09: That should work fine, I think? Vaguely recall discussion on that a long time ago, most likelt when building assertMyBench() and the other similar statics.

Leaving it like this for now, can be talked further later on though, in a more general sense.

src/tcgwars/logic/impl/gen3/RubySapphireNG.groovy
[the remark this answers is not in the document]
flagrama: Yes. Gen8 is the only generation this has been consistently used in so far. I've been using it for anything that gets bug fixed too though.

src/tcgwars/logic/impl/gen3/RubySapphireNG.groovy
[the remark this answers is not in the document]
flagrama: If `draw 3` causes an NPE, it's probably better to fix that in `DrawCard` instead of having to do this logic on literally every card that draws cards. However, `DrawCard` does check if a Deck is already size 0 and skips drawing from it in that case from what I see.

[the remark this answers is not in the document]
flagrama: I did miss that this was an "up to" implementation. Maybe we could add this as a static `drawUpTo` instead?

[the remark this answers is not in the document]
starg09: Sounds like a great idea. Separate method, or integrated into the draw static itself?

In any case, probably worth its own PR (will add a couple of these to trello in a bit)

src/tcgwars/logic/impl/gen3/RubySapphireNG.groovy
flagrama: Figured it was just a general copy/pasted thing, not specific to this PR or even this Card. Figured I'd share some knowledge there in case it was relevant to whoever added the TODO in the first place. It'll get to whoever wrote it in the first place eventually.

starg09: ~~Yeah, wonder who wrote that...~~

src/tcgwars/logic/impl/gen3/RubySapphireNG.groovy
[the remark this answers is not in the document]
flagrama: I'll take your word on it since AFAIK being unable to discard energy causing an attack to fail is a retro-only issue.

src/tcgwars/logic/impl/gen3/RubySapphireNG.groovy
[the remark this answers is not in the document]
flagrama: Best to be it's own method I think. Easier to read the method name `drawUpTo` instead of having to read (and create) javadoc for a method that seems simple enough currently.

flagrama (approved):
Assuming proper Energy discarding is prioritized (or else Tegull will certainly throw it in the not-working pile during testing anyway) and drawUpTo are fixed in soon-to-be-implemented PRs it seems fine now.

luongthomas (approved):

PR #1259, opened 2021-05-29
Fix "Dragon Aura" in Rayquaza (DX)

starg09:
Should only check for basic Energy cards, not any. This should fix some issues with it regarding this ability.


starg09:
[RAYQUAZA_22](https://trello.com/c/R1HLCzGH/1103-rayquaza22)

PR #1260, opened 2021-05-31
Fixed SF Regigigas Lv X 100

cdosen:
Energies should no longer attempt to be attached to a KO'ed self.


starg09 (approved):

starg09:
[REGIGIGAS_LV_X_100](https://trello.com/c/7hS2lSHn/964-regigigaslvx100)

flagrama:
Personally I'd prefer links to rulings rather than just a text comment with no reference. Unless the code is doing something that needs to be explained to grok it easier, it's best that you don't leave a comment that could leave someone asking "Okay, but why?" (although the older compendium pages do make it more difficult, in this case you'd use https://compendium.pokegym.net/compendium-lvx.html#314 )

But, this is just a nitpick.

PR #1261, opened 2021-06-01
LA 1-29 fixes

cdosen:
Fixed issues with Legends Awakened cards numbers 1-29. Also fixed issues with Castforms, since it is the same fix as Form Change for Deoxys. There will probably still be residual issues with those cards and certain effects, but they should at least let you swap them out now.

### DIFF
--- a/src/tcgwars/logic/impl/gen4/LegendsAwakened.groovy
+++ b/src/tcgwars/logic/impl/gen4/LegendsAwakened.groovy
@@ -1744,8 +1744,8 @@ public enum LegendsAwakened implements LogicCardInfo {
               if (selected) {
                 def tpc = self.topPokemonCard
                 selected.moveTo(suppressLog: true, self.cards)
-                my.deck.add(tpc)
-                self.cards.remove(tpc)
+                tpc.moveTo(suppressLog: true, my.deck)
+                bc "${tpc.name} was swapped with ${selected.name}."
                 new CheckAbilities().run(bg)
                 checkFaint()
               }
@@ -1784,8 +1784,8 @@ public enum LegendsAwakened implements LogicCardInfo {
               if (selected) {
                 def tpc = self.topPokemonCard
                 selected.moveTo(suppressLog: true, self.cards)
-                my.deck.add(tpc)
-                self.cards.remove(tpc)
+                tpc.moveTo(suppressLog: true, my.deck)
+                bc "${tpc.name} was swapped with ${selected.name}."
                 new CheckAbilities().run(bg)
                 checkFaint()
               }
@@ -1818,8 +1818,8 @@ public enum LegendsAwakened implements LogicCardInfo {
               if (selected) {
                 def tpc = self.topPokemonCard
                 selected.moveTo(suppressLog: true, self.cards)
-                my.deck.add(tpc)
-                self.cards.remove(tpc)
+                tpc.moveTo(suppressLog: true, my.deck)
+                bc "${tpc.name} was swapped with ${selected.name}."
                 new CheckAbilities().run(bg)
                 checkFaint()
               }
@@ -1853,8 +1853,8 @@ public enum LegendsAwakened implements LogicCardInfo {
               if (selected) {
                 def tpc = self.topPokemonCard
                 selected.moveTo(suppressLog: true, self.cards)
-                my.deck.add(tpc)
-                self.cards.remove(tpc)
+                tpc.moveTo(suppressLog: true, my.deck)
+                bc "${tpc.name} was swapped with ${selected.name}."
                 new CheckAbilities().run(bg)
                 checkFaint()
               }
--- a/src/tcgwars/logic/impl/gen4/LegendsAwakened.groovy
+++ b/src/tcgwars/logic/impl/gen4/LegendsAwakened.groovy
@@ -414,15 +414,13 @@ public enum LegendsAwakened implements LogicCardInfo {
             energyCost R, R, C, C
             attackRequirement {}
             onAttack {
+              damage 40
               if (confirm("Would you like to discard basic energy cards attached to Heatran?")) {
                 assert self.cards.findAll(cardTypeFilter(BASIC_ENERGY): "There are no basic energies attached to Heatran.")
                 def atchEnergy = self.cards.filterByType(BASIC_ENERGY)
                 def lostEnergy = atchEnergy.select(max: atchEnergy.size(), "Choose basic energies to discard.")
-                damage 40 + 20 * lostEnergy.size()
+                damage 20 * lostEnergy.size()
                 lostEnergy.discard()
-              }
-              else {
-                damage 40
               }
             }
           }
--- a/src/tcgwars/logic/impl/gen4/LegendsAwakened.groovy
+++ b/src/tcgwars/logic/impl/gen4/LegendsAwakened.groovy
@@ -869,12 +869,13 @@ public enum LegendsAwakened implements LogicCardInfo {
             energyCost G
             attackRequirement {}
             onAttack {
+              def pcs = defending
               if (opp.bench && confirm("Switch the Defending Pokémon with 1 of your opponent’s Benched Pokémon?")) {
-                sw2 opp.bench.select("New Defending Pokémon")
+                pcs = opp.bench.select("New Defending Pokémon")
+                sw2 pcs
               }
 
               delayed {
-                def pcs = defending
                 before KNOCKOUT, pcs, {
                   if (self.numberOfDamageCounters && confirm("Remove all damage counters from Cradily?")) {
                     healAll self
--- a/src/tcgwars/logic/impl/gen4/LegendsAwakened.groovy
+++ b/src/tcgwars/logic/impl/gen4/LegendsAwakened.groovy
@@ -1046,7 +1046,7 @@ public enum LegendsAwakened implements LogicCardInfo {
             getterA (GET_FULL_HP, self) { Holder h->
               if (self.active) {
                 def tar = self.owner.opposite.pbg.active
-                if (tar.name == "Ditto" && tar.hasPokeBody() && tar.getAbilities().keySet().first().name == "Ditto DNA") {
+                if (tar.fullName == "Ditto (LA 27)") {
                   h.object = hp(60)
                 }
                 else {
--- a/src/tcgwars/logic/impl/gen4/LegendsAwakened.groovy
+++ b/src/tcgwars/logic/impl/gen4/LegendsAwakened.groovy
@@ -802,7 +802,7 @@ public enum LegendsAwakened implements LogicCardInfo {
                 powerUsed()
                 def tar = my.prizeCardSet.faceDownCards.select(hidden: false, min: 0, "Choose a Pokemon card from your prizes.", cardTypeFilter(POKEMON))
                 if (tar) {
-                  tar.moveTo(my.hand)
+                  tar.moveTo(my.hand).showToOpponent("Your opponent has swapped this prize card with a card from their hand.")
                   my.hand.select("Card to put back into Prizes").moveTo(hidden:true, my.prizeCardSet)
                 }
               }
--- a/src/tcgwars/logic/impl/gen4/LegendsAwakened.groovy
+++ b/src/tcgwars/logic/impl/gen4/LegendsAwakened.groovy
@@ -805,7 +805,6 @@ public enum LegendsAwakened implements LogicCardInfo {
                   newPrize.moveTo(hidden: true, my.prizeCardSet)
                   tar.moveTo(my.hand)
                 }
-                rearrange(my.prizeCardSet.faceDownCards)
               }
             }
           }
--- a/src/tcgwars/logic/impl/gen4/LegendsAwakened.groovy
+++ b/src/tcgwars/logic/impl/gen4/LegendsAwakened.groovy
@@ -1093,19 +1093,15 @@ public enum LegendsAwakened implements LogicCardInfo {
           move "Drought", {
             text "Choose up to 2 basic [F] Energy cards from your hand and attach them to 1 of your Pokémon."
             attackRequirement {
-              assert my.hand.filterByEnergyType(F) : "No [F] Energy cards in your hand"
+              assert my.hand.filterByBasicEnergyType(F) : "No [F] Energy cards in your hand"
             }
             onAttack {
-              def energies = my.hand.filterByType(BASIC_ENERGY).filterByEnergyType(F).select(max: 2, "Select 2 basic [F] Energy cards to attach to 1 of your Pokémon")
+              def energies = my.hand.filterByBasicEnergyType(F).select(max: 2, "Select 2 basic [F] Energy cards to attach to 1 of your Pokémon")
               my.hand.removeAll(energies)
+              def tar = my.all.select("Attach to which Pokemon?")
               energies.each {
-                if (my.bench) {
-                  attachEnergy(my.bench.select("Attach to which?"))
-                } else {
-                  attachEnergy(self, it)
-                }
+                attachEnergy(tar, it)
               }
-              heal energies.size() * 10, self
             }
           }
           move "Major Earthquake", {
--- a/src/tcgwars/logic/impl/gen4/LegendsAwakened.groovy
+++ b/src/tcgwars/logic/impl/gen4/LegendsAwakened.groovy
@@ -798,7 +798,7 @@ public enum LegendsAwakened implements LogicCardInfo {
           pokePower "Time Walk", {
             text "Once during your turn, when you put Azelf from your hand onto your Bench, you may look at all of your face-down Prize cards. If you do, you may choose 1 Pokémon you find there, show it to your opponent, and put it into your hand. Then, choose 1 card in your hand and put it as a Prize card face down."
             onActivate {r->
-              if (r==PLAY_FROM_HAND && confirm("Use Time Walk?")){
+              if (r==PLAY_FROM_HAND && self.benched && confirm("Use Time Walk?")){
                 assert my.hand : "No cards in hand"
                 powerUsed()
                 def tar = my.prizeCardSet.faceDownCards.cardTypeFilter(POKEMON).select(hidden: false, min: 0, "Choose a Pokemon in your prizes to replace with a card in your hand.").first()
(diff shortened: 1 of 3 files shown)


src/tcgwars/logic/impl/gen4/LegendsAwakened.groovy
```diff
@@ -794,20 +801,20 @@ public enum LegendsAwakened implements LogicCardInfo {
...
-              my.prizeCardSet.set(indexOfOldPrize, newPrize)
-              my.prizeCardSet.setVisible(newPrize, true)
-              my.hand.remove(newPrize)
+            onActivate {r->
+              if (r==PLAY_FROM_HAND && confirm("Use Time Walk?")){
+                assert my.hand : "No cards in hand"
+                powerUsed()
+                def tar = my.prizeCardSet.faceDownCards.select(hidden: false, min: 0, "Choose a Pokemon in your prizes to replace with a card in your hand.").first()
+                assert tar : "No Pokemon selected."
+                assert tar.hasTypes(POKEMON) : "Only Pokemon can be grabbed by Time Walk."
```
flagrama, line 810: `assert`s are inappropriate after `powerUsed()` has been called. They are to prevent attacks and abilities from running, not for ending them prematurely.

src/tcgwars/logic/impl/gen4/LegendsAwakened.groovy
```diff
@@ -794,20 +801,20 @@ public enum LegendsAwakened implements LogicCardInfo {
...
+                def newPrize = my.hand.select(hidden: true, "Card to put into Prizes").first()
+                def indexOfOldPrize = my.prizeCardSet.indexOf(tar)
+                my.hand.add(tar)
+                my.prizeCardSet.set(indexOfOldPrize, newPrize)
+                my.prizeCardSet.setVisible(newPrize, false)
+                my.hand.remove(newPrize)
```
flagrama, line 816: Please do not directly manipulate persistent CardLists with methods from the ArrayList base class.

src/tcgwars/logic/impl/gen4/LegendsAwakened.groovy
```diff
@@ -1052,7 +1051,7 @@ public enum LegendsAwakened implements LogicCardInfo {
           pokeBody "Ditto DNA", {
             text "As long as Ditto is your Active Pokémon, its maximum HP is the same as your opponent's Active Pokémon. Ditto can use the attacks of that Pokémon as its own. (You still need the necessary Energy to use each attack.) If that Pokémon is no longer your opponent's Active Pokémon, choose 1 of your opponent's Active Pokémon for Ditto to copy."
             getterA (GET_FULL_HP, self) { Holder h->
-              if (self.active) h.object = hp(self.owner.opposite.pbg.active.getFullHP())
+              if (self.active) h.object = hp(self.owner.opposite.pbg.active.topPokemonCard.getFullHP())
```
flagrama, line 1054: `getFullHP` does not exist on the PokemonCard class. It only exists on the `PokemonCardSet` class.

src/tcgwars/logic/impl/gen4/LegendsAwakened.groovy
```diff
@@ -1094,19 +1093,15 @@ public enum LegendsAwakened implements LogicCardInfo {
           move "Drought", {
             text "Choose up to 2 basic [F] Energy cards from your hand and attach them to 1 of your Pokémon."
             attackRequirement {
-              assert my.hand.filterByEnergyType(F) : "No [F] Energy cards in your hand"
+              assert my.hand.filterByBasicEnergyType(F) : "No [F] Energy cards in your hand"
             }
             onAttack {
-              def energies = my.hand.filterByType(BASIC_ENERGY).filterByEnergyType(F).select(max: 2, "Select 2 basic [F] Energy cards to attach to 1 of your Pokémon")
+              def energies = my.hand.filterByBasicEnergyType(F).select(max: 2, "Select 2 basic [F] Energy cards to attach to 1 of your Pokémon")
               my.hand.removeAll(energies)
```
flagrama, line 1100: Please correct this while you are working on this card.

src/tcgwars/logic/impl/gen4/LegendsAwakened.groovy
```diff
@@ -1745,15 +1740,13 @@ public enum LegendsAwakened implements LogicCardInfo {
...
-              if (newCastform) {
-                newCastform.moveTo(self.cards)
-                my.deck.add(oldCastform)
-                self.cards.remove(oldCastform)
+              def selected = my.deck.search(max:1,"Select a Castform",{it.name.contains("Castform")})
+              if (selected) {
+                def tpc = self.topPokemonCard
+                selected.moveTo(suppressLog: true, self.cards)
+                my.deck.add(tpc)
+                self.cards.remove(tpc)
```
flagrama, line 1748: Same as the Deoxys cards. @starg09

flagrama (changes requested):

src/tcgwars/logic/impl/gen4/LegendsAwakened.groovy
```diff
@@ -637,8 +640,10 @@ public enum LegendsAwakened implements LogicCardInfo {
             onAttack {
               damage 60
               def maxCount = Math.min(self.cards.filterByEnergyType(M).size(), opp.bench.size())
-              multiSelect(opp.bench, maxCount, text).each {
-                damage 20, it
+              if (maxCount != 0) {
```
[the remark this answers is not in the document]
cdosen: This was actually included because the attack text doesn't include "you may." If there is at least 1 [M] on probopass and your opponent has a bench, you would have to select at least 1 benched pokemon. (No ruling on probopass explicitly, but is consistent with rulings on team galactic's wager, regigigas lv x, and looker's investigation, among others)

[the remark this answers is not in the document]
flagrama: Which is why you set the `min` for `multiSelect` to 1, but per the source code of the method it calls to actually do the work, the code if the value passed to `selectMultiPokemon`'s `max` is smaller than the `min` it ensures you select that smaller amount, and if the `max` is 0 it just skips the selection loop entirely. This `if` condition doesn't do anything that code doesn't already do.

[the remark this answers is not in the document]
cdosen: got it, in the process of fixing.

src/tcgwars/logic/impl/gen4/LegendsAwakened.groovy
```diff
@@ -805,15 +801,12 @@ public enum LegendsAwakened implements LogicCardInfo {
...
-                def indexOfOldPrize = my.prizeCardSet.indexOf(tar)
-                my.hand.add(tar)
-                my.prizeCardSet.set(indexOfOldPrize, newPrize)
-                my.prizeCardSet.setVisible(newPrize, false)
-                my.hand.remove(newPrize)
+                def tar = my.prizeCardSet.faceDownCards.cardTypeFilter(POKEMON).select(hidden: false, min: 0, "Choose a Pokemon in your prizes to replace with a card in your hand.").first()
+                if (tar) {
+                  def newPrize = my.hand.select(hidden: true, "Card to put into Prizes").first()
+                  newPrize.moveTo(hidden: true, my.prizeCardSet)
+                  tar.moveTo(my.hand)
```
flagrama, line 808: > Q. Can I rearrange the order of my prize cards after using Azelf's "Time Walk" Poke-POWER?
A. Yes, you can. (Oct 16, 2008 PUI Rules Team)

Might be more true to the game to allow a user to rearrange prize cards after performing these actions. Though it might be overkill and be an option players don't often choose to use too. Might be worth discussing in the Discord.

src/tcgwars/logic/impl/gen4/LegendsAwakened.groovy
```diff
@@ -1051,7 +1044,7 @@ public enum LegendsAwakened implements LogicCardInfo {
           pokeBody "Ditto DNA", {
             text "As long as Ditto is your Active Pokémon, its maximum HP is the same as your opponent's Active Pokémon. Ditto can use the attacks of that Pokémon as its own. (You still need the necessary Energy to use each attack.) If that Pokémon is no longer your opponent's Active Pokémon, choose 1 of your opponent's Active Pokémon for Ditto to copy."
             getterA (GET_FULL_HP, self) { Holder h->
-              if (self.active) h.object = hp(self.owner.opposite.pbg.active.topPokemonCard.getFullHP())
+              if (self.active) h.object = self.owner.opposite.pbg.active.getFullHP()
```
flagrama, line 1047: I think this may infinitely recurse if two DITTO_27 face each other.

flagrama (changes requested):
A couple of rulings issues. 

Also, fixing the possible issue with Ditto v Ditto may require one of us who can use a debugger on it to fix properly if it is an actual issue, so if you can't think of a good and accurate way to check that, don't sweat it.

src/tcgwars/logic/impl/gen4/LegendsAwakened.groovy
```diff
@@ -249,7 +249,7 @@ public enum LegendsAwakened implements LogicCardInfo {
               bg.em().storeObject("Form_Change",bg.turnCount)
               powerUsed()
 
-              def selected = my.deck.search(max:1,"Select a Deoxys",{it.name.contains("Deoxys") && it.cardTypes.isNot(EX)}) //Deoxys cannot Form Change into Deoxys ex.
+              def selected = my.deck.search(max:1,"Select a Deoxys",{it.name.contains("Deoxys") && it.cardTypes.isNot(EX) && it.cardTypes.isNot(LEGEND) && it.cardTypes.isNot(POKEMON_EX) && it.cardTypes.isNot(POKEMON_GX) && it.cardTypes.isNot(POKEMON_V)}) //Deoxys cannot Form Change into Deoxys ex.
```
flagrama, line 252: Using `it.name == "Deoxys"` or something similar should resolve needing to continue to add to this every time a new card type is released.

src/tcgwars/logic/impl/gen8/BattleStyles.groovy
```diff
@@ -2964,11 +2964,13 @@ public enum BattleStyles implements LogicCardInfo {
         onPlay {reason->
           eff = delayed {
             def disable = { PokemonToolCard card, pcs ->
-              if (card.name == "Tool Jammer" || !pcs.active || card.player == self?.owner)
+              if (card.name == "Tool Jammer" || card.player == self?.owner)
```
flagrama, line 2967: I see github is confused again.

src/tcgwars/logic/impl/gen4/LegendsAwakened.groovy
cdosen: The problem is the name isn't deoxys, but deoxys attack forme, etc. That was why the card was bugged in the first place. I can probably simplify it by refusing rulebox pokemon instead of the giant list though. That should prevent said issue.

flagrama: It won't. Rule-boxes aren't what you think they are.

flagrama: `it.name == "Deoxys" || (it.name.contains("Deoxys") && it.name.contains("Forme")`

src/tcgwars/logic/impl/gen4/LegendsAwakened.groovy
```diff
@@ -794,20 +797,18 @@ public enum LegendsAwakened implements LogicCardInfo {
...
-              def tar = my.prizeCardSet.faceDownCards.select(hidden: false, "Choose a Prize card to replace with one in your hand.").first()
-              my.hand.add(tar)
-
-              def indexOfOldPrize = my.prizeCardSet.indexOf(tar)
-              my.prizeCardSet.set(indexOfOldPrize, newPrize)
-              my.prizeCardSet.setVisible(newPrize, true)
-              my.hand.remove(newPrize)
+            onActivate {r->
+              if (r==PLAY_FROM_HAND && confirm("Use Time Walk?")){
+                assert my.hand : "No cards in hand"
```
flagrama, line 802: > Q 自分のサイドがすべてオモテのとき、アグノムのポケパワー｢タイムウォーク｣を使うことはできますか?
A いいえ、できません。

Translation:
> Q Can I use Azelf's PokéPower "Time Walk" when all my Prize Cards are face-up?
A No, you cannot.

src/tcgwars/logic/impl/gen4/LegendsAwakened.groovy
```diff
@@ -794,20 +797,18 @@ public enum LegendsAwakened implements LogicCardInfo {
...
+                powerUsed()
+                def tar = my.prizeCardSet.faceDownCards.cardTypeFilter(POKEMON).select(hidden: false, min: 0, "Choose a Pokemon in your prizes to replace with a card in your hand.").first()
+                if (tar) {
+                  def newPrize = my.hand.select(hidden: true, "Card to put into Prizes").first()
+                  newPrize.moveTo(hidden: true, my.prizeCardSet)
+                  tar.moveTo(my.hand)
```
flagrama, line 808: An explicit `showToOpponent("string")` is probably a good idea.

src/tcgwars/logic/impl/gen4/LegendsAwakened.groovy
```diff
@@ -794,20 +797,18 @@ public enum LegendsAwakened implements LogicCardInfo {
...
-
-              def indexOfOldPrize = my.prizeCardSet.indexOf(tar)
-              my.prizeCardSet.set(indexOfOldPrize, newPrize)
-              my.prizeCardSet.setVisible(newPrize, true)
-              my.hand.remove(newPrize)
+            onActivate {r->
+              if (r==PLAY_FROM_HAND && confirm("Use Time Walk?")){
+                assert my.hand : "No cards in hand"
+                powerUsed()
+                def tar = my.prizeCardSet.faceDownCards.cardTypeFilter(POKEMON).select(hidden: false, min: 0, "Choose a Pokemon in your prizes to replace with a card in your hand.").first()
```
flagrama, line 804: `cardTypeFilter(POKEMON)` on the `faceDownCards` means the player won't see their other prize cards, only the ones that are Pokémon. `select` has an argument that lets you pass a filter for what you can select from a CardList instead.

See: DUSKNOIR_DP33 in DiamondPearlPromos or WEEDLE_86 in RisingRivals

src/tcgwars/logic/impl/gen4/LegendsAwakened.groovy
flagrama, line 802:
```suggestion
```
> Q 自分の手札が0枚のとき、アグノムのポケパワー｢タイムウォーク｣を使って、ウラになっているサイドからポケモンをえらぶことはできますか?
A はい、できます。
ただし、サイドからえらんだポケモン以外に手札がないため、えらんだポケモンをもう一度サイドとしてもどして、ポケパワーを終わります。

Translation:
> Q When I have no cards in my hand, can I use Azelf's PokéPower "Time Walk" to select a Pokémon from my facedown Prize Cards?
A Yes, you can. However, since you don't have any other Pokémon [rather, cards] in your hand other than the one you selected from the Prize Cards, you have to return the selected Pokémon to your Prize Cards again to finish the PokéPower.

src/tcgwars/logic/impl/gen4/LegendsAwakened.groovy
```diff
@@ -794,20 +797,18 @@ public enum LegendsAwakened implements LogicCardInfo {
...
-
-              def tar = my.prizeCardSet.faceDownCards.select(hidden: false, "Choose a Prize card to replace with one in your hand.").first()
-              my.hand.add(tar)
-
-              def indexOfOldPrize = my.prizeCardSet.indexOf(tar)
-              my.prizeCardSet.set(indexOfOldPrize, newPrize)
-              my.prizeCardSet.setVisible(newPrize, true)
-              my.hand.remove(newPrize)
+            onActivate {r->
+              if (r==PLAY_FROM_HAND && confirm("Use Time Walk?")){
```
flagrama, line 801: Not sure if this one matters in the actual Play server, but on the Dev server at least it'd be kind of annoying to place this Pokémon as the "active" as the first Pokémon and be asked this.
```suggestion
              if (r==PLAY_FROM_HAND && self.benched && confirm("Use Time Walk?")){
```

flagrama (changes requested):
A couple more ruling issues based on [the Japanese rulings](https://www.pokemon-card.com/rules/faq/search.php?freeword=%E3%82%BF%E3%82%A4%E3%83%A0%E3%82%A6%E3%82%A9%E3%83%BC%E3%82%AF+&regulation_faq_main_item1=DP)

Sorry about all the review comments spread around. This is a slightly complicated effect.

src/tcgwars/logic/impl/gen4/LegendsAwakened.groovy
flagrama: We'll table whether you are supposed to rearrange only face down cards, or all prize cards for now until we can find a proper ruling.

src/tcgwars/logic/impl/gen4/LegendsAwakened.groovy
```diff
@@ -794,20 +796,16 @@ public enum LegendsAwakened implements LogicCardInfo {
...
+              if (reason == PLAY_FROM_HAND && self.benched && confirm("Use Time Walk?")) {
+                assert my.prizeCardSet.faceDownCards.size() != 0 : "You cannot use Time Walk if all of your prizes are face-up."
+                powerUsed()
+                def tar = my.prizeCardSet.faceDownCards.select(hidden: false, min: 0, "Choose a Pokemon card from your prizes.", cardTypeFilter(POKEMON))
+                if (tar) {
+                  tar.moveTo(my.hand).showToOpponent("Your opponent has swapped this prize card with a card from their hand.")
+                  my.hand.select("Card to put back into Prizes").moveTo(hidden:true, my.prizeCardSet)
+                }
+              }
             }
```
flagrama, line 809: The rearrange seems to have gotten lost here.

flagrama: I'm going to merge this as-is though since this isn't a critical issue. You can just create a new PR that adds it. Remember to test all of these on [the Dev server](https://dev.tcgone.net), especially the Ditto interactions.

flagrama (approved):

PR #1262, opened 2021-06-01
Fix Azelf LA

Grayahm:


starg09:
[AZELF_19](https://trello.com/c/AGvWESzJ/1124-azelf19)

PR #1263, opened 2021-06-02
Fix Tool Jammer

flagrama:
Affect the opponent's current active Pokémon instead of stale values


flagrama:
[TOOL_JAMMER_136](https://trello.com/c/qxrZJd3K/1050-tooljammer136)

axpendix (approved):

PR #1264, opened 2021-06-03
Fix BST Kingdra

flagrama:
playerType is not a param value.


flagrama:
[KINGDRA_33](https://trello.com/c/XiBR6tLe/947-kingdra33)

PR #1265, opened 2021-06-03
Updateing PokemodFossil.groovy

TheAquaPiplup:
Adding in some secret rare cards and fixing bugs


flagrama (changes requested):
Will not compile. There are multiple issues with keeping scope with curly braces.

PR #1275, opened 2021-06-15
Implement "Dark Genes" for Honchkrow (MT)

starg09:
Appears to work correctly at a first testing glance.

### DIFF
--- a/src/tcgwars/logic/impl/gen4/MysteriousTreasures.groovy
+++ b/src/tcgwars/logic/impl/gen4/MysteriousTreasures.groovy
@@ -31,8 +31,9 @@ import tcgwars.logic.effect.basic.*;
 import tcgwars.logic.effect.blocking.*;
 import tcgwars.logic.effect.event.*;
 import tcgwars.logic.effect.getter.*;
-import tcgwars.logic.effect.special.*;
-import tcgwars.logic.util.*;
+import tcgwars.logic.effect.special.*
+import tcgwars.logic.exception.NotEnoughEnergyException;
+import tcgwars.logic.util.*
 
 /**
  * @author axpendix@hotmail.com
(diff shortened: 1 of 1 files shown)


src/tcgwars/logic/impl/gen4/MysteriousTreasures.groovy
```diff
@@ -561,9 +562,32 @@ public enum MysteriousTreasures implements LogicCardInfo {
...
             text "As long as Honchkrow has the Energy necessary to use its attack, each of your Murkrow can use Honchkrow’s attack as its own without the Energy necessary to use that attack."
-            delayedA {
-              //TODO
-              //Future TODO: This should work even when Honchkrow levels up, as it has the attack and Poké-Power of the base one (and Lv.X's attacks should also be available to be copied)
+            def selfOwner = self.owner
+            def selfPcs = self
+            getterA GET_MOVE_LIST, { h ->
+              PokemonCardSet hTarget = h.effect.target
+              if (hTarget.active && hTarget.owner == selfOwner && hTarget.name == "Murkrow" && selfPcs.benched) {
+                List<Move> honchkrowMoves = bg.em().activateGetter(new GetMoveList(selfPcs))
```
axpendix, line 570: Is this necessary? The body mentions "Honchkrow’s attack" and this should only mean "Dark Wing Flaps" therefore instead of this getter, using `thisCard.moves` would be safer to avoid any potential infinite loops with some weird combo.

starg09: If Honchkrow is leveled up, it should obtain the attacks from both its base form and the Level X. I believe originally this code took the attacks that way, if you think it'll avoid corruption issues it can be reverted.

axpendix: Ah, I am not used to Level Xs yet :D Well in this case I believe [a recursion guard like this](https://github.com/axpendix/tcgone-engine-contrib/blob/master/src/tcgwars/logic/impl/gen4/LegendsAwakened.groovy#L1048) could be useful here.

starg09:
> I believe there should be a built-in infinite loop guard in the `EventManager` inorderto prevent such edge cases by default.

Ah, great! Need to look back into this soon, been a while since I've tested it so it might be completely broken for all I know 😅 

But that should be really helpful, this one was quite tricky to have going properly!

PR #1280, opened 2021-06-21
Chilling Reign fixes / SWSH Promos update

scoot404:
This PR does the following:
* Fix incorrect attack names from Chilling Reign
* Fix small bugs regarding incorrect numbers in Chilling Reign
* Fixes Ditto V's ability name
* Updates SWSH promos up to SWSH119 (Eevee VMAX is not implemented though)

PR #1283, opened 2021-06-26
Chilling Reign Fixes

luongthomas:
Addresses some issues brought up on trello


starg09 (changes requested):
Seeing some changes that may have been applied in #1280 , would look into merging from master into the PR and fixing conflicts before reviewing it.

PR #1288, opened 2021-07-01
Fix Chilling Reign Bugs

luongthomas:
Screen Shot 2021-07-01 at 10 43 19 AM

### DIFF
--- a/src/tcgwars/logic/impl/gen8/ChillingReign.groovy
+++ b/src/tcgwars/logic/impl/gen8/ChillingReign.groovy
@@ -2964,7 +2964,30 @@ public enum ChillingReign implements LogicCardInfo {
             attackRequirement {}
             onAttack {
               damage 50
-              // TODO
+
+              if (bg.em().retrieveObject("last_rapid_strike_supporter_play_turn") == bg.turnCount && opp.bench) {
+                multiSelect(opp.bench, 2, "Deal damage to which?").each {
+                  targeted(it) {
+                    damage 50, it
+                  }
+                }
+              }
+            }
+            globalAbility {
+              def flag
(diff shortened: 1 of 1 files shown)


src/tcgwars/logic/impl/gen8/ChillingReign.groovy
```diff
@@ -2218,7 +2218,7 @@ public enum ChillingReign implements LogicCardInfo {
             delayedA {
               after PROCESS_ATTACK_EFFECTS, {
                 bg.dm().each {
-                  if (it.from == self && it.to.benched && (it.to.pokemonV || it.to.pokemonVMAX || it.to.pokemonGX) && it.dmg.value) {
+                  if (it.from.owner == self.owner && it.from.rapidStrike && it.to.benched && (it.to.pokemonV || it.to.pokemonVMAX || it.to.pokemonGX) && it.dmg.value) {
```
flagrama, line 2221: I don't see anything here for
> You can't use more than 1 Throwing Coach Ability at a time.

Also PokemonVMAX are PokemonV so no need for both checks, only PokemonV is needed.

```suggestion
                  if (it.from.owner == self.owner && it.from.rapidStrike && it.to.benched && (it.to.pokemonV || it.to.pokemonGX) && it.dmg.value) {
```

flagrama: Marked unresolved. Only the second half was handled.

luongthomas: Thanks for catching that, I've added it in my latest commit

flagrama (approved):

PR #1289, opened 2021-07-09
add best of the game set

TheAquaPiplup:
best of the gane cards added
expect bad code


luongthomas (approved):
Looks good!  Next we'll need scans

PR #1291, opened 2021-07-17
Fix Gengar_57 CRE Last Gift

sujayvadlakonda:
# Bug Fix Gengar (CRE 57), Ability Last Gift

Expected Behavior:
When Gengar's knocked out from damage by opposing pkmn, owner of Gengar selects up to 2 cards from deck and adds to hand

What's actually happening:
When Gengar's knocked out from damage by opposing pkmn, **opponent of** owner of Gengar selects up to 2 cards from deck and adds to **opponent's** hand

### DIFF
--- a/src/tcgwars/logic/impl/gen8/ChillingReign.groovy
+++ b/src/tcgwars/logic/impl/gen8/ChillingReign.groovy
@@ -1540,10 +1540,9 @@ public enum ChillingReign implements LogicCardInfo {
             text "If this Pokémon is Knocked Out by damage from an attack from your opponent’s Pokémon, search your deck for up to 2 cards and put them into your hand. Then, shuffle your deck."
             delayedA (priority: LAST) {
               before (KNOCKOUT, self) {
-                // Here opp refers to the owner of Gengar
-                if ((ef as Knockout).byDamageFromAttack && bg.currentTurn == self.owner.opposite && opp.deck) {
+                if ((ef as Knockout).byDamageFromAttack && bg.currentTurn == self.owner.opposite && self.owner.pbg.deck) {
                   bc "$thisAbility activates"
-                  self.owner.pbg.deck.select(min: 1, max: 2, "Search for up to 2 cards").moveTo(opp.hand, hidden: true)
+                  self.owner.pbg.deck.select(min: 1, max: 2, "Search for up to 2 cards").moveTo(self.owner.pbg.hand, hidden: true)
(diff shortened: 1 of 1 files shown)


src/tcgwars/logic/impl/gen8/ChillingReign.groovy
```diff
@@ -1540,9 +1540,10 @@ public enum ChillingReign implements LogicCardInfo {
             text "If this Pokémon is Knocked Out by damage from an attack from your opponent’s Pokémon, search your deck for up to 2 cards and put them into your hand. Then, shuffle your deck."
             delayedA (priority: LAST) {
               before (KNOCKOUT, self) {
-                if ((ef as Knockout).byDamageFromAttack && bg.currentTurn == self.owner.opposite && my.deck) {
+                // Here opp refers to the owner of Gengar
+                if ((ef as Knockout).byDamageFromAttack && bg.currentTurn == self.owner.opposite && opp.deck) {
                   bc "$thisAbility activates"
-                  self.owner.pbg.deck.select(min: 1, max: 2, "Search for up to 2 cards").moveTo(my.hand, hidden: true)
+                  self.owner.pbg.deck.select(min: 1, max: 2, "Search for up to 2 cards").moveTo(opp.hand, hidden: true)
```
starg09, line 1546: Proposed edit should work, but would propose using a reference to the pokemon itself (instead of a value dependant on the battleground, such as my/opp in this instance)

```suggestion
                if ((ef as Knockout).byDamageFromAttack && bg.currentTurn == self.owner.opposite && self.owner.pbg.deck) {
                  bc "$thisAbility activates"
                  self.owner.pbg.deck.select(min: 1, max: 2, "Search for up to 2 cards").moveTo(self.owner.pbg.hand, hidden: true)
```

sujayvadlakonda: Fixed, why are we avoiding using battleground?

starg09: Not really avoiding it, but in case the card's effect is reused somewhere else (which isn't that uncommon of a thing), having `self.owner.pbg` will always point to the correct player, regardless on what turn it happens on. Using `my/opp` is totally fine, but reference based on current turn and might not be as clear to follow when checking for potential bugs/misinteractions.

starg09 (approved):

PR #1298, opened 2021-08-07
eevee heroes vmax special set

TheAquaPiplup:
adding in the eevee heroes vmax special set cards


flagrama:
Do you plan to add more to this, or should I just merge this as is?

PR #1300, opened 2021-08-15
Implement Blue Sky Stream Grass

flagrama:
Depends on https://github.com/axpendix/tcgone/pull/186

### DIFF
--- a/src/tcgwars/logic/impl/gen8/BlueSkyStream.groovy
+++ b/src/tcgwars/logic/impl/gen8/BlueSkyStream.groovy
@@ -201,15 +201,13 @@ public enum BlueSkyStream implements LogicCardInfo {
               if (ef.reason == PLAY_FROM_HAND
                 && bg.currentTurn == self.owner
                 && self.owner.pbg.deck
-                && lastUsedTurn != bg.turnCount
                 && confirm("Use $thisAbility?")) {
                 powerUsed()
                 def list = deck.search text, {
                   it.cardTypes.is(EVOLUTION) && it.predecessor == self.name
                 }
-                if (list) {
-                  def card = list.first()
-                  evolve self, card, OTHER
+                list.each {
+                  evolve self, it, OTHER
                 }
                 shuffleDeck()
               }
@@ -423,6 +421,9 @@ public enum BlueSkyStream implements LogicCardInfo {
           onAttack {
             def list = my.hand.select min:1, max:2, text, { BASIC_ENERGY in it.cardTypes }
             damage 60 * list.size()
+            afterDamage {
+              list.discard()
+            }
           }
         }
       };
@@ -435,8 +436,10 @@ public enum BlueSkyStream implements LogicCardInfo {
           onAttack {
             def list = my.discard.findAll { BASIC_ENERGY in it.cardTypes }
             damage 20 * list.size()
-            list.moveTo my.deck
-            if (list.notEmpty()) shuffleDeck()
+            afterDamage {
+              list.moveTo my.deck
+              if (list.notEmpty()) shuffleDeck()
+            }
           }
         }
         move "Fire Blast", {
@@ -608,6 +611,7 @@ public enum BlueSkyStream implements LogicCardInfo {
             selectedEnergyList.each {
               attachEnergy my.all.select("Attach $it.name to?"), it
             }
+            shuffleDeck()
           }
         }
         move "Icy Snow", {
--- a/src/tcgwars/logic/impl/gen8/BlueSkyStream.groovy
+++ b/src/tcgwars/logic/impl/gen8/BlueSkyStream.groovy
@@ -488,7 +488,6 @@ public enum BlueSkyStream implements LogicCardInfo {
         move "Tongue Slap", {
           text "10 damage."
           energyCost W
-          attackRequirement {}
           onAttack {
             damage 10
           }
@@ -496,7 +495,6 @@ public enum BlueSkyStream implements LogicCardInfo {
         move "Wave Splash", {
           text "20 damage."
           energyCost W, C
-          attackRequirement {}
           onAttack {
             damage 20
           }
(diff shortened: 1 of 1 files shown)


src/tcgwars/logic/impl/gen8/BlueSkyStream.groovy
```diff
@@ -195,13 +196,29 @@ public enum BlueSkyStream implements LogicCardInfo {
         weakness R
         bwAbility "Solar Evolution", {
           text "When you attach an Energy from your hand to this Pokémon during your turn, you may use this Ability. Search your deck for a card that evolves from this Pokémon and put it onto this Pokémon to evolve it. Then, shuffle your deck."
-          actionA {
+          delayedA {
+            after ATTACH_ENERGY, self, {
+              if (ef.reason == PLAY_FROM_HAND
+                && bg.currentTurn == self.owner
+                && self.owner.pbg.deck
+                && lastUsedTurn != bg.turnCount
```
ufodynasty, line 204: This ability can be used multiple times in a turn.

src/tcgwars/logic/impl/gen8/BlueSkyStream.groovy
```diff
@@ -195,13 +196,29 @@ public enum BlueSkyStream implements LogicCardInfo {
...
+                && confirm("Use $thisAbility?")) {
+                powerUsed()
+                def list = deck.search text, {
+                  it.cardTypes.is(EVOLUTION) && it.predecessor == self.name
+                }
+                if (list) {
```
ufodynasty, line 210: do you think list.each would be more clear here?

src/tcgwars/logic/impl/gen8/BlueSkyStream.groovy
```diff
@@ -262,15 +276,18 @@ public enum BlueSkyStream implements LogicCardInfo {
         move "Return", {
           text "30 damage. You may draw cards until you have 6 cards in your hand."
           energyCost G
-          attackRequirement {}
           onAttack {
             damage 30
+            def count = 6 - my.hand.size()
+            if (count > 0
```
ufodynasty, line 282: Perhaps add a deck check here?

src/tcgwars/logic/impl/gen8/BlueSkyStream.groovy
```diff
@@ -363,9 +381,13 @@ public enum BlueSkyStream implements LogicCardInfo {
         move "Victory Dive", {
           text "30 damage. You may search your deck for up to 2 cards and put them into your hand. Then, shuffle your deck."
           energyCost C, C
-          attackRequirement {}
           onAttack {
             damage 30
+            if (my.deck && confirm(text)) {
```
ufodynasty, line 386: I personally don't like to use confirm(text) for damaging attacks as it could give the impression that selecting no would result in the attack not doing damage.

src/tcgwars/logic/impl/gen8/BlueSkyStream.groovy
```diff
@@ -387,37 +410,41 @@ public enum BlueSkyStream implements LogicCardInfo {
...
           text "60× damage. Discard up to 2 basic Energy cards from your hand. This attack does 60 damage for each card you discarded in this way."
           energyCost R
-          attackRequirement {}
+          attackRequirement {
+            assert my.hand.basicEnergyCardCount() : "No basic Energy cards in your hand"
+          }
           onAttack {
-            damage 60
+            def list = my.hand.select min:1, max:2, text, { BASIC_ENERGY in it.cardTypes }
+            damage 60 * list.size()
```
ufodynasty, line 425: You need to discard the energy cards after damage.

src/tcgwars/logic/impl/gen8/BlueSkyStream.groovy
```diff
@@ -387,37 +410,41 @@ public enum BlueSkyStream implements LogicCardInfo {
...
-        move "Surging Flame", {
+        move "Surging Flames", {
           text "20+ damage. This attack does 20 more damage for each basic Energy card in your discard pile. Then, shuffle those cards into your deck."
           energyCost R
-          attackRequirement {}
           onAttack {
-            damage 20
+            def list = my.discard.findAll { BASIC_ENERGY in it.cardTypes }
+            damage 20 * list.size()
+            list.moveTo my.deck
```
ufodynasty, line 438: This needs to be done after damage.

src/tcgwars/logic/impl/gen8/BlueSkyStream.groovy
```diff
@@ -548,15 +574,20 @@ public enum BlueSkyStream implements LogicCardInfo {
...
           onAttack {
             damage 30
+            applyAfterDamage ASLEEP
           }
         }
         move "Daruma Headbutt", {
           text "130 damage. If this Pokemon has any damage counters on it, this attack can be used for [W]."
-          energyCost W, W, C
-          attackRequirement {}
+          energyCost W
```
ufodynasty, line 584: I think this should be done as a custom ability that modifies the attack cost. The current implementation breaks with effects that increase attack cost. Ex. If this Pokémon has WWC attached and no damage counters, and there is a wondrous labyrinth in play, it could still attack.

src/tcgwars/logic/impl/gen8/BlueSkyStream.groovy
```diff
@@ -568,15 +599,20 @@ public enum BlueSkyStream implements LogicCardInfo {
...
+          attackRequirement {
+            assert my.deck : "Your deck is empty"
+          }
           onAttack {
-
+            def top6List = my.deck.subList 0, 6
+            def selectedEnergyList = top6List.select text, { ENERGY in it.cardTypes }
+            selectedEnergyList.each {
+              attachEnergy my.all.select("Attach $it.name to?"), it
+            }
```
ufodynasty, line 610: The deck should be shuffled after this effect.

ufodynasty (commented):
Thank you for doing this ahead of the release date!

src/tcgwars/logic/impl/gen8/BlueSkyStream.groovy
flagrama: `draw` uses `DrawCard` which properly handles no more cards in deck and broadcasts a message saying as such.

src/tcgwars/logic/impl/gen8/BlueSkyStream.groovy
flagrama: I'm 99% sure you aren't allowed to look at your deck and choose nothing. You either choose to check and get at least 1 card, or you choose not to check.

src/tcgwars/logic/impl/gen8/BlueSkyStream.groovy
flagrama: This is copied from Boltund BST. While I agree that if it is causing bugs it should be fixed, I don't think it should delay the set if another set already does it this way. It can be fixed for both at the same time (and I don't really have time to figure it out atm)

src/tcgwars/logic/impl/gen8/BlueSkyStream.groovy
flagrama: Please check that the new logic makes sense.

src/tcgwars/logic/impl/gen8/BlueSkyStream.groovy
flagrama: > When I use Togekiss VMAX's move "Die Glide" and look at my deck, can I not select any cards from my deck?
No, you can't.
When you look at the deck, you must choose at least one card to add to your hand.

ufodynasty: You are correct about that. 
```suggestion
            if (my.deck && confirm("Search your deck for up to 2 cards?")) {
```
Is more what I had in mind. While it is an extremely minor tweak having the damage portion of the attack in the confirm message could create ambiguity as to whether selecting no would result in the whole attack being canceled or just the optional portion of it. I'm fine to leave this as is but wanted to clarify my comment.

src/tcgwars/logic/impl/gen8/BlueSkyStream.groovy
ufodynasty: I think Boltund copied it from tapu fini? I agree that this seems like enough of an edge case that we can leave it for now.

ufodynasty (approved):
LGTM

src/tcgwars/logic/impl/gen8/BlueSkyStream.groovy
flagrama: Oh, fair enough, you can go ahead and commit that suggestion.

PR #1301, opened 2021-09-10
Bug fixes from the forums

ufodynasty:
This branch will accumulate small bug fixes reported on the forums
*https://forum.tcgone.net/t/br-rockets-meowth-ex-team-rocket-returns-46-snatch-and-run-swoop-teleporter-cannot-be-selected/12484

### DIFF
--- a/src/tcgwars/logic/impl/gen3/TeamRocketReturns.groovy
+++ b/src/tcgwars/logic/impl/gen3/TeamRocketReturns.groovy
@@ -1428,7 +1428,7 @@ public enum TeamRocketReturns implements LogicCardInfo {
               assert my.deck
             }
             onAttack {
-              if(my.deck.search(max : 1,"Select a Pokémon Tool card or Rocket’s Secret Machine.",cardTypeFilter(POKEMON_TOOL)).showToOpponent("Selected card.").moveTo(my.hand)){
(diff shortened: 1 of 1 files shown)


src/tcgwars/logic/impl/gen4/DiamondPearlPromos.groovy
```diff
@@ -651,7 +654,7 @@ public enum DiamondPearlPromos implements LogicCardInfo {
             delayedA {
               before BEGIN_TURN, {
                 if (my.all.find {it.name == "Darkrai" }) {
-                  heal 10, self, POKEBODY
```
flagrama, line 654: Doesn't the IDE yell at you that `Source.` isn't needed after adding the import on line 8?

ufodynasty: Not sure why but removing the `Source.` causes POKEBODY to still be interpreted as an EffectType instead of a Source despite EffectType not being a legal argument.

flagrama: Oh, I see EffectType is also wildcarded there, and it is done first. Yeah that makes sense.

ufodynasty:
Really unhappy with the telepass fix as it still is bugged with ampharos' jamming pokebody. Eventually I would like to rework this style of effect to use the effect of the card without using playTrainer.

PR #1303, opened 2021-09-21
Implement Diamond and Pearl Trainer Kit

wertercatt:
This is technically part of the 2008 Standard Format. Mostly did this to ease myself back into working with the codebase though, if I'm honest.


src/tcgwars/logic/impl/gen4/MysteriousTreasures.groovy
```diff
@@ -3598,10 +3598,9 @@ public enum MysteriousTreasures implements LogicCardInfo {
             energyCost R, R, R, R
             attackRequirement {}
             onAttack {
+              discardSelfEnergy R,R
```
flagrama, line 3601: We now have `discardSelfEnergyAfterDamage` in TcgStatics which will handle discarding energies in an attack in the correct order (announce attack, select energy to discard, perform attack, discard selected energies)

```suggestion
              discardSelfEnergyAfterDamage R,R
```

wertercatt: Technically, that doesn't work for this card. The change I made was based on a TODO that's still not entirely complete. https://www.tcgcollector.com/cards/3803/magmortar-lv-x-mysterious-treasures-123-123 The card explicitly requires the energy to be discarded, and then the damage is dealt.

src/tcgwars/logic/impl/gen4/DPTrainerKit.groovy
```diff
@@ -0,0 +1,151 @@
...
+import static tcgwars.logic.groovy.TcgBuilders.*;
+import static tcgwars.logic.groovy.TcgStatics.*
+import static tcgwars.logic.card.Resistance.ResistanceType.*
+
+
+import tcgwars.logic.card.*
+import tcgwars.logic.util.*;
+
+/**
+ * @author axpendix@hotmail.com
```
wertercatt, line 18: Also I'm not sure if these author tags are formatted right. The groovy file I based this off of was an old one I implemented in 2019.

flagrama: The author javadoc tag doesn't really do anything anyway. I find the git history to have the same use and haven't bothered adding my email to any of the files I've worked on. So long as it isn't causing an error whatever you put there should be fine.

wertercatt: Sounds good then.

flagrama:
This needs to be added to src/tcgwars/logic/card/Collection.java as well. I also need to double check with the others if we need to have an equivalent https://github.com/tcgone/carddb set entry before it can be merged.

src/tcgwars/logic/impl/gen4/DPTrainerKit.groovy
```diff
@@ -0,0 +1,151 @@
...
+
+
+  static Type C = COLORLESS, R = FIRE, F = FIGHTING, G = GRASS, W = WATER, P = PSYCHIC, L = LIGHTNING, M = METAL, D = DARKNESS, Y = FAIRY, N = DRAGON;
+
+  protected CardTypeSet cardTypes;
+  protected String name;
+  protected Rarity rarity;
+  protected String collectionLineNo;
+
+  WorldCollection(String name, String collectionLineNo, Rarity rarity, List<CardType> cardTypes) {
```
flagrama, line 50: This line doesn't seem to have been updated from where it was copied. In addition, the file, enum, and any other references should probably use the full name like we do for all the other sets without any acronyms, so all `DPTrainerKit`s should be `DiamondPearlTrainerKit`

```suggestion
  DiamondPearlTrainerKit(String name, String collectionLineNo, Rarity rarity, List<CardType> cardTypes) {
```

wertercatt: Fixed

flagrama:
If you are good with this as is I can merge it, however it needs to be added to Collection.java to show up on the Dev server and Play sandbox.

We also think it might be better to split the trainer kits into their own set files before that as that seems to be the norm (Bulbapedia, PTCGO) for the places that don't ignore it entirely (Pokemon.com, Pokemon TCG API).

As these aren't going to cause any errors it's up to you whether you want to do them in this PR or in another one after this is merged.

PR #1306, opened 2021-10-03
Automatically shuffle deck on search & Pokémon of Towering Perfection

flagrama:
Trainers still need done.

@axpendix Tried to add auto deck shuffling any time the deck is searched and seems to work. Would appreciate you take a look to make sure it doesn't seem like it will break anything.


axpendix:
I am okay with deck shuffle change

PR #1318, opened 2022-05-21
Add missing imports from the request builders refactor

starg09:
This fixes the build process after the refactors made in https://github.com/axpendix/tcgone/commit/a551752e545af93d320a3bf047dca816ea7e61e3.

### DIFF
--- a/src/tcgwars/logic/util/CardList.groovy
+++ b/src/tcgwars/logic/util/CardList.groovy
@@ -14,9 +14,9 @@ import tcgwars.logic.card.Type;
 import tcgwars.logic.card.energy.EnergyCard;
 import tcgwars.logic.effect.advanced.ExtraEnergyCalculator;
 import tcgwars.logic.effect.basic.Discard;
-import tcgwars.logic.client.CardSelectUIRequestBuilder;
-import tcgwars.logic.client.CardSelectUIRequestBuilder.CustomCardFilter;
-import tcgwars.logic.client.CardSelectUIRequestBuilder.CustomPassFilter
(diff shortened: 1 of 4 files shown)


src/tcgwars/logic/util/CardList.groovy
```diff
@@ -348,14 +348,14 @@ public class CardList extends ArrayList<Card> {
     }
     if (playerType != TcgStatics.bg().currentThreadPlayerType) {
       TcgStatics.block()
+      def ret = TcgStatics.bg().getClient(playerType).selectCard(new CardSelectUIRequestBuilder()
+        .setMinMax(min, max)
+        .setInfo(info)
+        .setCards(cards)
+        .setCustomCardFilter(filter != null ? filter as CustomCardFilter : null)
+        .setCustomPassFilter(passFilter != null ?  passFilter as CustomPassFilter : null)
+        .setShowAsHidden(hidden))
```
axpendix, line 357: I think this should have stayed outside the if clause, don't you think?

starg09: Honestly didn't look past keeping it as it was, but correcting the casting to the new reference. But yes, I may be wrong but shouldn't this work properly without the ternary if? Will test it out in a bit

starg09: Yup, seems to be what's causing some Engine errors. Will push a fix on a separate PR in a bit :)

starg09: Fixed in https://github.com/axpendix/tcgone-engine-contrib/pull/1319

PR #1323, opened 2022-07-17
Add VSTAR enum and VSTAR Power check

flagrama:
Some helpful VSTAR stuff in case someone wants to start implementing them


starg09 (approved):
LGTM

PR #1324, opened 2022-08-05
Update EeveeHeroesVmaxSpecialSet.groovy

kazam1228:
Wanna try and contribute to EVS here and there so we can work on updating tcgONE to current standard


flagrama:
As written, this now won't damage the active Pokémon. You should add a second `damage` call.

flagrama:
Also, feel free to join the Discord if you have development questions: https://discord.gg/MTBGxVE

PR #1330, opened 2022-10-09
Allow setting Source of a Devolve

flagrama:
Allows fixing bugs like https://forum.tcgone.net/t/bug-report-bent-spoon-fco-93-espeon-exs-miraculous-s/2867

### DIFF
--- a/src/tcgwars/logic/groovy/TcgStatics.groovy
+++ b/src/tcgwars/logic/groovy/TcgStatics.groovy
@@ -611,10 +611,14 @@ class TcgStatics {
     bg().em().run(new CantEvolve(pcs, bg().getTurnCount()));
     bg().em().run(new ActivateAbilities((PokemonCard) card, pcs, reason));
   }
-  static devolve (PokemonCardSet pcs, Card card, CardList newLocation){
-    def blocked = bg().em().run(new MoveCard(card, newLocation));
-    if (blocked) {
-      return;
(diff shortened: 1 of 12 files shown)


src/tcgwars/logic/groovy/TcgStatics.groovy
```diff
@@ -611,10 +611,14 @@ class TcgStatics {
...
   }
-  static devolve (PokemonCardSet pcs, Card card, CardList newLocation){
-    def blocked = bg().em().run(new MoveCard(card, newLocation));
-    if (blocked) {
-      return;
+  static devolve (PokemonCardSet pcs, Card card, CardList newLocation, Source source=ATTACK){
+    def blockedDevolve = bg().em().run(new Devolve(pcs, source))
+    if (blockedDevolve) {
+      return
+    }
```
axpendix, line 618: `Devolve` runs `CheckAbilities`. By the time it runs, the card has to be moved away, but this change prevents that, thus the abilities will be incorrect after this change.

axpendix:
@flagrama any updates on this?

flagrama:
Not yet, though I'm not sure exactly how to handle it either since either the card move or the devolve can be blocked, and if the card gets moved but the devolve gets blocked it can trigger effects that shouldn't be triggered at that time. Will probably look into moving it to a temporary card list, then devolve, then move to the desired location.

axpendix:
This PR is irrelevant thus closed. `Devolve` has already been refactored as a full effect and I'm working on source refactoring to fix sources of all use cases at the core.

PR #1331, opened 2022-10-09
Start of massive refactor to set targets and source properly

flagrama:
TcgStatics changes show the  changes that should be made to `delayed` effects to set targets and sources properly. I haven't really thought much about setting EffectTypes yet, but will probably work well in addition to the new event management.

The Unbroken Bonds STEALTHY_HOOD change shows how the new event management in the sister PR in the engine can be used to ensure we only block effects from abilities used by the opponent's Pokémon. Currently both `resolvedTarget` and `thisPokemon` are checked (all Groovy delayed checks extend AbstractDelayed**Move**Effect so currently thisPokemon will work on more abilities this way for now, but once all abilities set their target


axpendix:
@flagrama any updates on this?

flagrama:
This is the sister PR to PR 204 in the engine repo. My questions over there pertain to this, but I'm discussing over there since the solution is going to require changes to the engine.

PR #1332, opened 2022-10-11
Batch of various card fixes

starg09:
* Added a static method for "Bring Down" attacks, which is now used for both Gardevoir Lv.X (Secret Wonders) and Greninja (Unbroken Bonds).
  - This should fix some issues that Gardevoir used to have.
* Fixed Trapinch (Secret Wonders) not asserting the opponent having any benched Pokémon in order to use its "Inviting Trap" attack.
* Fixed Beldum Delta (Delta Species 59) not reducing its retreat cost properly. Now follows the (working?) code of the modern "spiritual reprint" of Beldum (Celestial Storm)
  - In addition, both cards now use `.count{...}` instead of `.findAll{...}.size()`
* Attempted fixing Coating [M] Energy (Vivid Voltage). It should now properly remove weaknesses from the [M] Pokémon it's attached to.
* Fixed Ho-Oh ex (Unseen Forces) missing it's "ex" CardType (thus giving only 1 prize).
* Tried fixing an unreported NPE in Cape of Toughness (Darkness Ablaze) due to not checking self not being null.


src/tcgwars/logic/groovy/TcgStatics.groovy
```diff
@@ -1678,6 +1678,23 @@ class TcgStatics {
...
+      int min = list.get(0).remainingHP.value
+      while (list.notEmpty && list.get(0).remainingHP.value==min) {
+        tar.add(list.remove(0))
+      }
+      //TODO: Heavily improve this selection, in case both players have tied Pokémon. Make it clearer to pick.
+      def pcs = tar.select("Choose which of these Pokémon will be knocked out.")
+      bc "${pcs.owner}'s $pcs was selected for Knock Out"
+      targeted (pcs) {
+        new Knockout(pcs).run(bg)
+      }
```
axpendix, line 1694: Knockout is already targeted, no?

src/tcgwars/logic/groovy/TcgStatics.groovy
```diff
@@ -1678,6 +1678,23 @@ class TcgStatics {
...
 
+  static bringDown (Object delegate) {
+    delegate.onAttack {
+      def list = all.findAll{it!=delegate.self}.sort(false) {p1,p2 -> p1.remainingHP.value <=> p2.remainingHP.value} as PcsList
+      def tar = new PcsList()
+      int min = list.get(0).remainingHP.value
+      while (list.notEmpty && list.get(0).remainingHP.value==min) {
+        tar.add(list.remove(0))
+      }
+      //TODO: Heavily improve this selection, in case both players have tied Pokémon. Make it clearer to pick.
```
axpendix, line 1689: What does this mean?

src/tcgwars/logic/groovy/TcgStatics.groovy
```diff
@@ -1678,6 +1678,23 @@ class TcgStatics {
...
   }
 
+  static bringDown (Object delegate) {
+    delegate.onAttack {
+      def list = all.findAll{it!=delegate.self}.sort(false) {p1,p2 -> p1.remainingHP.value <=> p2.remainingHP.value} as PcsList
+      def tar = new PcsList()
+      int min = list.get(0).remainingHP.value
+      while (list.notEmpty && list.get(0).remainingHP.value==min) {
+        tar.add(list.remove(0))
+      }
```
axpendix, line 1688:
```suggestion
      def list = all.findAll{it!=delegate.self}.sort(false) {p1,p2 -> p1.remainingHP.value <=> p2.remainingHP.value}
      def min = list.get(0).remainingHP
      def tar = list.findAll{it.remainingHP==min} as PcsList
``````

File: src/tcgwars/logic/impl/gen8/AmazingVoltTackle.groovy [hunk 4]
```diff
@@ -2167,7 +2167,7 @@ public enum AmazingVoltTackle implements LogicCardInfo {
         def eff
         onPlay {reason->
           eff = getter GET_WEAKNESSES, self, { holder ->
-            if (self.types.contains(M)) {
+            if (self != null && self.types.contains(M)) {
```
axpendix, line 2170:
```suggestion
            if (self && self.types.contains(M)) {
```

src/tcgwars/logic/groovy/TcgStatics.groovy
starg09: For context, this was already in the bring down code from greninja, it wasn't added now.

src/tcgwars/logic/groovy/TcgStatics.groovy
axpendix: Okay

src/tcgwars/logic/groovy/TcgStatics.groovy
axpendix: As you might be aware, the new pokemon selection system has been rolled out. It makes all possible selections glow on the board. So there can be no confusion.

starg09: Ahh, great this is solved then! Can remove that comment 😅 

Will check on test to confirm tho

starg09: Yup, works as a charm!

PR #1334, opened 2022-11-08
Reimplement most RS-HL trainers and special energies in groovy, fix some other various cards

starg09:
### (Depends on https://github.com/axpendix/tcgone/pull/208 being merged first)

Main change is the reimplementations of the trainers and special energies from the first 4 main sets in the EX era. These were implemented in Legacy Java code, and should eventually be fully migrated but the trainers feel like a good first step, as to avoid reocurring bugs related to them. Some are still pending the migration, including:
  - Balloon Berry (EX Dragon 82): Currently there's some hoops done around the legacy implementation, need to take proper time into removing those and doing this purely through delayed effects.
  - Ancient Technical Machines (EX Hidden Legends 84-86): No TM has been redone in groovy as of yet, this is a pending matter to be addressed later on in a separate PR.
  - All Team Magma vs Team Aqua trainers/energies: Yeah this one I just didn't feel like pushing yet, the trainers are mostly niche at best anyway. Will push a separate PR for the whole set eventually, as some Pokémon do see a decent amount of play.

---

### Other notable changes:

* Poké Ball (Jungle) now shows the selected Pokémon to the opponent, as other similar cards also do already.
* Removed some redundant checks from Energy Switch (FRLG 90)
* Unown (Unseen Forces Q) should now require having cards in deck before using Hidden Power. It should also reveal the selected cards to the opponent.
* Redirected Energy Switch (Black & White 94) into the implementation from FireRed LeafGreen.
* Reimplemented Energy Recycle System (Celestial Storm 128) as to not depend on the legacy implementation from Dragon. That version now copies this version.


starg09:
[Reimplement legacy GEN1, ECARD and GEN3 sets in Groovy](https://trello.com/c/LRDpynRc/14-reimplement-legacy-gen1-ecard-and-gen3-sets-in-groovy)

axpendix (approved):

PR #1351, opened 2023-09-06
Mass purge of semicolons from groovy files

starg09:
Legacy from really old commits from years ago, that trigger over 20k warnings across multiple IDEs (could be ignored, but it's honestly unnecessary additions in these files)

No logic changes, all automatically pruned via IntelliJ.

Requires https://github.com/axpendix/tcgone/pull/220 to be merged


starg09:
Closing this while there's active developments, as it'd force to fix merge conflicts on most if not all of them.

PR #1359, opened 2024-08-25
Aug 2024 fixes

dxdydzd:


src/tcgwars/logic/impl/gen3/DeltaSpecies.groovy
```diff
@@ -2546,7 +2546,7 @@ public enum DeltaSpecies implements LogicCardInfo {
         def eff
         def eff2
         onPlay { reason ->
-          eff = getter(GET_RETREAT_COST) { h ->
+          eff = getterA(GET_RETREAT_COST, BEFORE_LAST) { h ->
```
axpendix, line 2549:
```suggestion
          eff = getter(GET_RETREAT_COST, BEFORE_LAST) { h ->
```
getterA only works in abilities, it won't work here.

axpendix (changes requested):

src/tcgwars/logic/impl/gen3/FireRedLeafGreen.groovy
```diff
@@ -1583,7 +1583,7 @@ public enum FireRedLeafGreen implements LogicCardInfo {
             onAttack {
               damage 10
               afterDamage {
-                apply PARALYZED
+                flipThenApplySC PARALYZED
```
axpendix, line 1586: `flipThenApplySC` already includes an `afterDamage` clause, so it's best to rewrite this. 
```suggestion
                flip { apply PARALYZED }
```

axpendix (approved):

PR #1365, opened 2024-09-19
Ensure metronome works properly if attack itself is not bugged.

flagrama:


src/tcgwars/logic/impl/gen4/DiamondPearl.groovy
```diff
@@ -20,21 +20,12 @@ import static tcgwars.logic.effect.special.SpecialConditionType.*
...
-import tcgwars.logic.effect.*;
-import tcgwars.logic.effect.ability.*;
-import tcgwars.logic.effect.advanced.*;
-import tcgwars.logic.effect.basic.*;
-import tcgwars.logic.effect.blocking.*;
-import tcgwars.logic.effect.event.*;
-import tcgwars.logic.effect.getter.*;
-import tcgwars.logic.effect.special.*;
+import tcgwars.logic.card.trainer.*
+import tcgwars.logic.effect.basic.*
```
axpendix, line 28: Can you please restore these star imports? This is probably result of IntelliJ "optimizing" them but we observed in the past some silent issues because it removed some legit ones too.

flagrama: I do notice that `import tcgwars.logic.card.{energy,pokemon,trainer}` all are "errors" in groovy files. I assume that's a remnant of a refactoring of card stuff? Can those be removed at least?

axpendix: Yeah of course, my bad.

PR #1367, opened 2024-09-23
Sep 2023 fixes

dxdydzd:


src/tcgwars/logic/impl/gen8/SwordShield.groovy
```diff
@@ -952,7 +952,7 @@ public enum SwordShield implements LogicCardInfo {
         }
       };
       case CINDERACE_36:
-      return evolution (this, from:"Raboot", hp:HP170, type:R, retreatCost:1) {
+      return evolution (this, from:"Raboot", hp:HP170, type:R, retreatCost:2) {
```
flagrama, line 955: Should either add an errata note comment for when those are displayed at a later date, or should upload an updated card render to the carddb.
```suggestion
      return evolution (this, from:"Raboot", hp:HP170, type:R, retreatCost:2) {
        // errata: Misprinted, retreat cost is actually 2. https://www.pokemon.com/us/pokemon-news/errata-for-cinderace-in-pokemon-tcg-sword-shield
```

flagrama (approved):

PR #1371, opened 2024-12-24
Evolution Pokémon

dxdydzd:
Some Gen 4 Pokémon incorrectly use "Evolved Pokémon" to refer to Evolution Pokémon. Only a few of them received official errata (the ones that refer to Stage 2s, i.e. Alakazam MT, Glalie MT, Staraptor SF).

Two of them (Dodrio MT and Flygon RR) have rulings on the Japanese FAQ page to confirm that they refer to Evolution Pokémon. This PR fixes those two, plus any others that don't have Japanese rulings, but use identical wording.

More info [here](https://docs.google.com/document/d/1T3YmOaQ2Dhv_Lc_dRXsglv27qzsSShnJOYOwDuOKeL0/pub).

### DIFF
--- a/src/tcgwars/logic/impl/gen4/RisingRivals.groovy
+++ b/src/tcgwars/logic/impl/gen4/RisingRivals.groovy
@@ -343,10 +343,11 @@ public enum RisingRivals implements LogicCardInfo {
             }
           }
           move "Power Swing", {
-            text "60+ damage. Does 60 damage plus 10 more damage for each Evolved Pokémon on your Bench."
+            text "60+ damage. Does 60 damage plus 10 more damage for each Evolution Pokémon on your Bench."
+            //Errata: Evolution 「進化ポケモン」, not Evolved 進化している. https://www.pokemon-card.com/rules/faq/search.php?freeword=パワースイング
             energyCost C, C, C
             onAttack {
-              damage 60 + 10 * my.bench.findAll{it.evolution}.size()
+              damage 60 + 10 * my.bench.findAll{it.realEvolution}.size()
             }
           }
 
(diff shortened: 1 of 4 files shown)


src/tcgwars/logic/impl/gen4/RisingRivals.groovy
```diff
@@ -2717,14 +2718,18 @@ public enum RisingRivals implements LogicCardInfo {
...
           weakness L, PLUS10
           move "Cosmic Draw", {
-            text "If your opponent has any Evolved Pokémon in play, draw 3 cards."
+            text "If your opponent has any Evolution Pokémon in play, draw 3 cards."
+            //Errata: Evolution 「進化ポケモン」, not Evolved 進化している. https://www.pokemon-card.com/card-search/details.php/card/22091/regu/all
             energyCost C
             attackRequirement {
-              assert my.deck : "Your deck is empty"
-              assert opp.all.find{it.evolution} : "Your opponent doesn't have any Evolved Pokémon in play"
+              assert my.deck : "Your deck is empty"
```
axpendix, line 2725: why did we remove the condition here?

dxdydzd: It moved into onAttack, so the attack can still be used if the opponent doesn't have any Evolutions in play, but it will just do nothing.

Side note, compulsory switch Drag Off should also work like this, but that's a problem for future me to solve.

axpendix: but why should we allow an attack to do nothing? in case that's a misplay, isn't that bad for the user?

dxdydzd: True, but that's the way it works in the physical TCG. If your opponent misplays and tries to attack even though it will do nothing, you are within your rights to not allow them to take it back. Unlike Trainers, players can use attacks even though those attacks will have no effect.

One card that does make a difference between an attack that can't be used vs an attack that will do nothing is Holon Circle: if you're trying to break a Holon Circle with a Staryu that has a non-Water Energy attached, then having a Cosmic Draw that does nothing will let you do so (you declare the attack, then Holon Circle triggers and gets discarded), but having a Cosmic Draw that has an attack requirement will not (because you can't even declare the attack to begin with).

That said, I am OK with it staying as an attack requirement for user-friendliness. It's unlikely to matter much either way since neither Staryu nor Holon Circle are commonly played cards, so for it to manifest, not only would you need two players with rarely played cards in their decks matching against each other, the Staryu player would also need to be in a situation where he has a non-Water Energy attached, which puts it further in the realm of impossibility.

axpendix: Great, do you want me to make the change?

dxdydzd: Sure

PR #1374, opened 2025-05-13
Add Ability MEW_11 Mysterious Tail

Amwagner45:
Hey I am interested in contributing to the project and implement the attacks and abilities for these cards that are not yet added to TCGOne. I would like to test if this ability works as I intended and would appreciate your feedback on how best to do this testing on my end locally before issuing a pull request. 

This PR is to add the Mysterious Tail ability for MEW_11 in the Celebrations set (the set after Evolving skies which is the latest updated set in TCGOne).


axpendix (changes requested):

axpendix:
Hey, thanks for the PR!

PR #1382, opened 2025-10-01
TCGONE is down

TheAquaPiplup:
I am not sure if you are aware but TCGONE has been down since yesterday. I am contacting you through this because i believe it may get to you faster.


TheAquaPiplup:
Site is back up

PR #1387, opened 2026-06-23
Legacy select format card support

crait:


src/tcgwars/logic/impl/gen8/AstralRadiance.groovy
```diff
@@ -1829,13 +1830,8 @@ public enum AstralRadiance implements ImplOnlyCardInfo {
...
 
-      case DARK_PATCH_139: return cardng (stub) {
-        // Attach a basic [D] Energy card from your discard pile to 1 of your Benched [D] Pokémon.
-        onPlay {
-        }
-        playRequirement{
-        }
-      }
+      case DARK_PATCH_139:
+        return copy(DarkExplorers.DARK_PATCH_93, this)
```
axpendix, line 1834: With the new implementations this is now specified in cards/yaml files in "variantOf" field instead of the implementation files. Once that's specified then during the implementation discovery phase, it automatically receives the implementation of the variant. Then these lines should be deleted. Sorry, it's an undocumented new part of the engine.

axpendix, line 1834: With the new implementations this is now specified in cards/yaml files in "variantOf" field instead of the implementation files. Once that's specified then during the implementation discovery phase, it automatically receives the implementation of the variant. Then these lines should be deleted. Sorry, it's an undocumented new part of the engine.

src/tcgwars/logic/impl/gen8/AstralRadiance.groovy
```diff
@@ -1899,13 +1895,26 @@ public enum AstralRadiance implements ImplOnlyCardInfo {
...
+                my.prizeCardSet.set(my.prizeCardSet.indexOf(tar.first()), thisCard)
+                my.hand.add(tar.first())
+                bc "Took $tar and shuffled $thisCard into prizes"
+              }
+              my.prizeCardSet.shuffle()
+            }
+          }
+          playRequirement{
+          }
+        };
```
axpendix, line 1917: Need go one step above, meaning that we are now using cardng structure instead with no metadata.

src/tcgwars/logic/impl/gen8/AstralRadiance.groovy
```diff
@@ -1959,15 +1968,24 @@ public enum AstralRadiance implements ImplOnlyCardInfo {
...
+                  if(it.from==self && it.to.active){
+                    it.flags.add(DamageManager.DamageFlag.FORCE_X3_WEAKNESS)
+                  }
+                }
+              }
+          }
+          onRemoveFromPlay {
+            effect.unregister()
+          }
+        };
```
axpendix, line 1988: Same as before the structure must match cardng

src/tcgwars/logic/impl/gen8/BrilliantStars.groovy
```diff
@@ -160,15 +163,15 @@ public enum BrilliantStars implements ImplOnlyCardInfo {
   TORNADUS_126,
   HAWLUCHA_127,
   DRAMPA_V_128,
-  ACEROLA_S_PREMONITION_129,
+  ACEROLA_S_PREMONITION_129 ("Acerola's Premonition", "129", Rarity.UNCOMMON, [SUPPORTER, TRAINER]),
```
axpendix, line 166: Metadata is now split to YAML files so this line should be reverted

src/tcgwars/logic/impl/gen8/BrilliantStars.groovy
```diff
@@ -1843,13 +1852,8 @@ public enum BrilliantStars implements ImplOnlyCardInfo {
...
 
-      case COLLAPSED_STADIUM_137: return cardng (stub) {
-        // Each player can't have more than 4 Benched Pokémon. If a player has 5 or more Benched Pokémon, they discard Benched Pokémon until they have 4 Pokémon on the Bench. The player who played this card discards first. If more than one effect changes the number of Benched Pokémon allowed, use the smaller number.
-        onPlay {
-        }
-        onRemoveFromPlay{
-        }
-      }
+      case COLLAPSED_STADIUM_137:
+        return copy(GymHeroes.NARROW_GYM_124, this)
```
axpendix, line 1856: Should be specified as a variant as well since it has the same text.

